PR #18, opened 2017-01-28
Check type parameter kinds

### LINKED ISSUE #16: check kinds in lowering
Currently we check that when you instantiate a struct or trait you supply the right number of parameters, but we don't check that they have the right kind (i.e., type vs lifetime). I added an (ignored) test `check_struct_kinds` that demonstrates this. It'd be nice to fix this.

To fix it, one would go into `chalk-rust/src/lower/mod.rs` and look for places where we check the number of parameters, e.g.:

```rust
        if args.len() != info.addl_parameter_kinds.len() {
            bail!("wrong number of parameters for associated type (expected {}, got {})",
                  info.addl_parameter_kinds.len(), args.len())
        }
```

and adapt those to also check that the `kinds` in `args` match the kinds of `info.addl_parameter_kinds` (in that particular case).

cramertj:
Fix #16.

### DIFF
--- a/chalk-rust/src/lower/mod.rs
+++ b/chalk-rust/src/lower/mod.rs
@@ -511,6 +511,15 @@ impl LowerStructDefn for StructDefn {
     }
 }
 
+fn check_type_kinds<A: Kinded, B: Kinded>(msg: &str, expected: &A, actual: &B) -> Result<()> {
+    let expected_kind = expected.kind();
+    let actual_kind = actual.kind();
+    if expected_kind != actual_kind {
+        bail!("{}: expected {}, found {}", msg, expected_kind, actual_kind);
+    } else {
+        Ok(())
+    }
+}
 
 trait LowerTraitRef {
     fn lower(&self, env: &Env) -> Result<ir::TraitRef>;
@@ -536,13 +545,7 @@ impl LowerTraitRef for TraitRef {
         }
 
         for (binder, param) in k.binders.binders.iter().zip(parameters.iter().skip(1)) {
-            match (binder, param) {
-                (&ir::ParameterKind::Ty(_), &ir::ParameterKind::Lifetime(_)) =>
-                    bail!("type parameter to trait has wrong kind: expected type, found lifetime"),
-                (&ir::ParameterKind::Lifetime(_), &ir::ParameterKind::Ty(_)) =>
-                    bail!("type parameter to trait has wrong kind: expected lifetime, found type"),
-                _ => {}
-            }
+            check_type_kinds("incorrect kind for trait parameter", binder, param)?;
         }
 
         Ok(ir::TraitRef {
@@ -572,13 +575,7 @@ impl LowerProjectionTy for ProjectionTy {
         }
 
         for (param, arg) in info.addl_parameter_kinds.iter().zip(args.iter()) {
-            match (param, arg) {
-                (&ir::ParameterKind::Ty(_), &ir::ParameterKind::Lifetime(_)) =>
-                    bail!("associated type parameter has wrong kind: expected type, found lifetime"),
-                (&ir::ParameterKind::Lifetime(_), &ir::ParameterKind::Ty(_)) =>
-                    bail!("associated type parameter has wrong kind: expected lifetime, found type"),
-                _ => {}
-            }
+            check_type_kinds("incorrect kind for associated type parameter", param, arg)?;
         }
 
         args.extend(trait_parameters);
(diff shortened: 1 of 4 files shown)


cramertj:
Ping @nikomatsakis

nikomatsakis:
@cramertj woah! :) Sorry, I totally didn't see this!

chalk-rust/src/lower/mod.rs
```diff
@@ -521,6 +521,16 @@ impl LowerTraitRef for TraitRef {
                   k.binders.len() + 1, parameters.len())
         }
 
+        for (binder, param) in k.binders.binders.iter().zip(parameters.iter().skip(1)) {
+            match (binder, param) {
+                (&ir::ParameterKind::Ty(_), &ir::ParameterKind::Lifetime(_)) =>
+                    bail!("type parameter to trait has wrong kind: expected type, found lifetime"),
+                (&ir::ParameterKind::Lifetime(_), &ir::ParameterKind::Ty(_)) =>
+                    bail!("type parameter to trait has wrong kind: expected lifetime, found type"),
+                _ => {}
```
nikomatsakis: Can you go ahead and make these matches exhaustive? I would probably write it like this:

```
match (binder, param) {
    (&ir::ParameterKind::Ty(_), &ir::ParameterKind::Ty(_)) => { }
    (&ir::ParameterKind::Lifetime(_), &ir::ParameterKind::Lifetime(_)) => { }
    (&ir::ParameterKind::Ty(_), _) |
    (&ir::ParameterKind::Lifetime(_), _) => {
        bail!("incorrect parameter kind: expected {}, found {}", kind(binder), kind(param))
    }
}
```

where `kind` is some helper function. Note that this library is not (yet, anyway) really intended for *end-users per se*, so error messages don't have to be super-duper. Anyway, exhaustive matches really help when refactoring (i.e, adding a new sort of `ParameterKind`), so I try to avoid `_` (in chalk in particular, though also other places).

chalk-rust/src/lower/mod.rs
```diff
@@ -547,6 +557,16 @@ impl LowerProjectionTy for ProjectionTy {
                   info.addl_parameter_kinds.len(), args.len())
         }
 
+        for (param, arg) in info.addl_parameter_kinds.iter().zip(args.iter()) {
+            match (param, arg) {
+                (&ir::ParameterKind::Ty(_), &ir::ParameterKind::Lifetime(_)) =>
+                    bail!("associated type parameter has wrong kind: expected type, found lifetime"),
+                (&ir::ParameterKind::Lifetime(_), &ir::ParameterKind::Ty(_)) =>
+                    bail!("associated type parameter has wrong kind: expected lifetime, found type"),
+                _ => {}
```
nikomatsakis: (also exhaustive here please)

chalk-rust/src/lower/mod.rs
```diff
@@ -594,6 +614,16 @@ impl LowerTy for Ty {
 
                 let parameters = args.iter().map(|t| Ok(t.lower(env)?)).collect::<Result<Vec<_>>>()?;
 
+                for (param, arg) in k.binders.binders.iter().zip(args.iter()) {
+                    match (param, arg) {
+                        (&ir::ParameterKind::Ty(_), &Parameter::Lifetime(_)) =>
+                            bail!("type parameter has wrong kind: expected type, found lifetime"),
+                        (&ir::ParameterKind::Lifetime(_), &Parameter::Ty(_)) =>
+                            bail!("type parameter has wrong kind: expected lifetime, found type"),
+                        _ => {}
```
nikomatsakis: ( and here =) )

chalk-rust/src/lower/test.rs
```diff
@@ -108,14 +108,87 @@ fn atc_accounting() {
...
+    lowering_error! {
+        program {
+            struct Foo<'a> { }
+            struct i32 { }
+            trait Bar { }
+            impl Bar for Foo<i32> { }
+        }
+        error_msg {
+            "type parameter has wrong kind: expected lifetime, found type"
+        }
```
nikomatsakis: 👍 to pretty tests. Nice. =)

cramertj:
@nikomatsakis Changed to an exhaustive check against a shared `Kind` enum that's exposed via a `Kinded` trait. Thoughts?

cramertj:
Ping @nikomatsakis.

nikomatsakis:
@cramertj I like it, very nice. Sorry again for delay, I've been AFK basically all this week due to a trip.

cramertj:
@nikomatsakis No worries! Hope you had a good trip.

PR #28, opened 2017-03-31
Separate query binders into three sets of indexes.

### LINKED ISSUE #27: Questionable use of "punning" variable indexes
The code [here](https://github.com/nikomatsakis/chalk/blob/85b55bc9e2fc9c91ac27962248c02622010116a3/src/solve/match_clause.rs#L26) looks suspect. It appears to assume that the inference table keys generated will go from `0..q.binders.len()`, but this doesn't happen since difference kinds of variables (types, lifetimes, crates) are indexed separately.

If this is indeed a bug, you'd expect it to trigger panics due to not being able to find keys. I'm working toward a test case illustrating that behavior.

withoutboats:
Resolves #27 

I have very little confidence that this is actually correct. Best way to learn things. :-)

In order to make sure param indexes match unification table indexes, I separated the binders param on Query into one vec per kind.

This still doesn't compile because I haven't figured out how `InferenceTable::instantiate` works. I figured it would be good to stop now and find out if I'm on the wrong track.

### DIFF
--- a/src/solve/infer/query.rs
+++ b/src/solve/infer/query.rs
@@ -29,7 +29,7 @@ impl InferenceTable {
         debug!("make_query({:#?})", value);
         let mut q = Querifier {
             table: self,
-            free_vars: Vec::new(),
+            free_vars: QueryBinders::default(),
         };
         let r = value.fold_with(&mut q, 0).unwrap();
         Query {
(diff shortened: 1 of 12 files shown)


nikomatsakis:
@withoutboats argh sorry, will take a look asap :)

withoutboats:
No worries, I know y'all are at some kind of thing

src/solve/infer/query.rs
```diff
@@ -45,33 +45,34 @@ struct Querifier<'q> {
...
+                        InferenceValue::Unbound(ui) => binders.krates.push(ui),
+                        InferenceValue::Bound(_) => panic!("free var now bound"),
+                    }
+                }
+            }
+        }
+        binders
     }
 
     fn add(&mut self, free_var: ParameterInferenceVariable) -> usize {
```
nikomatsakis: I think you're going to have to change the `add()` method too, right? In particular, I think you want 3 sets of `free_vars` lists.

withoutboats:
Separated the Querifier free vars into three vecs (using the same `QueryBinders` type). Still have to deal with instantiate.

withoutboats:
Got it to compile and pass existing tests, added the test from #27, which panics trying to unwrap a variable as a lifetime which isn't one.. Seems like `Instantiator` also needs to keep separate vecs but I'm not certain.

withoutboats:
Changed instantiator, tests still failing though.

withoutboats:
Okay, I did some investigation and it seemed that the Instantiator's fold was not operating on types correctly, causing the index of the variables to become misaligned. Here's some of the debug output from the `mixed_indices` test case:

```rust
// This branch:
push_goal((Ref<'?1, ?1> = Ref<'?1, ?1>), Env(U0, []))

// master (correct):
push_goal((Ref<'?0, ?0> = Ref<'?0, ?1>), Env(U0, []))
```

This seems to be because ApplicationTys are still using a single vector of params, causing this math to malfunction:

```rust
// From instantiate's fold_free_var:
Ok(Ty::Var(depth + binders - self.vars.tys.len()))
// In a Ty, this also needs to subtract lifetime & crate binders lens
```

It seems like an `ApplicationTy` needs to keep three vectors of params instead of one. I'm guessing the other types need to as well. I suspect its the case that *every* place chalk currently uses a `Vec<ParameterKind<T, L, C>>` it should actually use three vectors.

withoutboats:
Having put some more work into making that transformation, which touches absolutely everything I'm wondering if it wouldn't be easier to just create a mapping in InferenceTable from the shared de Bruijn indexes to their kind-specific index.

withoutboats:
Tried that approach for a minute, but the internal mappings would need to be updated as the indexes are shifted so I'm not sure how realistic it is.

At a bit of an impasse about what to do, will wait for feedback. :-)

nikomatsakis:
So initially @aturon and I were planning not to change how the binders work (i.e., by splitting them by kind), but rather to change the code that assumes that the indices line up (i.e., when we make a new inference table). Is there a reason that approach won't work? It seemed like it would be more limited, but perhaps it's only handling some of the cases?

aturon:
@nikomatsakis Yeah, @withoutboats and I also talked about both possibilities. It clearly seems that splitting up the binders is more pain than it's worth.

I've opened https://github.com/nikomatsakis/chalk/pull/29 to fix the original problem more directly, while keeping the overall numbering strategy in place.

nikomatsakis:
Closing this in favor of #29. =)

PR #29, opened 2017-04-11
Remove punning of bound variables and fresh inference table variables

### LINKED ISSUE #27: Questionable use of "punning" variable indexes

aturon:
Resolves #27


withoutboats:
Well this is certainly a much simpler solution. :) Very educational, thanks

PR #30, opened 2017-04-13
Add a repl as an included binary

aturon:
Pretty straightforward, but really helpful for exploration.

### DIFF
--- a/src/bin/repl.rs
+++ b/src/bin/repl.rs
@@ -1,0 +1,129 @@
+extern crate rustyline;
+extern crate chalk_parse;
+extern crate chalk;
+
+#[macro_use]
+extern crate error_chain;
+
+use std::io::Read;
+use std::fs::File;
+use std::sync::Arc;
+
+use chalk::ir;
+use chalk::lower::*;
+use chalk::solve::prove::Prove;
+use chalk::solve::solver::Solver;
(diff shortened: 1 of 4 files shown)


aturon:
cc @withoutboats, I bet you'll find this useful too!

src/bin/repl.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::sync::Arc;
+
+use chalk::ir;
+use chalk::lower::*;
+use chalk::solve::prove::Prove;
+use chalk::solve::solver::Solver;
+
+use rustyline::error::ReadlineError;
+
+error_chain! {
```
nikomatsakis: I ❤️ `error_chain`

cc @brson :)

aturon: Same here! We should consider it for the nursery, tbh...

withoutboats:
Haha awesome! I had definitely thought about this; in the long term it seems like one of the best uses for chalk.

nikomatsakis:
rustup seems to be broken with travis :(

aturon:
@nikomatsakis there's a rustup outage for nightly right now. should resolve in a day's time.

nikomatsakis:
Forgot about this.

PR #32, opened 2017-04-27
Remove primitive negated goals, crates, and LocalTo predicates

aturon:
This is all in preparation for the [move to a modal world](http://aturon.github.io/blog/2017/04/24/negative-chalk/), where negation will be a "derived" operator, and we only need to know whether a given type/trait is "upstream" or not"


src/solve/infer/mod.rs
```diff
@@ -45,31 +39,21 @@ impl InferenceTable {
...
     }
 
     pub fn snapshot(&mut self) -> InferenceSnapshot {
         let ty_unify_snapshot = self.ty_unify.snapshot();
-        let krate_unify_snapshot = self.krate_unify.snapshot();
-        InferenceSnapshot {
-            ty_unify_snapshot,
-            krate_unify_snapshot,
-        }
+        InferenceSnapshot { ty_unify_snapshot }
```
nikomatsakis: Hmm, this seems like a pre-existing bug, but shouldn't there be a `lifetime_snapshot` in this struct too...? I guess I forgot to add one.

src/solve/infer/mod.rs
```diff
@@ -45,31 +39,21 @@ impl InferenceTable {
...
     }
 
     pub fn rollback_to(&mut self, snapshot: InferenceSnapshot) {
         self.ty_unify.rollback_to(snapshot.ty_unify_snapshot);
```
nikomatsakis: Similarly, we would need to rollback here.

aturon:
I'm removing snapshots entirely on my current branch, so we should just merge as-is.

nikomatsakis:
OK.

PR #36, opened 2017-05-27
Use substitutions and suggestions in Solution; implement negation

aturon:
This commit is a major refactoring with a few goals:

- Move `Solution` to a Unique/Ambiguous model, with the ambiguous case
  encompassing multiple forms of guidance for type inference (including
  "suggestions" intended to be applied as fallback).

- Remove the notion of goal refinement, and instead use substitutions
  for communicating solutions.

- Draw a more clear line between the generic Chalk engine (which "only
  knows Prolog", i.e. doesn't know anything special about Rust) and
  lowering. Put differently, everything specific to Rust should now be
  encoded in the lowering, rather than though special cases in the Chalk
  engine. 

- Consolidate the solver code so that the "strategies" for solving leaf
  goals are more clear and all co-located.

- Introduce some of the heuristics that rustc performs for inference,
  e.g. preferring `where` clauses for influencing inference. We do this
  in a way that decouples the heuristics from the rest of the solver;
  the `favor_over` method, in particular, contains the relevant logic.

- Introduce `not { ... }` as a logical operator, using negation-as-failure
semantics adjusted for our "three-valued" logic.


aturon:
cc @nikomatsakis @withoutboats

src/solve/infer/test.rs
```diff
@@ -190,43 +190,43 @@ const U0: UniverseIndex = UniverseIndex { counter: 0 };
...
-        .unwrap();
-
-    assert_eq!(
-        table.make_query(&ty!(apply (item 0) (expr v2b) (expr v2a) (expr v1) (expr v0))),
-        Query {
-            value: ty!(apply (item 0) (apply (item 1) (var 0) (var 1)) (var 2) (var 0) (var 1)),
-            binders: vec![ParameterKind::Ty(U1), ParameterKind::Ty(U0), ParameterKind::Ty(U2)],
-        });
-}
+// #[test]
```
aturon: Oops, forgot to get back to these.

aturon:
I've now pushed up a commit with a preliminary implementation of negation, which is straightforward to add. However, it doesn't currently handle `forall` variables correctly when it comes to unification: you can prove `forall<T, U> { not { T = U } }`. This is straightforward to fix, but will require a refactor of unification to plumb the needed information. Will do that soon.

aturon:
I've now pushed up an additional commit which adjusts the treatment of unification for skolemized type variables so that negation does the Right Thing. TLDR: we now treat `forall<T, U> { T = U }` as ambiguous, rather than false (by treating the inner equality as ambiguous); that means that we correctly *also* say that `forall<T, U> { not { T = U } }` is ambiguous.

aturon:
Update: now includes a commit that implements normalization fallback via lowering to negation!

aturon:
OK, this PR is as big as it's going to get -- it now does not regress Chalk in any way, and fleshes out the complete story for negation.

I'll be working on the `compat` modality next, bound for a fresh PR.

withoutboats:
Do you think this will be stable enough to branch off for overlap/specialization or should I wait until its landed?

aturon:
@withoutboats I expect it to be stable, unless @nikomatsakis has a very surprising reaction to it :-)

src/solve/fulfill.rs
```diff
@@ -180,87 +202,82 @@ impl<'s> Fulfill<'s> {
...
+            // package up what we learned about type inference as an ambiguous
+            // solution.
 
-        debug!("fulfill::solve_one: new_type_info={}", new_type_info);
+            if subst.is_trivial(&mut self.infer) {
+                // In this case, we didn't learn *anything* definitively. So
+                // now, we go one last time through the obligations, this time
+                // applying even *tentative* inference suggestions, so that we
+                // can yield these upwards as our own suggestions. In
+                // particular, we yield up the first one we can find.
```
nikomatsakis: First in what sense? we're popping the obligations, after all, so this isn't the first we would have encountered had we been applying the substs "in order", right? (Or am I missing something?)

src/solve/fulfill.rs
```diff
@@ -180,87 +202,82 @@ impl<'s> Fulfill<'s> {
...
+                        let subst = self.infer.canonicalize(&subst);
+                        return Ok(Solution::Ambig(Guidance::Suggested(subst.quantified)));
+                    }
+                }
 
-            if new_type_info {
-                *inference_progress = true;
+                Ok(Solution::Ambig(Guidance::Unknown))
+            } else {
+                // While we failed to prove the goal, we still leared that
```
nikomatsakis: Nit: s/leared/learned/

src/solve/fulfill.rs
```diff
@@ -180,87 +202,82 @@ impl<'s> Fulfill<'s> {
...
+                // impl Display for Bar {}
+                // impl Display for Baz {}
+                //
+                // impl<T> Debug for Foo<T> where T: Display {}
+                // ```
+                //
+                // If we pose the goal `exists<T> { T: Debug }`, we can't say
+                // for sure what `T` must be (it could be either `Foo<Bar>` or
+                // `Foo<Baz>`, but we *can* say for sure that it must be of the
+                // form `Foo<?0>`.
```
nikomatsakis: Nice comment.

src/solve/fulfill.rs
```diff
@@ -217,67 +241,119 @@ impl<'s> Fulfill<'s> {
...
+    }
+
+    fn fulfill_negative(&mut self) -> Result<Outcome> {
+        // Attempt to refute each negative obligations by in a fresh Fulfill
+        for goal in self.to_refute.drain(..) {
+            let canonicalized = self.infer.canonicalize(&goal);
+            let goal = canonicalized.quantified.value;
+
+            // Negation cannot be used to resolve existential variables, and do
+            // not have a useful (for us) logial meaning when they contain
```
nikomatsakis: Nit: s/logial/logical/

src/solve/fulfill.rs
```diff
@@ -217,67 +241,119 @@ impl<'s> Fulfill<'s> {
...
+            // existential variables; treat negative goals with free variables
+            // as ambiguous
+            if !canonicalized.free_vars.is_empty() {
+                return Ok(Outcome::Incomplete);
+            }
+
+
```
nikomatsakis: Nit: extra blank line

src/solve/fulfill.rs
```diff
@@ -217,67 +241,119 @@ impl<'s> Fulfill<'s> {
...
+            let canonicalized = self.infer.canonicalize(&goal);
+            let goal = canonicalized.quantified.value;
+
+            // Negation cannot be used to resolve existential variables, and do
+            // not have a useful (for us) logial meaning when they contain
+            // existential variables; treat negative goals with free variables
+            // as ambiguous
+            if !canonicalized.free_vars.is_empty() {
+                return Ok(Outcome::Incomplete);
+            }
```
nikomatsakis: This is stronger than necessary, right? For example, imagine `Vec<?T>: Copy` -- given that there are no impls of `Copy` for `Vec` at all, and assuming a closed world, we could figure out that this is false. In particular, if we get a `Unique` result with an empty substitution, we are satisfied, even if `free_vars` is non-empty...?

src/solve/fulfill.rs
```diff
@@ -217,67 +241,119 @@ impl<'s> Fulfill<'s> {
...
-                //
-                // If we pose the goal `exists<T> { T: Debug }`, we can't say
-                // for sure what `T` must be (it could be either `Foo<Bar>` or
-                // `Foo<Baz>`, but we *can* say for sure that it must be of the
-                // form `Foo<?0>`.
-                let subst = self.infer.canonicalize(&subst);
-                Ok(Solution::Ambig(Guidance::Definite(subst.quantified)))
+        // Next we look at the goals we must refute; doing so cannot create new
+        // obligations of any form
+        let negative = self.fulfill_negative()?;
```
nikomatsakis: Clever to stratify these like that, though probably not the most performant thing ever (i.e., we might have been able to see that we will fail to refute a "refute goal" and hence abort the process much earlier).

src/lower/mod.rs
```diff
@@ -957,3 +981,49 @@ impl ir::TraitDatum {
...
+        //
+        // Given:
+        //
+        //    trait SomeTrait {
+        //        type Assoc;
+        //    }
+        //
+        // we generate:
+        //
+        //    ?T: SomeTrait<Assoc = (SomeTrait::Assoc)<?T>> :-
```
nikomatsakis: question -- currently, for each `type Foo = Bar` in an impl or the where-clause list, we generate *two* clauses:

    <T as Trait>::Foo ==> Bar

and 

    known { <T as Trait>::Foo ==> Bar }

Maybe we should just translate the impl declarations along with `T: Trait<Foo=Bar>` where-clauses into the "known" form to start, and then, for each declaration in the trait, have two instances of the "general rule":

    <?T as Trait>::Foo ==> Bar :- known { <?T as Trait>::Foo ==> Bar }
    <?T as Trait>::Foo ==> <?T as Trait>::Foo :- not { known { <?T as Trait>::Foo ==> Bar } }

Obviously, this is equivalent. Just seems a bit clearer to me somehow.

src/solve/fulfill.rs
```diff
@@ -115,152 +133,235 @@ impl<'s> Fulfill<'s> {
...
+        for goal in self.to_refute.drain(..) {
+            let canonicalized = self.infer.canonicalize(&goal);
+            let goal = canonicalized.quantified.value;
+
+            // Negation cannot be used to resolve existential variables, and do
+            // not have a useful (for us) logial meaning when they contain
+            // existential variables; treat negative goals with free variables
+            // as ambiguous
+            if !canonicalized.free_vars.is_empty() {
+                return Ok(Outcome::Incomplete);
```
nikomatsakis: **Question:** This is stronger than necessary, right? It seems like any result of `Unique` with an empty substitution is probably good enough. Consider something like `not { Vec<?T>: Copy }`... we know that it is false (given closed-world, at least), simply because there are no impls of `Copy` *at all* that could apply.

(I made this comment earlier, but it got marked as "outdated" by GH for some reason.)

src/solve/fulfill.rs
```diff
@@ -115,152 +133,235 @@ impl<'s> Fulfill<'s> {
...
+        // haven't proved that it's *impossible* to satisfy out obligations. we
+        // need to determine how to package up what we learned about type
+        // inference as an ambiguous solution.
+
+        if subst.is_trivial(&mut self.infer) {
+            // In this case, we didn't learn *anything* definitively. So now, we
+            // go one last time through the positive obligations, this time
+            // applying even *tentative* inference suggestions, so that we can
+            // yield these upwards as our own suggestions. In particular, we
+            // yield up the first one we can find.
```
nikomatsakis: **Question:** First in what sense? we're popping the obligations, after all, so this isn't the first we would have encountered had we been applying the substs "in order", right? (Or am I missing something?)

(I made this comment earlier, in my review, but GH marked it as "outdated" for some reason.)

src/solve/fulfill.rs
```diff
@@ -115,152 +133,235 @@ impl<'s> Fulfill<'s> {
...
+                if let Some(constrained_subst) = solution.constrained_subst() {
+                    self.apply_solution(free_vars, constrained_subst);
+                    let subst = self.infer.canonicalize(&subst);
+                    return Ok(Solution::Ambig(Guidance::Suggested(subst.quantified)));
+                }
+            }
+
+            Ok(Solution::Ambig(Guidance::Unknown))
+        } else {
+            // While we failed to prove the goal, we still leared that
```
nikomatsakis: Typo: s/leared/learned/

(I made this comment earlier, in my review, but GH marked it as "outdated" for some reason.)

nikomatsakis:
> I expect it to be stable, unless @nikomatsakis has a very surprising reaction to it :-)

Closing with prejudic--- oh, no, I guess it's fine.

src/fold/mod.rs
```diff
@@ -139,6 +140,23 @@ impl Fold for Lifetime {
...
 }
 
+impl Fold for Substitution {
+    type Result = Substitution;
+    fn fold_with(&self, folder: &mut Folder, binders: usize) -> Result<Self::Result> {
+        let mut tys = BTreeMap::new();
+        let mut lifetimes = BTreeMap::new();
+
+        for (var, ty) in &self.tys {
+            tys.insert(*var, ty.fold_with(folder, binders)?);
```
aturon: The `?` here is what seems to make the imperative more clear and nearly as concise, sadly.

nikomatsakis: Maybe something like this?

```
let tys = self.tys.iter().map(|(&var, ty)| (var, ty.fold_with(folder, binders)?)).collect::<Result<BTreemap<_, _>>>()?;
```

Anyway, doesn't matter. More readable this way =)

scalexm:
Actually I think I've found a regression. Load this Chalk program:
```rust
struct Foo { }

trait Marker { }
impl Marker for Foo { }
```

and then try to solve the goal
```
forall<T> { if (T: Marker) { T: Marker } }
```

current master branch answers: `Solution { successful: Yes ...`
while aturon's maybe-subst branch answers: `Ambiguous; suggested substitution []`

If you remove `impl Marker for Foo { }` then everything works fine.

In fact, aturon's branch will see the two rules `!1: Marker` and `Foo: Marker` into the environment, and it will try to unify the goal `!1: Marker` against these two rules and end up unifying `!1` both as `!1` and `Foo`. But I guess it should not unify `!1` as `Foo` since it is a ! variable.

This bug is not caught by the `prove_forall` test because in this test, `impl Marker for Foo { }` is replaced by `impl<T> Marker for Vec<T> { }` and then the unification `!1` as `Vec<?0>` cannot be done, so the issue does not appear.

src/fold/mod.rs
aturon: Right, that's exactly what I mean -- I find this iterator-chained version much harder to read...

aturon:
OK, I've pushed up a couple of commits addressing @nikomatsakis's feedback. In particular, I added some commentary trying to explain in greater depth why we don't want to allow negations over free variables. Note also that this follows the [original paper](ww.doc.ic.ac.uk/~klc/NegAsFailure.pdf) on negation as failure.

Also, I've collapsed the "clever" stratification of negative and positive goals. I think this new version is nicer.

@scalexm, I'll look at the issue you raised next.

aturon:
@scalexm OK, I've investigated, and believe this comes down to the imprecision discussed in [this comment](https://github.com/nikomatsakis/chalk/pull/36/files#diff-9dbb1406b7c2c4a613771ce806a83e60R57) about combining solutions. I don't see any harm in tightening that up, but am out of time for the moment.

nikomatsakis:
@aturon 

> In particular, I added some commentary trying to explain in greater depth why we don't want to allow negations over free variables. 

Thanks, that comment made sense, but I still think we could be yet more precise (not a blocker by any means, though). In your example:

```rust
struct Vec<T> {}
struct i32 {}
struct u32 {}
trait Foo {}
impl Foo for Vec<u32> {}
```

Given the query `exists<T> { not { Vec<T>: Foo } }`, I would expect a response of "ambiguous" or "maybe" or whatever your branch is calling it. In particular, we would try to solve `Vec<?T>: Foo` and we would get back a result that is `Unique(?T = u32)`. Since the resulting substitution is non-trivial (what I called non-empty before), we would return `Ok(NegativeSolution::Ambiguous)`.

Anyway, seems fine as is.

> OK, I've investigated, and believe this comes down to the imprecision discussed in this comment about combining solutions. I don't see any harm in tightening that up, but am out of time for the moment.

yeah, I agree with this diagnosis.

nikomatsakis:
Test failure:

```
failures:

---- solve::test::mixed_indices_normalize_application stdout ----
	program { struct Ref < 'a , T > {  } trait Foo { type T ; } }
----------------------------------------------------------------------
goal {
exists < T > {
exists < 'a > { exists < U > { < Ref < 'a , T > as Foo > :: T = U } } } }
expected:
Unique
actual:
Ambiguous; definite substitution [?0 := ?0, ?1 := (Foo::T)<Ref<'?1, ?0>>, '?0 := '?1]
thread 'solve::test::mixed_indices_normalize_application' panicked at 'assertion failed: !expected1.is_empty() && result1.starts_with(&expected1)', src/solve/test.rs:57
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    solve::test::mixed_indices_normalize_application
```

nikomatsakis:
On IRC, @scalexm and I were saying that maybe they will investigate this regression that they found. I also pointed them at [this particular snippet of code](https://github.com/nikomatsakis/chalk/blob/master/src/solve/solver/mod.rs#L125-L148), which I think may be the bit of logic that @aturon's branch doesn't have. (But I could be wrong.)

aturon:
@nikomatsakis 

> Since the resulting substitution is non-trivial (what I called non-empty before), we would return Ok(NegativeSolution::Ambiguous).

Yeah, the confusion seemed to be around the precise conditions under which we'd go to ambiguous. But I agree we can punt this in any case.

aturon:
> On IRC, @scalexm and I were saying that maybe they will investigate this regression that they found. I also pointed them at this particular snippet of code, which I think may be the bit of logic that @aturon's branch doesn't have. (But I could be wrong.)

It's possible -- there are probably a few ways to deal with this. As much as possible, though, I'd like to push this kind of thing into the general logic for combining solutions.

nikomatsakis:
> As much as possible, though, I'd like to push this kind of thing into the general logic for combining solutions.

I agree that seems like the right place, didn't mean to suggest otherwise.

Issue #12, opened 2017-01-16
Elaborating where clauses on traits, structs


nikomatsakis:
Related to #11, we don't really do anything clever with where clauses on traits/structs. We need to support "elaboration" -- ideally, a richer form than what the current compiler supports. In particular, we should be able to show that e.g. `T: Ord => T: PartialOrd` (supertrait) but also any other where-clause on traits, and in particular I'd like a better form of support around projections of associated types (e.g., `T: Foo<Bar = U>` should let us prove about `U` whatever we can prove about `<T as Foo>::Bar`).

nikomatsakis:
We now support a basic form of elaboration, but I am not happy with it. In particular, we have some hard-coded rules that take in the environment and "elaborate" it at various points, much like the compiler does. My ideal would be to find a solution that encodes these relationships "into the logic" as program-clauses. For example, you might imagine:

```rust
pub trait Foo: Bar { }
```

would imply a rule like

```rust
(T: Bar) :- (T: Foo).
```

However, that doesn't *quite* work. In particular, I *also* want a program like this to be illegal:

```rust
struct X;
impl Bar for X { }
// impl Foo for X { } <-- this impl is missing!
```

Now the question is, why would this be illegal? If the WF requirements of the impl are that `X: Foo`, then we can prove that by showing that `X: Bar` (because the impl exists). That's sort of circular reasoning, clearly. So you could imagine that the impl has an implied where-clause, i.e., it's sort of equivalent to:

```rust
impl Bar for X
    where X: Foo
{ }
```

The only way to prove *that* would be to have an actual impl of `Foo` for `X` so that things terminate. This seems fine, except that I would like to extend elaboration to *all* where-clauses on a trait, and [we currently support circular where-clauses on traits](https://is.gd/tWTl6T):

```rust
trait Foo<T: ?Sized> where T: Bar<Self> { }
trait Bar<T: ?Sized> where T: Foo<Self> { }
impl Foo<i32> for i32 { }
impl Bar<i32> for i32 { }
fn main() { }
```

Under the approach I proposed, the impls would generate rules like this:

```
(i32: Foo<i32>) :- (i32: Bar<i32>).
(i32: Bar<i32>) :- (i32: Foo<i32>).
```

This would clearly never terminate. 

I'm not sure yet how properly to think about this.

withoutboats:
Note the first code block should be `pub trait Bar: Foo { }`

withoutboats:
[...]

PR #37, opened 2017-05-30
Merge `WhereClause` and `WhereClauseGoal` into a single enum

scalexm:
The two enums were merged into a single one named `WhereClause`. In particular, chalk syntax is now a superset of rust syntax and `WellFormed` types are now allowed into the environment. This is a preparation for solving #12.

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -24,17 +24,10 @@ macro_rules! reflexive_impl {
 
 reflexive_impl!(TraitRef);
 reflexive_impl!(WhereClause);
-reflexive_impl!(WhereClauseGoal);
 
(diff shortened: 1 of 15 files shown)


aturon:
Note, this change is subsumed by the overhaul [here](https://github.com/nikomatsakis/chalk/pull/36).

nikomatsakis:
OK, I'm going to close this in favor of #36, which does indeed do a similar thing -- though it separates out unifications, which I overlooked when talking to @scalexm in the first place (unification goals is probably why I made the split between "where-clause-goal" and "where-clause" goal to start).

nikomatsakis:
Sorry @scalexm for leading you down a false start. =)

PR #39, opened 2017-06-03
First version of overlap check.

withoutboats:
This check works by pairing every 2 impls of the same trait and then
generating a goal to prove that a type lies in the intersection of
those two impls. If the goal succeeds, we have found the overlap.

The goal that is lowered looks like this:

```rust
impl Foo for i32 { }
impl Foo for u32 { }

exists<> { i32 = u32 }
```

```rust
impl<T> Foo for T { }
impl Foo for i32 { }
exists<T> { T = i32 }
```

```rust
impl<T> Foo for Vec<T>
impl<T> Foo for Vec<i32>
exists<T> { Vec<T> = Vec<i32>
```

And so on.

Currently, this goal does not take into consideration where clauses
on the impl. For that reason, the local negative reasoning, as in the
example below, is not implemented:

```rust
trait Bar { }
trait Foo { }
struct MyType;
impl<T: Bar> Foo for T { }
impl T for MyType { }
```

A test exists to check this, which is failing.

### DIFF
--- a/src/lower/test.rs
+++ b/src/lower/test.rs
@@ -6,6 +6,30 @@ use ir::*;
 use std::sync::Arc;
 use super::{LowerProgram, LowerGoal};
 
+macro_rules! lowering_success {
+    (program $program:tt) => {
+        let program_text = stringify!($program);
+        assert!(program_text.starts_with("{"));
+        assert!(program_text.ends_with("}"));
+        assert!(parse_and_lower(&program_text[1..program_text.len()-1]).is_ok());
+    }
+}
+
+macro_rules! lowering_error {
+    (program $program:tt error_msg { $expected:expr }) => {
+        let program_text = stringify!($program);
+        assert!(program_text.starts_with("{"));
+        assert!(program_text.ends_with("}"));
+        let error = parse_and_lower(&program_text[1..program_text.len()-1]).unwrap_err();
+        let expected = Error::from($expected);
+        assert_eq!(
+            error.to_string(),
+            expected.to_string()
+        );
+    }
+}
+
+
 fn parse_and_lower(text: &str) -> Result<Program> {
     chalk_parse::parse_program(text)?.lower()
 }
@@ -15,18 +39,42 @@ fn parse_and_lower_goal(program: &Program, text: &str) -> Result<Box<Goal>> {
 }
 
 #[test]
-fn lower() {
-    parse_and_lower("struct Foo { field: Foo } trait Bar { } impl Bar for Foo { }").unwrap();
+fn lower_success() {
+    lowering_success! {
+        program {
+            struct Foo { field: Foo }
+            trait Bar { }
+            impl Bar for Foo { }
+        }
+    }
 }
 
 #[test]
 fn not_trait() {
-    parse_and_lower("struct Foo { } trait Bar { } impl Foo for Bar { }").unwrap_err();
+    lowering_error! {
+        program {
+            struct Foo { }
+            trait Bar { }
+            impl Foo for Bar { }
+        }
+        error_msg {
+            "expected a trait, found `Foo`, which is not a trait"
+        }
+    }
 }
 
 #[test]
 fn invalid_name() {
-    parse_and_lower("struct Foo { } trait Bar { } impl Bar for X { }").unwrap_err();
+    lowering_error! {
+        program {
+            struct Foo { }
+            trait Bar { }
+            impl Bar for X { }
+        }
+        error_msg {
+            "invalid type name `X`"
+        }
+    }
 }
 
 #[test]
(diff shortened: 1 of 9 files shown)


withoutboats:
Okay added consideration for where clauses. The test still fails, I'm guessing because that kind of negative reasoning just hasn't been implemented in chalk.

Where clauses from both impls are joined and then added as an Implies. As in:

```rust
trait Foo { }
trait Bar { }
trait Baz { }
impl<T> Foo for T where T: Bar { }
impl<T> Foo for T where T: Baz { }
/*
exists<0, 1> { (?0 = ?1) :- (?0: Bar /\ ?1: Baz) }
*/
```

```rust
trait Foo { }
trait Bar { }
struct Baz { }
impl<T> Foo for T where T: Bar { }
impl Foo for Baz { }
/*
exists<0> { (?0 = Baz) :- (?0: Bar) |
*/
```

This lowering feels too easy... I suspect I must have done something wrong. 😅

withoutboats:
That makes sense! I also just looked at the code and realized the rhs where clauses need to be upshifted also, not a good bug to have.

withoutboats:
There's a bug which is causing impls of the same trait not to group together in the `two_blanket_impls` spec. Its flakey so it works on some runs and not on others, will investigate.

withoutboats:
Fixed the bug, I had misunderstood itertools' group_by. This should be ready to merge if @nikomatsakis is happy with it.

src/lower/test.rs
```diff
@@ -191,3 +225,126 @@ fn check_parameter_kinds() {
...
+            impl Baz for Quux { }
+        }
+        error_msg {
+            "overlapping impls of trait \"Foo\""
+        }
+    }
+}
+
+#[test]
+#[ignore] // Ignored until we have the compat modality
```
nikomatsakis: rather than use `ignore`, I would prefer you write the test so that it tests the current behavior, but leave a comment explaining how it's wrong. This way we're more likely to notice when you fix it and re-enable the test...

src/overlap.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // Join the two impls' binders together 
+    let mut binders = lhs.binders.binders.clone();
+    binders.extend(rhs.binders.binders.clone());
+
+    // Upshift the rhs variables in params to account for the joined binders
+    let lhs_params = params(lhs).iter().cloned();
+    let rhs_params = params(rhs).iter().map(|param| param.up_shift(lhs.binders.len()));
+
+    // Create an equality goal for every input type the trrait, attempting
```
nikomatsakis: Nit: s/trrait/trait/

src/overlap.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let lhs_len = lhs.binders.len();
+
+    // Join the two impls' binders together 
+    let mut binders = lhs.binders.binders.clone();
+    binders.extend(rhs.binders.binders.clone());
+
+    // Upshift the rhs variables in params to account for the joined binders
+    let lhs_params = params(lhs).iter().cloned();
+    let rhs_params = params(rhs).iter().map(|param| param.up_shift(lhs.binders.len()));
```
nikomatsakis: Nit: you use `lhs.binders.len()` here but `lhs_len` below; made me think they were different

src/overlap.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // Upshift the rhs variables in where clauses
+    let lhs_where_clauses = lhs.binders.value.where_clauses.iter().cloned();
+    let rhs_where_clauses = rhs.binders.value.where_clauses.iter().map(|wc| wc.up_shift(lhs_len));
+
+    // Create a goal for each clause in both where clauses
+    let wc_goals = lhs_where_clauses.chain(rhs_where_clauses)
+                .map(|wc| Goal::Leaf(LeafGoal::DomainGoal(wc)));
+ 
+    // Join all the goals we've created together with And, then quanitfy them
```
nikomatsakis: Nit: s/quanitfy/quantify/

src/overlap.rs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+// The goal to test overlap.
+//
+// If this goal succeeds, these two impls overlap.
```
nikomatsakis: Even though this function is very nicely commented, I think it'd be *that much more* helpful to give an example of two impls and the resulting goal from this function.

src/lower/test.rs
```diff
@@ -191,3 +225,126 @@ fn check_parameter_kinds() {
...
+            trait Foo<T> { }
+            struct Baz { }
+
+            impl<T> Foo<Baz> for T { }
+            impl<T> Foo<T> for Baz { }
+        } error_msg {
+            "overlapping impls of trait \"Foo\""
+        }
+    }
+}
```
nikomatsakis: I think one area where it'd be good to have more test coverage would be associated types.

For example, I imagine something like this ought to be considered overlap:

```rust
trait Foo<T> { }

trait Iterator { type Item; }


struct Vec<T> { }
impl<T> Iterator for Vec<T> { type Item = T; }

impl<T: Iterator> Foo<<T as Iterator>::Item> for T { }
impl<A, B> Foo<A> for B { }
```

Hmm, that's not quite as interesting as I had thought. Well, I feel like there are interesting tests to be had here. =) Let's at least add that one ...

withoutboats:
Responded to all the nits, which were very good.

nikomatsakis:
Nice!

PR #40, opened 2017-06-05
Some trivial changes

fabric-and-ink:


src/ir/mod.rs
```diff
@@ -480,7 +480,9 @@ pub struct ProgramClause {
     pub implication: Binders<ProgramClauseImplication>,
 }
 
-/// Represents one clause of the form `consequence :- conditions`.
+/// Represents one clause of the form `consequence :- conditions` where
+/// `conditions = cond_1 || cond_2 || ...` is the disjunction of the individual
+/// conditions.
```
nikomatsakis: Actually, it's the **conjunction** =)

fabric-and-ink: Haha dang – got it right at first and changed it later in the 'Fix mix-up' commit :)

fabric-and-ink:
No problem! I'm in no hurry :)

PR #43, opened 2017-06-12
Record specialization priority in impls

withoutboats:
Building off the overlap branch, begin to implement specialization.

As we check overlapping impls, instead of just erroring when an overlap is found, we instead determine if one of the two impls specializes the other. If it does, we record this specialization in a forest (implemented with petgraph). If there's no specialization, the overlap is an error.

Once we've visited every impl, we walk the forest from the roots, using it to set specialization priorities in every impl datum in the original program.

Still to do:

* Actually determining the specialization relationship (more below). Currently this just panics, which is why the tests will fail.
* Using the specialization priority to dispatch associated items correctly. This involves changes to the solver.

---

My current plan for determining specialization is this:

If the overlap solution returned a definite guidance for how to proceed, we can apply that substitutiion in turn to each the lhs and rhs impls. We then attempt to compare the input types for the substituted lhs impl to the original rhs impl, and vice versa to see if they are an exact match (same as intersection but universally quantiifed). If exactly one comparison succeeds, we have a specialization relationship (the unsubstituted impl is the more special).

However, the Substitution returned by the solver is not currently set up to be easily fed back into the system; there's some technical debt here that needs to be resolved.

### DIFF
--- a/src/coherence/mod.rs
+++ b/src/coherence/mod.rs
@@ -29,6 +29,7 @@ impl Program {
         // Record them in the forest by adding an edge from the less special
         // to the more special.
         self.find_specializations(|less_special, more_special| {
+            println!("found edge");
             forest.add_edge(less_special, more_special, ());
         })?;
 
--- a/src/coherence/mod.rs
+++ b/src/coherence/mod.rs
@@ -23,7 +23,6 @@ impl Program {
 
         self.find_specializations(|less_special, more_special| {
             forest.add_edge(less_special, more_special, ());
-            Ok(())
         })?;
 
         Ok(forest.into_graph())
--- a/src/coherence/mod.rs
+++ b/src/coherence/mod.rs
@@ -13,7 +13,7 @@ impl Program {
         // Visit every root in the forest & set specialization
         // priority for the tree that is the root of.
         for root_idx in forest.externals(Direction::Incoming) {
-            self.set_priorities(root_idx, &forest, 0);
+            self.set_priorities(root_idx, &forest, 1);
         }
 
         Ok(())
--- a/src/coherence/mod.rs
+++ b/src/coherence/mod.rs
@@ -5,12 +5,13 @@ use ir::{Program, ItemId};
 
 mod solve;
 
-type Forest = Graph<ItemId, ()>;
 
 impl Program {
     pub fn record_specialization_priorities(&mut self) -> Result<()> {
         let forest = self.build_specialization_forest()?;
 
+        // Visit every root in the forest & set specialization
+        // priority for the tree that is the root of.
         for root_idx in forest.externals(Direction::Incoming) {
             self.set_priorities(root_idx, &forest, 0);
         }
@@ -18,9 +19,15 @@ impl Program {
         Ok(())
     }
 
-    fn build_specialization_forest(&self) -> Result<Forest> {
+    // Build the forest of specialization relationships.
+    fn build_specialization_forest(&self) -> Result<Graph<ItemId, ()>> {
+        // The forest is returned as a graph but built as a GraphMap; this is
+        // so that we never add multiple nodes with the same ItemId.
         let mut forest = DiGraphMap::new();
 
+        // Find all specializations (implemented in coherence/solve)
+        // Record them in the forest by adding an edge from the less special
+        // to the more special.
         self.find_specializations(|less_special, more_special| {
             forest.add_edge(less_special, more_special, ());
         })?;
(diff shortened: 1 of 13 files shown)


nikomatsakis:
Can you rebase this atop the overlap PR so that the diffs are cleaner? Well, I guess it's just the last commit or two?

withoutboats:
@nikomatsakis yep been meaning to

withoutboats:
Rebased

withoutboats:
Okay I've done a bit more setup now, its just the actual "is this impl a specialization of that impl" query left. In particular, we convert the substitution to a `Vec<Parameter>` and then do the substitution.

Tests *should* pass right now, but they don't. Both failures have to do with projections. Will investigate.

withoutboats:
CI results show there's also been a regression in petgraph on nightly ugh

src/ir/mod.rs
```diff
@@ -609,6 +613,21 @@ impl Substitution {
     pub fn is_empty(&self) -> bool {
         self.tys.is_empty() && self.lifetimes.is_empty()
     }
+
+    pub fn into_parameters(self) -> Vec<Parameter> {
```
withoutboats: Would appreciate a review of this function @aturon. In particular it assumes:

* The index of the inference variables is the same index we started with in the query.
* Every variable will have a substitution, so there won't be any gaps.

I don't know if this is true or not but if it isn't I don't see how we could transform this to the subst folder, which takes a slice of parameters and uses their indexes.

withoutboats:
rewrote specialization based on conversation with @aturon, no longer needs to process the solution, instead we generate this query:

```
forall<more_special binders...> {
     if (more_special where_clauses...>) {
         exists<less_special binders...> {
              type_equalities... /\ less_special where_clauses...
         }
     }
}
```

For example:

```rust
impl<U: Clone> Foo for U { }
impl<T: Clone> Foo for Vec<T> { }

forall<T> {
     if (T: Clone) {
          exists<U> {
               Vec<T> = U /\ U: Clone
          }
     }
}
```

withoutboats:
Needs tests also

withoutboats:
Okay this is ready for review & would be green if there weren't a nightly regression affecting petgraph

src/coherence/solve.rs
```diff
@@ -0,0 +1,143 @@
...
+        // over the joined binders. This is our query.
+        let goal = params_goals.chain(wc_goals)
+                    .fold1(|goal, leaf| Goal::And(Box::new(goal), Box::new(leaf)))
+                    .expect("Every trait takes at least one input type")
+                    .quantify(QuantifierKind::Exists, binders);
+
+        self.solve_closed_goal(InEnvironment::empty(goal)).ok().map_or(false, |sol| !sol.cannot_be_proven())
+    }
+
+    fn specializes(&mut self, less_special: &ImplDatum, more_special: &ImplDatum) -> bool {
```
withoutboats: I'll add comments to this method

withoutboats:
petgraph has been updated & hopefully this will be green now

src/coherence/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Build the forest of specialization relationships.
+    fn build_specialization_forest(&self) -> Result<Graph<ItemId, ()>> {
+        // The forest is returned as a graph but built as a GraphMap; this is
+        // so that we never add multiple nodes with the same ItemId.
+        let mut forest = DiGraphMap::new();
+
+        // Find all specializations (implemented in coherence/solve)
+        // Record them in the forest by adding an edge from the less special
+        // to the more special.
+        self.find_specializations(|less_special, more_special| {
```
aturon: I'll pretend to be @nikomatsakis here and ask: why not an iterator?

src/coherence/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        Ok(forest.into_graph())
+    }
+
+    // Recursively set priorities for those node and all of its children.
+    fn set_priorities(&mut self, idx: NodeIndex, forest: &Graph<ItemId, ()>, p: usize) {
+    
+        // Get the impl datum recorded at this node and reset its priority
+        {
+            let impl_id = forest.node_weight(idx).expect("received valid node index");
```
aturon: Minor nit: sadly, the `expect` output doesn't say "expected" so these strings are the reverse of what you want.

src/coherence/solve.rs
```diff
@@ -0,0 +1,143 @@
...
+            for ((&l_id, lhs), (&r_id, rhs)) in impls.into_iter().tuple_combinations() {
+                // Check if the impls overlap, then if they do, check if one specializes
+                // the other. Note that specialization can only run one way - if both
+                // specialization checks return *either* true or false, that's an error.
+                if solver.overlaps(lhs, rhs) {
+                    match (solver.specializes(lhs, rhs), solver.specializes(rhs, lhs)) {
+                        (true, false)   => record_specialization(l_id, r_id),
+                        (false, true)   => record_specialization(r_id, l_id),
+                        (x, y)          => {
+                            println!("{}, {}", x, y);
```
aturon: This wants to be `debug!`

src/coherence/mod.rs
withoutboats: we dont have the yield keyword

src/coherence/solve.rs
withoutboats: this was a mistake! but im going to put some debug prints into this anyway

src/coherence/mod.rs
withoutboats: Can't use `-> impl Iterator` because the iterator has lifetimes from `&self`

Issue #44, opened 2017-06-13
normalization and the meaning of unique


nikomatsakis:
We are still wrestling to some extent with the meaning of "unique" and how to handle negative reasoning. In https://github.com/nikomatsakis/chalk/pull/38, @aturon made a number of changes, but I have some reservations about some of the details of that PR. There are also some scenarions that are still not handled very well in that PR that we may want to consider.

When @aturon and I last spoke, we had decided to shift the approach of the PR in two ways -- at least, if I remember correctly. First, to introduce a `CannotProve` variant, which is used to mean that the system cannot prove a particular goal, and never will be able to, but that the goal **may** hold for **some** instantiations of the universally quantified variables in scope (but it also may **not** hold for **some** instantiations). In particular, this would be returned when you attempt to unify (e.g.) `!1` and `!2` or `!1` and `Vec<T>`. Currently, those unification attempts "succeed" with the "ambiguous" flag set to true, and hence `forall<T,U> { T = U }` yields an ambiguous result. Under this proposal it would yield `CannotProve`.

The meaning of the various return values for proving goal G is thus:

- Unique: There is a unique MGU of the existentials for which G is provable (and here is the corresponding MGU).
- Ambiguous: The goal may be provable in multiple ways, given a suitable instantiation of the existentials. Here is a *suggestion* that can lead you to one of those ways.
    - Definite: If this goal is to ever work, the existentials must take the given values. (edited)
    - Suggested: This is an opinionated suggestion. This is what we think the user *meant* ,but there are other things they could have typed.
    - Unknown: No advice, you're on your own buddy.
- CannotProve: There is no instantiation of the existentials for which we could prove this goal to be true. Nonetheless, the goal may yet be true for some instantiations of the universals. In other words, this goal is neither true nor false.
- Error: This goal is provably false.

In general, the modality controls whether we consider the set of impls/environmental clauses/etc to be a closed world.

[...]

PR #45, opened 2017-06-13
Add `CannotProve` solution

scalexm:
Added a `CannotProve` solution in order to deal with #44. In particular, unifying a *for all* variable with something else automatically leads to `CannotProve` (or error). `CannotProve` solutions have the lowest priority when combining multiple solutions.

Examples 1 and 2 in this issue are now all treated as `CannotProve`. Example 3 is not shown in this PR since this specific issue arises only when elaborating clauses à la #12, but this PR will indeed resolve this ambiguity in the future.

### DIFF
--- a/src/solve/fulfill.rs
+++ b/src/solve/fulfill.rs
@@ -326,6 +326,7 @@ impl<'s> Fulfill<'s> {
             }
 
             self.obligations.extend(obligations.drain(..));
+            debug!("end of round, {} obligations left", self.obligations.len());
         }
 
         // At the end of this process, `self.obligations` should have
--- a/src/solve/fulfill.rs
+++ b/src/solve/fulfill.rs
@@ -23,10 +23,6 @@ impl Outcome {
     }
 }
 
-pub struct UnifyOutcome {
-    pub ambiguous: bool,
-}
-
 /// A goal that must be resolved
 #[derive(Clone, Debug, PartialEq, Eq)]
 enum Obligation {
@@ -53,6 +49,7 @@ struct PositiveSolution {
 enum NegativeSolution {
     Refuted,
     Ambiguous,
+    CannotProve,
 }
 
 /// A `Fulfill` is where we actually break down complex goals, instantiate
@@ -75,10 +72,7 @@ pub struct Fulfill<'s> {
     /// Lifetime constraints that must be fulfilled for a solution to be fully
     /// validated.
     constraints: HashSet<InEnvironment<Constraint>>,
-
-    /// Record that a goal has been processed that can neither be proved nor
-    /// refuted, and thus the overall solution cannot be `Unique`
-    ambiguous: bool,
+    cannot_prove: bool,
 }
 
 impl<'s> Fulfill<'s> {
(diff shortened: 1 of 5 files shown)


src/solve/fulfill.rs
```diff
@@ -75,10 +72,7 @@ pub struct Fulfill<'s> {
     /// Lifetime constraints that must be fulfilled for a solution to be fully
     /// validated.
     constraints: HashSet<InEnvironment<Constraint>>,
-
-    /// Record that a goal has been processed that can neither be proved nor
-    /// refuted, and thus the overall solution cannot be `Unique`
-    ambiguous: bool,
+    cannot_prove: bool,
```
nikomatsakis: Nit: comment would be good

src/solve/fulfill.rs
```diff
@@ -312,24 +307,29 @@ impl<'s> Fulfill<'s> {
...
                     }
                 };
 
                 if ambiguous {
                     debug!("ambiguous result: {:?}", obligation);
                     obligations.push(obligation);
                 }
+
+                // If one of the obligations cannot be proven, then the whole goal
+                // cannot be proven either.
```
nikomatsakis: To make sure I understand: We don't stop immediately here because if we were to encounter a hard error somewhere else, that would "override" this cannot prove result, right?

src/solve/fulfill.rs
```diff
@@ -346,7 +346,12 @@ impl<'s> Fulfill<'s> {
     /// the outcome of type inference by updating the replacements it provides.
     pub fn solve(mut self, subst: Substitution) -> Result<Solution> {
         let outcome = self.fulfill()?;
-        if outcome.is_complete() && !self.ambiguous {
+
+        if self.cannot_prove {
```
nikomatsakis: Instead of having a `cannot_prove` field, can we use a local variable in `fulfill()` and extend the `Outcome` enum? In that case, I think we would replace these two `ifs` with something like:

```rust
match outcome {
    Outcome::CannotProve => return Ok(Solution::CannotProve),
    Outcome::Complete => {
        // No obligations remain, so we have definitively solved our goals,
        ...
        return ...; // as before
    }
    Outcome::Incomplete => { /* fallthrough to figure out some good suggestions */ }
}
```
    }

nikomatsakis: Ah, I see that in this case `unify` would have to return `Outcome` (or `UnifyOutcome`). This way is ok too. It feels a bit less direct to me somehow, but maybe some comments would suffice to clear it up.

src/solve/fulfill.rs
scalexm: Right, I'm changing that.

src/solve/fulfill.rs
scalexm: Yeah exactly. I'm adding a bit of documentation here too. Notice that for the same reason, in the case where the unification set `cannot_prove = true`, we don't return `CannotProve` early here:
[https://github.com/scalexm/chalk/blob/cannot-prove/src/solve/solver.rs#L214](https://github.com/scalexm/chalk/blob/cannot-prove/src/solve/solver.rs#L214)

src/solve/fulfill.rs
scalexm: Yes and moreover `unify` is called outside of `Fulfill` (see previous comment) so `unify` would still have to change an internal state somehow, so I thought that a flag was lighter.

PR #46, opened 2017-06-13
Rename repl to chalki.

withoutboats:
`cargo install chalk` installs a binary called `repl`, which is not great. I've changed the name of the repl to `chalki`.

Arguably this could just be called chalk, but there's a tradition of adding `i` to the end for the repl and I think chalki is cute. Bikeshed away.


nikomatsakis:
Sure, why not? :)

PR #48, opened 2017-06-14
Return early when treating ambiguous cycles

scalexm:
This should solve the problem where an `Ambig(Unknown)` answer was used to feed a cycle and not producing back any answer.

### DIFF
--- a/src/solve/solver.rs
+++ b/src/solve/solver.rs
@@ -183,12 +183,14 @@ impl Solver {
             match self.cycle_strategy {
                 CycleStrategy::Tabling if slot.cycle => {
                     let actual_answer = result.as_ref().ok().map(|s| s.clone());
-                    if actual_answer == answer {
-                        // Fixed point: break
-                        return result;
-                    } else {
-                        answer = actual_answer;
-                    }
+                    let fixed_point = actual_answer == answer;
+                    match (fixed_point, &actual_answer) {
+                        (_, &Some(Solution::Ambig(Guidance::Unknown))) | (true, _) =>
+                            return result,
+                        _ => ()
+                    };
+
+                    answer = actual_answer;
                 }
                 _ => return result,
             };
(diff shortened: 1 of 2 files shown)


src/solve/solver.rs
```diff
@@ -183,12 +183,14 @@ impl Solver {
...
-                        answer = actual_answer;
-                    }
+                    let fixed_point = actual_answer == answer;
+                    match (fixed_point, &actual_answer) {
+                        (_, &Some(Solution::Ambig(Guidance::Unknown))) | (true, _) =>
+                            return result,
+                        _ => ()
+                    };
+
+                    answer = actual_answer;
```
nikomatsakis: is it possible to make a test for this? or is that test basically just in your PR for #12?

src/solve/solver.rs
```diff
@@ -183,12 +183,14 @@ impl Solver {
             match self.cycle_strategy {
                 CycleStrategy::Tabling if slot.cycle => {
                     let actual_answer = result.as_ref().ok().map(|s| s.clone());
-                    if actual_answer == answer {
-                        // Fixed point: break
-                        return result;
-                    } else {
-                        answer = actual_answer;
-                    }
+                    let fixed_point = actual_answer == answer;
```
nikomatsakis: Can you add a comment here explaining why we have this special case? Also, I suspect that you can get into similar problems when you *have* guidance as well, no?

scalexm: I suspect it as well although ambig with guidance can perform some unification, but I cannot find examples where a cycle gives ambig with guidance after being fed. Nonetheless stopping when we get `Ambig` regardless of the guidance should not affect correctness of the final answer.

src/solve/solver.rs
scalexm: Well that's funny because in fact if I provide a test for that and deactivate the fix, the test program will not even load because of the overlap check which will loop forever. So I was indeed relying on this behaviour to test the fix in my PR for #12 (the `trait_wf` test would not load without the fix). Still I added a general test in this PR, which is better for documentation.

PR #49, opened 2017-06-14
Elaborating where clauses directly in the logic

scalexm:
These commits aim at implementing the additional rules described in #12 for elaborating where clauses
directly in the logic. Basically, each where clause of the form `T: Foo` expands to `T: Foo, WF(T: Foo)`
and each where clause of the form `T: Foo<Item = U>` expands to `T: Foo<Item = U>, T: Foo, WF(T: Foo)`.
We also added reverse rules on traits and associated types, and well-formedness checking of struct
parameters and fields.

Also, the `where_clauses` field on `AssociatedTyDatum` was removed since this can be emulated by writing:
```rust
trait Item where <Self as Item>::Out: Eq {
    type Out;
}
```

## Traits
The rules generated by a trait declaration are:
```rust
trait Bar { }
// WF(?Self: Bar) :- WF(?Self)

trait Foo<T> where T: Bar { }
// WF(?Self: Foo<?T>) :- WF(?Self), WF(?T), (?T: Bar), WF(?T: Bar)

// "Reverse" rules
// (?T: Bar) :- WF(?Self: Foo<?T>)
// WF(?T: Bar) :- WF(?Self: Foo<?T>)
```

## Structs
The rules generated by a struct declaration are:
```rust
trait Clone { }
struct Bar { }

struct Foo<T, U> where T: Clone { }
// WF(Foo<?T, ?U>) :- WF(?T), (T: Clone), WF(?T: Clone)
```

## Impls
The rules generated by an impl declaration are:
```rust
trait Clone { }
trait Foo<T> { }
struct X { }

impl Foo<T> for X where T: Clone { }
// (X: Foo<?T>) :- (?T: Clone), WF(?T: Clone)
// no need to check WF(?T) because this will be required by WF(X: Foo<?T>)
```

## Associated types
The rules generated by an associated type definition are:
```rust
trait Foo {
    type Assoc;
}
// <?T as Foo>::Assoc ==> (Foo::Assoc)<?T> :- (?T: Foo)
// forall<U> { (?T: Foo) :- <?T as Foo>::Assoc ==> U }
```

### DIFF
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -28,10 +28,14 @@ pub Goal: Box<Goal> = {
 Goal1: Box<Goal> = {
     "forall" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::ForAll(p, g)),
     "exists" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::Exists(p, g)),
-    "if" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, true)),
-    "if_raw" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, false)),
+    <i:IfKeyword> "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, i)),
     "not" "{" <g:Goal> "}" => Box::new(Goal::Not(g)),
     <w:WhereClause> => Box::new(Goal::Leaf(w)),
+};
+
+IfKeyword: bool = {
+    "if" => true,
+    "if_raw" => false,
 };
 
 StructDefn: StructDefn = {
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -28,7 +28,8 @@ pub Goal: Box<Goal> = {
 Goal1: Box<Goal> = {
     "forall" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::ForAll(p, g)),
     "exists" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::Exists(p, g)),
-    "if" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g)),
+    "if" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, true)),
+    "if_raw" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, false)),
     "not" "{" <g:Goal> "}" => Box::new(Goal::Not(g)),
     <w:WhereClause> => Box::new(Goal::Leaf(w)),
 };
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -1,4 +1,5 @@
 use ir::*;
+use std::marker::PhantomData;
 
 pub trait Cast<T>: Sized {
     fn cast(self) -> T;
@@ -130,7 +131,7 @@ impl<T, U> Cast<Vec<U>> for Vec<T>
     where T: Cast<U>
 {
     fn cast(self) -> Vec<U> {
-        self.into_iter().map(|v| v.cast()).collect()
+        self.into_iter().casted().collect()
     }
 }
 
@@ -143,5 +144,35 @@ impl Cast<Parameter> for Ty {
 impl Cast<Parameter> for Lifetime {
     fn cast(self) -> Parameter {
         ParameterKind::Lifetime(self)
+    }
+}
+
+pub struct Casted<I, U> {
+    iterator: I,
+    _cast: PhantomData<U>,
+}
+
+impl<I: Iterator, U> Iterator for Casted<I, U> where I::Item: Cast<U> {
+    type Item = U;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iterator.next().map(|item| item.cast())
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iterator.size_hint()
+    }
+}
+
+pub trait Caster<U>: Sized {
+    fn casted(self) -> Casted<Self, U>;
+}
+
+impl<I: Iterator, U> Caster<U> for I {
+    fn casted(self) -> Casted<Self, U> {
+        Casted {
+            iterator: self,
+            _cast: PhantomData,
+        }
     }
 }
--- a/src/lower/mod.rs
+++ b/src/lower/mod.rs
@@ -918,13 +918,11 @@ impl ir::StructDatum {
     fn to_program_clauses(&self, program: &ir::Program) -> Vec<ir::ProgramClause> {
         // Given:
         //
-        //    struct Foo<T: Eq> {
-        //        field: Bar
-        //    }
+        //    struct Foo<T: Eq> { }
         //
         // we generate the following clause:
         //
-        //    for<?T> WF(Foo<?T>) :- WF(?T), WF(Bar), (?T: Eq), WF(?T: Eq).
+        //    for<?T> WF(Foo<?T>) :- WF(?T), (?T: Eq), WF(?T: Eq).
 
         let wf = ir::ProgramClause {
             implication: self.binders.map_ref(|bound_datum| {
@@ -939,17 +937,12 @@ impl ir::StructDatum {
                                              .cloned()
                                              .map(|ty| ir::WellFormed::Ty(ty).cast());
 
-                        let fields = bound_datum.fields
-                                                .iter()
-                                                .cloned()
-                                                .map(|ty| ir::WellFormed::Ty(ty).cast());
-
                         let where_clauses = bound_datum.where_clauses.iter()
                                                        .cloned()
                                                        .flat_map(|wc| wc.expanded(program))
                                                        .map(|wc| wc.cast());
 
-                        tys.chain(fields).chain(where_clauses).collect()
+                        tys.chain(where_clauses).collect()
                     }
                 }
             }),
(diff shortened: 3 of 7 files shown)


src/lower/mod.rs
```diff
@@ -909,24 +910,42 @@ impl Anonymize for [ir::ParameterKind<ir::Identifier>] {
...
         //
-        //    struct Foo<T: Eq> { ... }
+        //    struct Foo<T: Eq> {
+        //        field: Bar
+        //    }
         //
         // we generate the following clause:
         //
-        //    for<?T> WF(Foo<?T>) :- (?T: Eq).
+        //    for<?T> WF(Foo<?T>) :- WF(?T), WF(Bar), (?T: Eq), WF(?T: Eq).
```
nikomatsakis: Hmm, I don't think we want to include `WF(Bar)` in this list.  That is, I think that `WF(Foo<T>)` is just saying if a **reference** to the type `Foo<T>` is well-formed. Not if the struct **definition** is well-formed -- for the latter, I would expect us to generate some rules like:

```
StructDefnWf(Foo) :-
    for<T> { if (T: Eq) { WF(Bar) } }.
```

src/solve/test.rs
```diff
@@ -175,7 +175,7 @@ fn prove_forall() {
         // Here, we do know that `T: Clone`, so we can.
         goal {
             forall<T> {
-                if (T: Clone) {
+                if (WellFormed(T: Clone), T: Clone) {
```
nikomatsakis: Man. I have to review the plan. I sort of forget why we need the WF's again. ;)

src/solve/test.rs
```diff
@@ -536,21 +536,47 @@ fn elaborate_normalize() {
...
+            impl Eq for i32 { }
+            impl Item for i32 {
+                type Out = i32;
+            }
         }
 
         goal {
             forall<T, U> {
-                if (T: Item<Out = U>) {
+                if (WellFormed(T: Item), T: Item<Out = U>) {
```
nikomatsakis: why is this `WF(T: Item)` and not `WF(T: Item<Out = U>)`? Because the former suffices?

src/lower/mod.rs
scalexm: I'm not sure to understand, when would we use `WF(Foo<T>)` and ˋStructDefnWF`?

src/solve/test.rs
scalexm: Yeah I was waiting for you review here

src/solve/test.rs
scalexm: Here indeed the former suffices. At first I thought we didn't need something like WF(Projection). But now that you say it, there's indeed no way to have something like that:
```rust
trait Foo where <Self as Foo>::Item: Eq {
    type Item;
}

if (WellFormed(T: Foo<Item = Bar>)) { ... }
```
Currently, we have to add ```T: Foo<Item = Bar>``` into the environment in order to trigger the normalization, and rely on ```WF(T: Foo)```. But as a consequence we are also pushing ```T: Foo``` into the environnement (it is implied by ```T: Foo<Item = Bar>```) so basically we are making a stronger hypothesis that initially wanted.
It may be not so difficult to fix however.

src/lower/mod.rs
nikomatsakis: We would never use `StructDefnWF`. Rather, `StructDefnWF` would be one of the things we must prove to decide if the whole program type-checks. This is basically a modularity thing -- it's the same as saying that if you have some function `foo`, we just need its *signature* to type-check the callers, but then we separately type-check the body of `foo` against its signature. 

Here, the idea is that, for some struct S, the users of S need not know the types of its fields; they can assume that if the type S is well-formed, then the types of its fields are well-formed.

nikomatsakis:
Sorry for delay @scalexm ! Will try to wrap this up ASAP.

src/lower/mod.rs
scalexm: Yeah ok I see!

src/solve/test.rs
scalexm: Basically given:
```rust
trait Clone { }
impl<T> Clone for Vec<T> where T: Clone { }
```
we need to generate `(Vec<T>: Clone) :- (T: Clone), WF(T: Clone)`

because of this example:
```rust
trait Clone where Self: Marker { }
impl<T> Clone for Vec<T> where T: Clone { }
```
here we need to have `WF(T: Clone)` in order to have `Vec<T>: Clone` because `WF(T: Clone)` implies that `T` correctly implements `Marker` whereas this is not the case with the assumption `T: Clone` alone

scalexm:
We also need more reverse rules, like:
```rust
struct S<T> where T: Clone { }
// current:
// WF(S<T>) :- T: Clone, WF(T: Clone)
//
// should add:
// (T: Clone) :- WF(S<T>)
// WF(T: Clone) :- WF(S<T>)

impl<T> Foo for T { }
// current:
// WF(T: Foo) :- WF(T)
//
// should add:
// WF(T) :- WF(T: Foo)
```

src/lower/mod.rs
```diff
@@ -791,14 +787,19 @@ impl ir::ImplDatum {
...
         ir::ProgramClause {
             implication: self.binders.map_ref(|bound| {
                 ir::ProgramClauseImplication {
                     consequence: bound.trait_ref.clone().cast(),
-                    conditions: bound.where_clauses.clone().cast(),
+                    conditions: bound.where_clauses
+                                     .iter()
+                                     .cloned()
+                                     .flat_map(|wc| wc.expanded(program))
+                                     .map(|wc| wc.cast())
```
nikomatsakis: apropos of nothing, we should make a `.casted()` adapter just to make this cuter :)

src/solve/test.rs
nikomatsakis: Yes. So I wonder if it's worth having two forms -- e.g., `if` and `if-raw` or something -- where the former "elaborates" so that `T: Clone` is "short" for `WF(T: Clone), T: Clone`, just as it would be in an impl. It seems useful to be able to suppress that elaboration, but I imagine that we'll almost always want it to happen, no?

src/lower/mod.rs
scalexm: Yeah right!

src/solve/test.rs
scalexm: I thought about this indeed, and I'm in favor of adding an `if_raw`.

chalk-parse/src/parser.lalrpop
```diff
@@ -28,7 +28,8 @@ pub Goal: Box<Goal> = {
 Goal1: Box<Goal> = {
     "forall" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::ForAll(p, g)),
     "exists" "<" <p:Comma<ParameterKind>> ">" "{" <g:Goal> "}" => Box::new(Goal::Exists(p, g)),
-    "if" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g)),
+    "if" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, true)),
+    "if_raw" "(" <w:Comma<WhereClause>> ")" "{" <g:Goal> "}" => Box::new(Goal::Implies(w, g, false)),
```
scalexm: @nikomatsakis, do you have an idea of how I could factorize these two lines?

src/lower/mod.rs
```diff
@@ -724,8 +720,19 @@ impl<'k> LowerGoal<Env<'k>> for Goal {
...
+                    where_clauses = ir::with_current_program(|program| {
+                        let program = program.expect("cannot elaborate without a program");
+                        where_clauses.into_iter()
+                                     .flat_map(|wc| wc.expanded(program))
+                                     .casted()
+                                     .collect()
+                    });
+                }
+                Ok(Box::new(ir::Goal::Implies(where_clauses, g.lower(env)?)))
+            }
```
scalexm: @nikomatsakis not sure about using `ir::with_current_program`, basically it prevents from having higher ranked bounds like `where if (I: Iterator<Item = U>) { U: Eq }` (but we don't support them).

chalk-parse/src/parser.lalrpop
nikomatsakis: not sure you need to, but you could do:

```
Goal1: Box<Goal> = {
    ...
    <i:IfKeyword> ... => Box::new(Goal::Implies(w, g, i)),
};

IfKeyword: bool = {
    "if" => true,
    "if_raw" => false,
};
```

nikomatsakis: (if that generates conflicts, you can add `#[inline]` onto the `IfKeyword`, but I can't see why it would)

scalexm: Great, that's what I was looking for

src/cast.rs
```diff
@@ -130,7 +131,7 @@ impl<T, U> Cast<Vec<U>> for Vec<T>
     where T: Cast<U>
 {
     fn cast(self) -> Vec<U> {
-        self.into_iter().map(|v| v.cast()).collect()
+        self.into_iter().casted().collect()
```
nikomatsakis: Nice :)

src/lower/mod.rs
nikomatsakis: hmm -- seems ok for now -- we can revisit at a later point if we ever care, I think.

src/solve/test.rs
```diff
@@ -175,13 +175,23 @@ fn prove_forall() {
...
-                if (WellFormed(T: Clone), T: Clone) {
+                if (T: Clone) {
                     Vec<T>: Clone
                 }
             }
         } yields {
             "Unique; substitution [], lifetime constraints []"
         }
+
+        goal {
```
nikomatsakis: maybe good to add a comment here that we are explicitly *not* "knowing" that `WF(T: Clone)`, right?

src/solve/test.rs
```diff
@@ -780,12 +792,18 @@ fn trait_wf() {
...
+        }
+
+        goal {
             Slice<Int>: Eq<Slice<Int>>
         } yields {
             "No possible solution"
         }
 
-        // not WF because previous equation doesn't hold
+        // not WF because previous equation doesn't hold, despite Slice<Int> having an impl for Ord<Int>
```
nikomatsakis: 👍

src/lower/mod.rs
nikomatsakis: Really all we need is enough data to know how many parameters are declared on trait vs associated types. I think eventually we are going to have to refactor to support a more general "query-style" lowering, like rustc does, which should enable us to uncover *that* data without full access to other things. But, as you say, for now we don't support hypotheses in user-defined goals anyhow... (though it seems like we may want that eventually)

nikomatsakis:
Rebased, added comment.

PR #50, opened 2017-06-15
bump to the latest version of LALRPOP

nikomatsakis:
I think this will fix travis.


nikomatsakis:
Oh, I guess the problem is really https://github.com/bluss/petgraph/pull/154

Issue #41, opened 2017-06-05
Improve documentation


fabric-and-ink:
- [x] Introduce a glossary at the root of the project with pointers to literature (#51) .

The following items are on hold until I learned more about logic and types. This thing is still like alien technology for me 😀 
- ~~"Massage-in" the essence of the wonderful blog posts by Niko and Aaron.~~
- ~~`#![warn(missing_docs)]`.~~

aturon:
Hah, love it! All good things to do.

tomwhoiscontrary:
It would be unimportant but nice to specify the desired pronunciation of the name. Is it the English word "chalk", or the Greek "χαλκ-", which, based on my rusty Ancient Greek, is more like "halk", rhyming with "talc"?

Also, Chalkidiki is a peninsula, rather than an island. Although it was temporarily and island after a [Persian megalomaniac dug a big trench across it](https://en.wikipedia.org/wiki/Xerxes_Canal).

fabric-and-ink:
I plan to start a mdbook and will open a separate issue for that when I am ready.

PR #51, opened 2017-06-21
Add glossary

fabric-and-ink:
This a draft for a glossary for this crate (see #41). I chose terms that I found inside the crate and added some that seemed appropriate.

Now I need feedback, @nikomatsakis, @aturon, @withoutboats :)

- Which terms are unnecessary? Which terms are still missing from the list? Which terms need more explanation?
- Some terms are still lacking an description. These are the ones that I would have to guess.
- Please point out all factual & grammatical errors and wording improvements.
- More pointers to literature :)

[Rendered](https://github.com/fabric-and-ink/chalk/blob/add_glossary/GLOSSARY.md)

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -1,0 +1,203 @@
+# Glossary
+This is a glossary of terminology (possibly) used in the chalk crate.
+
+## Binary connective
+There are sixteen logical connectives on two boolean variables. The most
+interesting in this context are listed below. There is also a truth table given
+which encodes the possible results of the operations like this
+
+```
+f(false, false) f(false, true) f(true, false) f(true, true).
+```
+
+As a shorthand the resulting truth table is encoded with `true = 1` and `false =
+0`.
+
+| Truth table | Operator symbol | Common name                      |
+|-------------|-----------------|----------------------------------|
+| 0001        | &&              | Conjunction; and                 |
+| 1001        | <=>             | Equivalence; if and only if; iff |
+| 1101        | =>              | Implication; if ... then         |
+
+## Binder
+A binder is an expression that binds a literal to a certain expression.
+Examples for binders:
+
+- The universal quantifier `forall(a)` states that a certain condition holds for
+  all allowed values for `a`.
+- A function definition `f(x) = a * x` is a binder for the variable `x` whereas
+  `a` is a free variable.
+- A sum `\sum_n x_n` binds the index variable `n`.
+
+## Canonical Form
+A formula in canonical form has the property that its DeBruijn indices are
+minimized. For example when the formula `forall<0, 1> { 0: A && 1: B }` is
+processed, both "branches" `0: A` and `1: B` are processed individually. The
+first branch would be in canonical form, the second branch not since the
+occurring DeBruijn index `1` could be replaced with `0`.
+
+## Clause
+In the A clause is the disjunction of several expressions. For example the clause
+`condition_1 || condition_2 || ...` states that at least one of the conditions
+holds.
+
+There are two notable special cases of clauses. A *Horn clause* has at most one
+positive literal. A *Definite clause* has exactly one positive literal.
+
+*Horn clauses* can be written in the form `A || !B || !C || ...` with `A` being
+the optional positive literal. Due to the equivalence `(P => Q) <=> (!P || Q)`
+the clause can be expressed as `B && C && ... => A` which means that A is true
+if `B`, `C`, etc. are all true. All rules in chalk are in this form. For example
+
+```
+struct A<T> {}
+impl<T> B for A<T> where T: C + D {}
+```
+
+is expressed as the *Horn clause* `(T: C) && (T: D) => (A<T>: B)`. This formula
+has to hold for all values of `T`. The second example
+
+```
+struct A {}
+impl B for A {}
+impl C for A {}
+```
+
+is expressed as the *Horn clause* `(A: B) && (A: C)`. Note the missing
+consequence.
+
+## DeBruijn Index
+DeBruijn indices numerate literals that are bound in an unambiguous way. The
+literal is given the number of its binder. The indices start at zero from the
+innermost binder increasing from the inside out.
+
+Given the example `forall<T> { exists<U> { T: Foo<Item=U> } }` the
+literal names `U` and `T` are replaced with `0` and `1` respectively: `forall<0>
+{ exists<1> { 0: Foo<Item=1> } }`.
+
+## Formula
+A formula is a logical expression consisting of literals and constants connected
+by logical operators.
+
+## Goal
+With a set of type variables, given types, traits and impls, a goal specifies a
+problem which is solved by finding types for the type variables that satisfy the
+formula. For example the goal `exists<T> { T: u32 }` can be solved with `T =
+u32`.
+
+## Literal
+A literal is an atomic element of a formula together with the constants `true`
+and `false`. It is equivalent to a variable in an algebraic expressions. Note
+that literals are *not* the same as the type variables used in specifying a
+goal.
+
+## Normal form
+To say that a statement is in a certain *normal form* means that the pattern in
+which the subformulas are arranged fulfil certain rules. The individual patterns
+have different advantages for their manipulation.
+
+### Conjunctive normal form (CNF)
+A formula in CNF is a disjunction of conjunctions. For example `(x1 || x2 ||
+x3) && (x4 || x5 || x6)` is in CNF.
+
+### Disjunctive normal form (DNF)
+A formula in DNF is a conjunction of disjunctions. For example `(x1 && x2 &&
+x3) || (x4 && x5 && x6)` is in DNF.
+
+### Negation normal form (NNF)
+A formula in NNF consists only of literals, the connectives `&&` and `||` and
+`true` and `false`.
+
+### Prenex normal form (PNF)
+All quantifiers are on the highest level of a formula and do not occur inside
+the subformulas of the expression.
+
+- `forall(x). exists(y). forall(z). P(x) && P(y) => P(z)` is in PNF.
+- `(exists(x). P(x)) => exists(y). P(y) && forall(z). P(z)` is *not* in PNF.
+
+## Normalization
+Normalization is the process of converting an associated type to a concrete
+type. In the case of an iterator this would mean that the associated `Item` type
+is replaced with something more meaningful with respect to the individual
+context (e.g. `u32`).
+
+## Projection
+Projection is the reference to a field or (in the context of Rust) to a type
+from another type.
+
+## Satisfiability
+A formula is satisfiable iff there is a valuation for the atoms inside the
+formula that makes it true.
+
+## Unification
+Unification is the process of solving a formula. That means unification finds
+values for all the free literals of the formula that satisfy it. In the context
+of chalk the values refer to types.
+
+## Universe
+A universe sets the scope in which a particular variable name is bound. (See
+*Binder*.) A universe can encapsulate other universes. A universe can
+be contained by only one parent universe. Universes have therefore a tree-like
+structure. A universe can access the variable names of itself and the parent
+universes but not of the sibling universes.
+
+## Well-formed
+A formula is well-formed if it is constructed according to a predefined set of
+syntactic rules.
+
+In the context of the Rust type system this means that basic rules for type
+construction have to be met. Two examples: 1) Given a struct definition
+
+```rust
+struct HashSet<T: Hash>
+```
+then a type `HashSet<i32>` is well-formed since `i32` implements `Hash`. A type
+`HashSet<NoHash>` with a type `NoHash` that does not implement the `Hash` trait
+is not well-formed.
+
+2) If a trait demands by its definition the implementation of further traits
+for a certain type then these secondary traits have to be implemented as well.
+If a type `Foo` implements `trait Eq: PartialEq` then this type has to implement
+`trait PartialEq` as well. If it does not, then the type `Foo: Eq` is not well
+formed according to Rust type building rules.
+
+## Quantifier
+
+### Existential quantifier
+A formula with the existential quantifier `exists(x). P(x)` is satisfiable if
+and only if there exists at least one value for all possible values of x which
+satisfies the subformula `P(x)`.
+
+In the context of chalk, the existential quantifier usually demands the
+existence of exactly one instance (i.e. type) that satisfies the formula (i.e.
+type constraints). More than one instance means that the result is ambiguous.
+
+### Universal quantifier
+A formula with the universal quantifier `forall(x). P(x)` is satisfiable
+if and only if the subformula `P(x)` is true for all possible values for x.
+
+### Helpful equivalences
+- `not(forall(x). P(x)) <=> exists(x). not(P(x))`
+- `not(exists(x). P(x)) <=> forall(x). not(P(x))`
+
+## Skolemization
+Skolemization is a technique of transferring a logical formula with existential
+quantifiers to a statement without them. The resulting statement is in general
+not equivalent to the original statement but equisatisfiable.
+
+## Validity
+An argument (*premisses* therefore *conclusion*) is valid iff there is no
+valuation which makes the premisses true and the conclusion false.
+
+Valid: `A && B therefore A || B`. Invalid: `A || B therefore A && B` because the
+valuation `A = true, B = false` makes the premiss true and the conclusion false.
+
+## Valuation
+A valuation is an assignment of values to all variables inside a logical
+formula.
+
+# Literature
+- Offline
+    - "Introduction to Formal Logic", Peter Smith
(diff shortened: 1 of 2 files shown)


GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+0`.
+
+```
+| Truth table | Operator symbol | Common name                      |
+|-------------|-----------------|----------------------------------|
+| 0001        | &&              | Conjunction; and                 |
+| 0111        | ||              | Disjunction; or                  |
+| 1001        | <=>             | Equivalence; if and only if; iff |
+| 1101        | =>              | Implication; if ... then         |
+```
```
fabric-and-ink: This table could not be rendered correctly in markdown because of the `||` operator and I didn't like the workaround :(

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+A binder is an expression that binds a variable to a certain expression.
+Examples for binders:
+
+- The universal quantifier `forall(a)` states that a certain condition holds for
+  all allowed values for `a`.
+- A function definition `f(x) = a * x` is a binder for the variable `x` whereas
+  `a` is a free variable.
+- A sum `\sum_n x_n` binds the index variable `n`.
+
+## Canonical
```
fabric-and-ink: "(empty)" missing

GLOSSARY.md
scalexm: Since the `||` operator is not used at all in Chalk, maybe you could just drop it? I think that would avoid some confusion. Or maybe you just want to be exhaustive in a *general* way and then that's fine.

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+
+There are two notable special cases of clauses. A *Horn clause* has at most one
+positive literal. A *Definite clause* has exactly one positive literal.
+
+## Datum
+(empty)
+
+## DeBrujin Index
+(empty)
+
```
scalexm: Once you have filled this section, you may want to move the Canonical section just after this one, since in particular Canonical deals with rebounding variables to lower DeBruijn indices.

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+`condition_1 || condition_2 || ...` states that at least one of the conditions
+holds.
+
+There are two notable special cases of clauses. A *Horn clause* has at most one
+positive literal. A *Definite clause* has exactly one positive literal.
+
+## Datum
+(empty)
+
+## DeBrujin Index
```
scalexm: Nit: Debruijn :)

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Canonical
+## Clause
+In the A clause is the disjunction of several expressions. For example the clause
+`condition_1 || condition_2 || ...` states that at least one of the conditions
+holds.
+
+There are two notable special cases of clauses. A *Horn clause* has at most one
+positive literal. A *Definite clause* has exactly one positive literal.
+
```
scalexm: Maybe you should explain how Horn clauses relate to Chalk and logic programming in general (ie they can be understood as `A <= B && C && D ... && Z` where `A, ..., Z` are literals)

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+(empty)
+
+## Formula
+A formula is a logical expression consisting of literals and constants connected
+by logical operators.
+
+## Goal
+With a set of given types, traits and impls, a goal specifies a problem where
+types need to be found that satisfy the problem.
+
```
scalexm: Maybe add an example of such a goal?

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+- `forall(x). exists(y). forall(z). P(x) && P(y) => P(z)` is in PNF.
+- `(exists(x). P(x)) => exists(y). P(y) && forall(z). P(z)` is *not* in PNF.
+
+## Normalization
+Normalization is the process of converting an associated type to a concrete
+type.
+
+## Projection
+Projection is the reference to a field or (in the context of Rust) to a type.
+
```
scalexm: You might want to give an example of what we call a projection in Chalk.

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+## Unification
+Unification is the process of solving a formula. That means unification finds
+values for all the free variables of the formula that satisfy it. In the context
+of chalk the values refer to types.
+
```
scalexm: An example here would be valuable :)

GLOSSARY.md
```diff
@@ -0,0 +1,128 @@
...
+(empty)
+
+## Quantifier
+(empty)
+
+### Existential quantifier
+A formula with the existential quantifier `exists(x). P(x)` is satisfiable if
+and only if there exists at least one value for all possible values of x which
+satisfies the subformula `P(x)`.
+
```
scalexm: You should add that in the context of Chalk, we actually ask for something stronger than "at least one": we want that there exists exactly one since we are trying to determine what type the user intended. If we have more than one possible values, we consider the result as ambiguous.

GLOSSARY.md
fabric-and-ink: Exhaustiveness is not my goal, so I will drop it.

GLOSSARY.md
fabric-and-ink: I would like to keep the terms in alphabetical order. Can you explain both terms and how they interact?

GLOSSARY.md
fabric-and-ink: Ah... yes :)

GLOSSARY.md
fabric-and-ink: This is an important point that I didn't really grasp. Could you elaborate a bit?

GLOSSARY.md
fabric-and-ink: Will do.

GLOSSARY.md
fabric-and-ink: Will do. I think the Iterator/Item case should be a good example, right?

GLOSSARY.md
fabric-and-ink: Ok.

GLOSSARY.md
fabric-and-ink: Ah, this is important :) and did not really know that. I will add this information.

GLOSSARY.md
scalexm: Oh right it's a glossary!
So I'll start with De Bruijn indices anyway. When you have a formula: `forall<T> { exists<U> { T: Foo<Item = U> } }`, you don't want to deal with variable names like `T` and `U` because these are difficult to handle. So De Bruijn indices transform each variable name into a natural number, for example the previous query would be transformed in `forall<0> { exists<1> { 0: Foo<Item = 1> } }`.

GLOSSARY.md
scalexm: In propositional calculus, a Horn clause is a clause of the form `A || !B || ... || !Z` with at most one positive literal (denoted `A` here). Taking the implication `P => Q` as a shortcut for `!P || Q`, we can rewrite the clause above like `B && C && ... && Z => A`.

So in logic programming, a Horn clause is understood as this latter form, ie exactly one consequence following from a certain number of conditions. Every logical rule that Chalk knows is of this form, and it will only use this kind of rules in order to give a result to a goal.

GLOSSARY.md
scalexm: Yes indeed :)

GLOSSARY.md
scalexm: So Canonical applies to an object which contains variables (e.g., a goal) and means that:
- we have replaced some variables with their value if any (i.e. we have applied our current "knowledge")
- we have rebound De Bruijn indices to lower ones

Chalk only deals with goals in canonical form.

For example, when dealing with a goal like: `forall<0,1> { 0: Foo && 1: Bar }`. Here we have two different indices because there are two different variables, and this goal is in canonical form.

In order to process it we will process each subgoal (namely `0: Foo` and `1: Bar`) separately, and so when we process `1: Bar`, this goal is not in canonical form because it could be written as `0: Bar` (lower De Bruijn indices). So we canonicalize this subgoal, process it, and then feed back the answer to the top level goal.

GLOSSARY.md
fabric-and-ink: Neat!

GLOSSARY.md
fabric-and-ink: Ok, then these indices aren't really more complicated than what I have read. I wasn't sure if there was some small, magical detail that I was missing :)

Reading the code I noticed that these indices seem to have a close relationship to the `UniverseIndex` (not sure about the exact name right now). What is an Universe and how is it related to our formulas?

GLOSSARY.md
fabric-and-ink: More questions came up.

- To me *Horn clauses* seem to be of a "weaker" form than *Definite clauses*, at least with respect to the logical equivalence `B && C && ... => A`. In the case of *Horn clauses* it is possible that there is no `A` as a consequence of `B && C && ...`. How is this possibility treated in chalk? I such a case rejected?
- You mentioned *rules* in your last sentence. Could you explain what these rules are exactly and give some examples?

Thanks!

scalexm: Sorry, I should have been more explicit indeed! So in Chalk we don't really deal with Horn clauses without a consequence (called negative Horn Clauses). Well in some sense *goals* can be seen as such clauses, but that's it.

By *rules*, I meant the clauses that Chalk infers from a (pseudo) Rust program and definitely holds for true. For example, given the following program:
```rust
struct MyStruct<T> { }
impl<T> Foo for MyStruct<T> where T: Clone, T: Bar { }
```
Chalk will infer something like this (in form of a Horn clause):
`(T: Clone) && (T: Bar) => (MyStruct<T>: Foo)`
which in logic programming is traditionally denoted as:
`(MyStruct<T>: Foo) :- (T: Clone), (T: Bar)`
i.e. "consequence IF(:-) condition1, conditon2, ..."

You'll note that since there is a type parameter `T`, the latter clause should hold for every type `T`. So that means that in fact we do not have one clause here, but an infinite family of them (one for each possible value of `T`).

Also note that there can be no conditions at all, like in:
```rust
struct MyStruct { }
impl Foo for MyStruct { }
```
which will be lowered into `(MyStruct: Foo).`, i.e. this is a ground fact that Chalk holds for true.

fabric-and-ink: I see! Thank you for the explanation.

fabric-and-ink:
The following descriptions are still missing/incomplete (and I need help with them because I have only a vague or no clue :) )

- [ ] Datum
- [ ] Universe
- [ ] Well-formed
- [ ] Projection

lqd:
@fabric-and-ink you mean a link to your branch's glossary.md like this ? [Rendered](https://github.com/fabric-and-ink/chalk/blob/add_glossary/GLOSSARY.md)

fabric-and-ink:
Exactly, thanks!

GLOSSARY.md
fabric-and-ink: I'm a bit unsure about *Projection*... I understand what *normalization* is and added an example accordingly. Can you provide an example for this one?

scalexm: So these two are related. A projection is a reference to an associated type, like `<T as Iterator>::Item`. Normalization is the process of telling what this associated type is (specifying the projection), which is denoted in Chalk by `<T as Iterator>::Item ==> i32` for example

GLOSSARY.md
scalexm: Sorry I missed this one! So `UniverseIndex` indicates the number of universall quantifiers we are within. Examples:
```
i32: Foo // in universe 0 here
```

```
forall<T> {
    // in universe 1 here
    exists<U> {
        forall<V> {
            // in universe 2 here
            ...
        }
    }
}
```
When we canonicalize, we just recall the universe indices of the re-bound variables.

nikomatsakis:
> Datum

This is just the singular of "data". It is used to indicate the "data" about an impl or trait, as I recall, in the IR, but has no *particular* meaning other than being relatively distinctive. In general, I think we're phasing most of those "datums" out anyway, since they were basically used to "cheat" in the trait handling, by looking at the original source program rather than the pure lowered form.

> Universe

When you are type-checking something like `fn foo<A>(x: A) { ... }`, we say that the name `A` is in a different "universe" from the structs. This is because it is a name that is only valid inside the body of `foo`, and it can never be named from outside `foo`. Hence "universe" refers to kind of "the set of things you can talk about at a particular point". Universes are arranged into a tree: things in the root universe can be named from anywhere, because if you are in universe X, you can name all the things in X or any ancestor of X, but not the other universes.

> Well-formed

The "well-formed" conditions are basic conditions that are needed to ensure that some request is valid. So, for example, if I have `struct HashSet<T: Hash>`, then `HashSet<i32>` is well-formed because `i32: Hash`. But something like `HashSet<NotHash>`, where `NotHash` is a type that does not implement `Hash`, would not be well-formed. Similarly, if you have `trait Foo<T: Hash>`, then `Bar: Foo<i32>` is *well-formed* (written `WF(Bar: Foo<i32>)`), because `i32: Hash`, but `Bar: Foo<NotHash>` would not be. Note that just because `Bar: Foo<i32>` is *well-formed* doesn't mean that it is *true* -- it basically means that the conditions are correct such that `Bar: Foo<i32>` *could* be implemented, but not that it necessarily is.

As a more concrete example, in the Rust stdlib we have `trait Eq: PartialEq`, which is short for `trait Eq where Self: PartialEq`. This implies that if you have `impl Eq for Foo`, you must *also* have `impl PartialEq for Foo`. We say then that `WF(Foo: Eq)` holds if `Foo: PartialEq` -- that is, those are the conditions in which one is *allowed* to implement `Eq` for `Foo`, even if you do not actually do so.

> Projection

A "projection" is a general term meaning to extract some "part" of something. In the case of chalk, it mostly refers to associated types. So when you have `<T as Iterator>::Item`, this is "projecting" `Item` "out of" `T as Iterator`. That is, it is determining what value is given for `Item` in the impl of `Iterator` for `T`.

GLOSSARY.md
```diff
@@ -0,0 +1,167 @@
...
+## Datum
+(empty)
+
+## DeBruijn Index
+DeBruijn indices are used to numerate free and bound literals inside a formula
+in an unambiguous way in order to work with numbers instead of the names of the
+literals. Given the example `forall<T> { exists<U> { T: Foo<Item=U> } }` the
+literal names `U` and `T` are replaced with `0` and `1` respectively: `forall<0>
+{ exists<1> { 0: Foo<Item=1> } }`.
```
nikomatsakis: Maybe: "More generally, the debruijn index is the index of the binder where a name was defined, starting from the innermost binder and working out."

GLOSSARY.md
```diff
@@ -0,0 +1,167 @@
...
+if and only if the subformula `P(x)` is true for all possible values for x.
+
+### Helpful equivalences
+- `not(forall(x). P(x)) <=> exists(x). not(P(x))`
+- `not(exists(x). P(x)) <=> forall(x). not(P(x))`
+
+## Skolemization
+Skolemization is a technique of transferring a logical formula with existential
+quantifiers to a statement without them. The resulting statement is in general
+not equivalent to the original statement but equisatisfiable.
```
nikomatsakis: I think we could make this a bit more intuitive. I would given an example like this:

To get an intuition for skolemization, consider this Rust function:

```rust
fn foo<T: Hash>() { 
    ...
}
```

When we type-check the body of `foo()`, we don't know yet what the type `T` will be. So we introduce a "skolemized" type -- meaning, basically, a fresh type that is distinct from all other types in the program. The only thing we know about this skolemized type `T` is that it implements `Hash`. Therefore, if we can successfully type-check the function, we know that it should type-check with any other type, so long as that type implements `Hash` (for this to work, we have to ensure that knowing *more* about a type never interferes with type-checking).

fabric-and-ink: Thank you for your thorough explanations! I will work them in during the next days :)

nikomatsakis: Great =)

nikomatsakis: A glossary is a great idea btw

nikomatsakis:
@fabric-and-ink do you mean rebasing? I'm reluctant to merge with merge commits in the PR itself.

fabric-and-ink:
Sorry! Made a mistake... I want to squash everything when this PR is ready.

fabric-and-ink:
(Still working on it :) )

nikomatsakis:
@fabric-and-ink is there any reason not to merge this? seems like good stuff so far! :)

fabric-and-ink:
Glad to hear! I think for a first iteration it is almost finished. There are still some descriptions missing or without an example. I will add them in a few days.

fabric-and-ink:
Ok, I think this should be good for now.

nikomatsakis:
🚀 thanks @fabric-and-ink !

PR #54, opened 2017-07-11
Auto traits

scalexm:
Implement auto traits and negative impls.

* For an auto trait `AutoTrait`, we generate a default impl for each type `MyType` where `MyType: (!)AutoTrait` does not unify with an existing positive or negative impl. A default impl will basically retrieve the field types `FieldType1<...>, ..., FieldTypeN<...>` within `MyType` and generate the impl:
```rust
impl<...> AutoTrait for MyType where FieldType1<...>: AutoTrait, ..., FieldTypeN<...>: AutoTrait
```

* Negative impls do not generate any rule during lowering and are just here to deactivate default implementations (although they can be implemented for any trait, not just auto traits). Negative impls can overlap arbitrarily. Positive and negative impls for a same trait cannot overlap. Negative impls cannot define associated type values.

* Goals of the form `T: AutoTrait` have coinductive semantics. If we encounter a cycle whose all components have coinductive semantics, we accept the cycle as an infinite proof and return `Ok` without performing any substitution.

* The syntax for auto traits is `#[auto] trait Send { }`


chalk-parse/src/parser.lalrpop
```diff
@@ -47,12 +47,17 @@ StructDefn: StructDefn = {
     }
 };
 
+AutoKeyword: () = "#" "[" "auto" "]";
```
nikomatsakis: if you wanted to make `auto` usable as an identifier, you could do this:

```
Id: Identifier = {
    <l:@L> <s:RawIdent> <r:@R> => Identifier {
        str: intern(s),
        span: Span::new(l, r),
    }
};

RawIdent: &'input str = {
    "auto",
    r"([A-Za-z]|_)([A-Za-z0-9]|_)*",
};
```

scalexm: Yeah I see, but actually I think `#[auto]` is fine now :)

src/solve/solver.rs
```diff
@@ -129,10 +129,26 @@ impl Solver {
...
-        // and return `Err`. If there is one, use this answer.
-        if let Some(slot) = self.stack.iter_mut().find(|s| { s.goal == goal }) {
+        // The goal was already on the stack: we found a cycle.
+        if let Some(index) = self.stack.iter().position(|s| { s.goal == goal }) {
+
+            // If we are facing a goal of the form `?0: AutoTrait`, we apply coinductive semantics:
+            // if all the components of the cycle also have coinductive semantics, we accept
+            // the cycle `(?0: AutoTrait) :- ... :- (?0: AutoTrait)` as an infinite proof for
+            // `?0: AutoTrait` and we do not perform any substitution.
+            if self.stack.iter().skip(index).all(|s| s.goal.is_coinductive(&*self.program)) {
```
scalexm: Actually I missed `chain(Some(goal.clone()))` here, adding that...

src/solve/test.rs
```diff
@@ -1399,3 +1399,159 @@ fn inapplicable_assumption_does_not_shadow() {
...
+
+        goal {
+            i32: Send
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            forall<T> {
+                Vec<T>: Send
```
nikomatsakis: So this through me for a bit, but I guess it's true because `struct Vec<T> { }` has no fields

src/ir/mod.rs
```diff
@@ -84,8 +87,11 @@ impl Environment {
     pub fn add_clauses<I>(&self, clauses: I) -> Arc<Environment>
         where I: IntoIterator<Item = DomainGoal>
     {
+        use std::collections::HashSet;
```
scalexm: I'm perfectly ok with moving it to the top. I don't know why but some times I prefer local imports if they're used only at one little place. Not every time though haha.

src/lower/default.rs
```diff
@@ -0,0 +1,80 @@
...
+                            trait_ref,
+                            accessible_tys: struct_datum.binders.value.fields.clone(),
+                        }
+                    }
+                });
+            }
+        }
+    }
+
+    fn provide_impl_for(&self, trait_ref: TraitRef, struct_datum: &StructDatum) -> bool {
```
scalexm: Right, actually I was reading this like "self provide_impl_for struct", but I agree this can be confusing.

src/solve/test.rs
scalexm: Indeed, but actually this test is not great, I'm going to change that.

src/solve/test.rs
```diff
@@ -1423,3 +1423,182 @@ fn inapplicable_assumption_does_not_shadow() {
...
+        }
+
+        // `WellFormed(T)` is needed here because of the expanded bound `WellFormed(Ptr<List<T>>: Send)`
+        // on the default `List<T>: Send` impl, which will need that `List<T>` is well-formed in order to be
+        // proven, which will in turn need that `T` is well-formed.
+        //
+        // In fact, as soon as there is a field which is referencing `T` with an indirection like `Foo<Bar<T>>`,
+        // we need to add the `WellFormed(T)` because the `if (T: Send)` elaborates `WellFormed(T: Send)` but not
+        // `WellFormed(T)`. This is not an issue, but is maybe a bit inconsistent.
+        goal {
```
scalexm: @nikomatsakis so actually I added a comment here, which I think is interesting.

src/solve/solver.rs
```diff
@@ -136,7 +136,12 @@ impl Solver {
             // if all the components of the cycle also have coinductive semantics, we accept
             // the cycle `(?0: AutoTrait) :- ... :- (?0: AutoTrait)` as an infinite proof for
             // `?0: AutoTrait` and we do not perform any substitution.
-            if self.stack.iter().skip(index).all(|s| s.goal.is_coinductive(&*self.program)) {
+            if self.stack.iter()
+                         .skip(index)
+                         .map(|s| &s.goal)
+                         .chain(Some(&goal))
```
nikomatsakis: Heh. =) Funny oversight. Maybe add a test for this? Maybe not worth it.

scalexm: Actually I realised that from a correctness point of view, that's the same thing because we already know that there is a cycle starting at index `index`, so the very first element of our iterator is also our current goal already. Anyway, let's keep it that way :)

scalexm:
Actually the comment I've added on my last commit made me think of something which is not great. Given:
```rust
#[auto] trait Send { }
trait Clone { }

struct Clonable<T> where T: Clone { }

struct S<T> where T: Clone {
    data: Clonable<T>
}

// default impl generated:
// impl<T> Send for S<T> where Clonable<T>: Send, WF(Clonable<T>: Send)
//
// WF(Clonable<T>: Send) :- WF(Clonable<T>)
// WF(Clonable<T>) :- WF(T), T: Clone, WF(T: Clone)
// WF(T: Clone) :- WF(T)
```
we only have:
```
forall<T> {
    if (WellFormed(T), T: Send, T: Clone) {
        S<T>: Send
    }
}
```

With implied bounds, it's a bit better:
```
forall<T> {
    if (WellFormed(T), WellFormed(S<T>), T: Send) {
        S<T>: Send
    }
}
```

Basically it exposes WF checks for fields.

I don't know if this is a problem, but it feels awkward. I think we shouldn't expand the WF bounds on default implementations? This seems fine because expanding WF bounds on impls is only there to ensure that in things like:
```rust
trait SuperTrait { }
trait Trait : SuperTrait { }

impl<T> Foo for MyStruct<T> where T: Trait { }
```
we correctly check that `T` implements both `Trait` and `SuperTrait` to conclude that `MyStruct<T>: Foo`. There are no such bounds on default implementations.

As was discussed previously in PR #49, we shouldn't expose WF checks for fields (especially for proving things like `S<T>: Send`).

nikomatsakis:
@scalexm remind me, are we waiting for any further changes here?

nikomatsakis:
I guess we decided to tweak (or maybe tweak) how we approach the `WF` rules, right... I remember now =)

scalexm:
@nikomatsakis Removing the `WF(FieldType: Send)` bounds on default implementations is fine for the moment (I did not push my commit yet, I'll do that within a few hours).

I'll open a second PR later for tweaking WF rules + implied bounds (I think what we discussed on Gitter works pretty well).

PR #55, opened 2017-08-05
`OccursCheck` is a folder

### LINKED ISSUE #25: refactor occurs check into a folder
The occurs check code in `src/solve/infer/unify.rs` (specifically, the methods defined on `OccursCheck`) follows a very folder-like pattern. Unfortunately, it can't quite use the `Folder` trait (as defined in `fold.rs`) because that trait only contains "callback methods" for processing free lifetime/type/krate variables. The occurs check needs to be able intercept all types, at least, as well as names that live in a universe and a few other things. We could add callback methods for those things to `Folder`. 

I imagine the resulting trait would look like:

```rust
pub trait Folder {
    // Methods today:
    fn fold_free_var(&mut self, depth: usize, binders: usize) -> Result<Ty>;
    fn fold_free_lifetime_var(&mut self, depth: usize, binders: usize) -> Result<Lifetime>;
    fn fold_free_krate_var(&mut self, depth: usize, binders: usize) -> Result<Krate>;

    // Methods we need:
    fn fold_ty(&mut self, ty: &Ty, binders: usize) -> Result<Ty>;
    fn fold_lifetime(&mut self, lifetime: &Lifetime, binders: usize) -> Result<Lifetime>;
}
```

One thing that I'm not crazy about is that `fold_ty` invokes `fold_free_var` etc, at least in its default incarnation, so putting them into the same trait might mean that if you override `fold_ty` you would not need the other methods. So it might be better to have `Folder` just include the higher-level methods (e.g., `fold_ty`) and then have a new trait `VarFolder` where we define a bridge impl like `impl<T: VarFolder> Folder for T`. Or something like that.

fabric-and-ink:
Attempt at solving #25. I kept as close as possible to the suggestions in the issue description. Thank you for helping me @scalexm :)

Closes #25

### DIFF
--- a/src/fold/instantiate.rs
+++ b/src/fold/instantiate.rs
@@ -39,7 +39,7 @@ macro_rules! subst_method {
 subst_method!(Goal);
 subst_method!(Ty);
 
-impl<'b> Folder for Subst<'b> {
+impl<'b> FolderVar for Subst<'b> {
     fn fold_free_var(&mut self, depth: usize, binders: usize) -> Result<Ty> {
         if depth >= self.parameters.len() {
             Ok(Ty::Var(depth - self.parameters.len() + binders))
--- a/src/fold/mod.rs
+++ b/src/fold/mod.rs
@@ -52,13 +52,25 @@ impl<T: FolderVar> Folder for T {
     }
 }
 
-impl<'f, F: Folder + ?Sized> Folder for &'f mut F {
+pub struct FolderRef<'f, F> where F: 'f + ?Sized {
+    folder: &'f mut F,
+}
+
+impl<'f, F: ?Sized> FolderRef<'f, F> {
+    pub fn new(folder: &'f mut F) -> Self {
+        FolderRef {
+            folder,
+        }
+    }
+}
+
+impl<'f, F: Folder + ?Sized> Folder for FolderRef<'f, F> {
     fn fold_ty(&mut self, ty: &Ty, binders: usize) -> Result<Ty> {
-        self.fold_ty(ty, binders)
+        self.folder.fold_ty(ty, binders)
     }
 
     fn fold_lifetime(&mut self, lifetime: &Lifetime, binders: usize) -> Result<Lifetime> {
-        self.fold_lifetime(lifetime, binders)
+        self.folder.fold_lifetime(lifetime, binders)
     }
 }
 
--- a/src/fold/mod.rs
+++ b/src/fold/mod.rs
@@ -22,17 +22,6 @@ pub trait Folder {
     fn fold_lifetime(&mut self, lifetime: &Lifetime, binders: usize) -> Result<Lifetime>;
 }
 
-fn fold_parameter(f: &mut GeneralFolder, binders: usize, arg: &Parameter)
-    -> Result<Parameter> {
-    match *arg {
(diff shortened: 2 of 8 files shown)


scalexm:
Looks good to me.
(We used to have a Travis hook, where has it gone?)

fabric-and-ink:
(No idea but a local `cargo test` runs fine 🙂)

nikomatsakis:
@fabric-and-ink nice! will look soon, sorry for delay.

nikomatsakis:
OK, this does indeed look good! Nice work.

Sadly, it didn't result in as much code de-duplication as I initially anticipated (well, it doesn't result in *any* code de-duplication, really). It may still be a good change: for example, it allows us to easily apply the occurs check to other foldable structures, should we ever need to do so. 

(I apologize again for my slow response; it's been a hard month for keeping up with things.)

fabric-and-ink:
Yes, that is true. `OccursCheck` has a lot of stuff going on for itself. What exactly feels a bit more complex than you anticipated?

(Do not worry about the quickness of you responses. It is obvious due to your involvements that things may take a while.)

nikomatsakis:
@fabric-and-ink 

> What exactly feels a bit more complex than you anticipated?

Well, it's not really your PR that feels more complex, it's mostly that we didn't get to kill any code from the occurs-check. The only real change is that instead of invoking `check_ty` (or whatever) it invokes `fold_ty`, which still winds up doing a case match and invoking custom code for each case (e.g., [`check_quantified()` etc](https://github.com/nikomatsakis/chalk/pull/55/files#diff-feebeff79f72868c3b657c0b21a702bfR393)). I didn't actually look at the contents of those functions though -- maybe some of them can be factored away?

Looking now, [this is the code in question](https://github.com/fabric-and-ink/chalk/blob/247345627e883ae2a42a44a858a7f3bcc1ad02e2/src/solve/infer/unify.rs#L338-L363). It does seem like, with the exception of [this check](https://github.com/fabric-and-ink/chalk/blob/247345627e883ae2a42a44a858a7f3bcc1ad02e2/src/solve/infer/unify.rs#L340), this is mostly boilerplate that *could* be part of the `Folder` trait, in principle, though I think it would require some deeper changes or at least extracting helpers.

fabric-and-ink:
I will take a look.

fabric-and-ink:
My first suggestion would be to introduce a third trait `CheckFolder` that has functions to process `ApplicationTy`, `QuantifiedTy`, etc. with default implementations. For this trait we implement `Folder`. Therefore `CheckFolder` would be similar to `FolderVar` but more verbose. But we could omit all `check_*` functions except `check_apply` which calls `universe_check` from the implementation of `CheckFolder` for `OccursCheck`. (If I am not mistaken.)

nikomatsakis:
@fabric-and-ink 

That sounds reasonable, though I wonder if `CheckFolder` and the `Folder`  that you added could be combined?

fabric-and-ink:
Good idea. I will look into it on sunday.

fabric-and-ink:
The functions of the check seem too riddled with little necessary additions to be moved into the trait(s) in a useful manner. My attempts so far have resulted in a mess on the trait side.

fabric-and-ink:
My pleasure :)

PR #59, opened 2017-10-04
introduce an experimental solver based on the SLG algorithm

nikomatsakis:
This branch introduces a new solver based on the SLG algorithm. This is based heavily on the algorithm from the paper "Efficient Top-Down Computation of Queries Under the Well-formed Semantics" by Chen et al, but extended to support first-order hereditary harrop predicates as well as overflow (which results in an "incomplete" result that may not contain all answers). 

The solver is usable via the repl but not yet complete:

- I am not sure if the handling of overflow really makes sense the way I did it.
- It cannot handle `cannot_prove` and `region_constraints` that result from unification yet
    - pushing `region_constraints` into `ExClause` (and then into the answers) seems relatively straightforward
    - `cannot_prove` can probably be handled by the overflow mechanism
- Its result is not yet packaged up in the same way as other chalk queries.
- It is not obvious how to give **advice** from an SLG solver in the event of ambiguity, though I have some thoughts.

cc @scalexm @aturon @arielb1


fabric-and-ink:
Thank you for documenting the `slg` module so well. What is the difference or improvement over the existing unification algorithm?

arielb1:
> I am not sure if the handling of overflow really makes sense the way I did it.

Is there a reason we need to "handle" overflow? I rather like the situation where the only purpose of the overflow limit is to protect the compiler from ill-behaved programs and the compiler works just fine with an overflow limit of infinity. That basically decouples the overflow limit from most "correctness" concerns.

Is there a particular reason we care about the limit here?

scalexm:
Two things:
* coherence is still checked with the default solver, even when `--slg` is passed, I guess this should be changed?
* if we don't manually set an overflow depth in the comment line, then given the following program:
```rust
trait Foo { }
struct Box<T> { }
impl<T> Foo for T where Box<T>: Foo { }
```
and the following goal: `exists<T> { T: Foo }`, I get: `fatal runtime error: stack overflow`

nikomatsakis:
@scalexm 

> coherence is still checked with the default solver, even when --slg is passed, I guess this should be changed?

Yes, it should, I just didn't get to it.

> if we don't manually set an overflow depth in the command line, then given the following program:

Weird. I also found another bug (where sometimes you get 5 answers to a particular query, sometimes 7). I'm inclined to land this PR and address those bugs as follow-ups, what do you think?

nikomatsakis:
@arielb1 

> Is there a reason we need to "handle" overflow? I rather like the situation where the only purpose of the overflow limit is to protect the compiler from ill-behaved programs and the compiler works just fine with an overflow limit of infinity. That basically decouples the overflow limit from most "correctness" concerns. Is there a particular reason we care about the limit here?

Sorry for the delay in answering you. I got distracted by exploring some of the pre-existing solutions. Ultimately, I wound up adopting two techniques for guaranteeing termination that are not based on a concept of overflow (that is, they impose no limits on the depth of the stack *per se*). The basic idea is that there is a maximum size to the types that we will explore, and we use this to "cap" subgoals and answers. I talk about it [in this blog post that I just wrote](http://smallcultfollowing.com/babysteps/blog/2017/10/21/chalk-meets-slg/#guaranteed-termination), so I'll hold off on a detailed discussion of how it works.

But I wanted to answer this part of your question in more detail:

> the only purpose of the overflow limit is to protect the compiler from ill-behaved programs 

That's not quite the scenario in this solver. In particular, it always tries to enumerate all the answers to all queries, so if given a simple query like `?T: Sized`, it would otherwise wind up enumerating answers for ever without some cutoff. This is the thing I am most interested in tweaking, actually.

nikomatsakis:
> Weird. I also found another bug (where sometimes you get 5 answers to a particular query, sometimes 7). I'm inclined to land this PR and address those bugs as follow-ups, what do you think?

OK, I found the bug why I sometimes get 5 answers. It's kind of a bone-headed -- but serious! -- mistake on my part. I'm not handling the inference table variables correctly at some points and wind up eliminating possibilities. I have to think about the best way to fix it. =)

nikomatsakis:
@scalexm 

> if we don't manually set an overflow depth in the command line

OK, so, I think this is not unexpected. What's happening is that I though the default overflow depth would be 10, but I must messed up the docopt crate syntax, so you wind up with a default of 0, which effectively serves as no default depth. This does indeed mean we will just run infinitely until we run out of memory (which I believe shows up as a stack overflow). Similarly, if you set the limit to something like 500, you will see the same result.

As a related aside, I wrote this code in a naive, "stack-oriented" fashion with no tail recursion. I just threw in stacker to prevent it from segfaulting under "reasonable" usage. I would like to make it less naive, maybe even -- ideally -- using async-await or futures.

nikomatsakis:
Just thought of a random thing -- the subgoal abstraction mechanism currently works by "capping" the size of any one type, but that may not be enough. For example, a loop through a `implies` could create bigger and bigger environments. We probably want to cap the total size of the environment as well.

PR #60, opened 2017-10-10
remove `cannot_prove` from unify, change approach to universals

nikomatsakis:
Currently unification does not reject `i32 = T` (where `T` is a universally quantified variable). This leads to a fair amount of complexity, since we must now distinguish another kind of solution ("cannot prove") which is "similar but different" to ambiguity.

This commit removes all of that logic in favor of introducing a refined concept of the "must be ground" conditions from the negation-as-failure paper. In particular, the original paper noted that free inference variables wind up acting as universal constraints when negated -- which is exactly what we want here. So we introduce a new function `negated` that transforms a goal G by (a) forbidding free existential variables and (b) converting free universal variables into existential ones. If the resulting query G' is satisfiable, then `not { G }` is false.

cc @scalexm

### DIFF
--- a/src/fold/mod.rs
+++ b/src/fold/mod.rs
@@ -1,3 +1,6 @@
+//! Traits for transforming bits of IR.
+
+use cast::Cast;
 use errors::*;
 use ir::*;
 use std::fmt::Debug;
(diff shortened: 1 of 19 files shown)


nikomatsakis:
cc @fabric-and-ink -- btw, I wound up playing around with the fold trait some more. The "tweak the folder traits" commit is, I think, what I was aiming at with turning occurs into a folder.

nikomatsakis:
Gonna merge this. We can always tweak more in the future.

fabric-and-ink:
@nikomatsakis Neat! I guess this was still a bit too involved for me to get "right" :)

Thank you for the doc-comments!

src/fold/mod.rs
```diff
@@ -136,7 +157,32 @@ impl<T: Fold> Fold for Option<T> {
...
     type Result = Self;
     fn fold_with(&self, folder: &mut Folder, binders: usize) -> Result<Self::Result> {
-        folder.fold_ty(self, binders)
+        match *self {
+            Ty::Var(depth) => if depth >= binders {
+                folder.fold_free_existential_ty(depth - binders, binders)
+            } else {
+                Ok(Ty::Var(depth))
+            },
+            Ty::Apply(ref apply) => {
```
fabric-and-ink: I think it was this match-arm that made me stumble. Your solution looks certainly better than everything I came up with :)

PR #61, opened 2017-10-12
Add tests for partial overlap

mikeyhew:
This hopefully solves #8

### DIFF
--- a/src/solve/test.rs
+++ b/src/solve/test.rs
@@ -1680,3 +1680,36 @@ fn partial_overlap_2() {
         }
     }
 }
+
+#[test]
+#[ignore]
+fn partial_overlap_3() {
+    test! {
+        program {
+            trait Marker {}
+            trait Foo {}
+            trait Bar {}
+
+            impl<T> Marker for T where T: Foo {}
+            impl<T> Marker for T where T: Bar {}
+
+            struct i32 {}
+            impl Foo for i32 {}
+            impl Bar for i32 {}
+        }
+
+        goal {
+            forall<T> {
+                if (T: Foo, T: Bar) { T: Marker }
+            }
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            i32: Marker
+        } yields {
+            "Unique"
+        }
+    }
+}
--- a/src/solve/test.rs
+++ b/src/solve/test.rs
@@ -1681,8 +1681,9 @@ fn partial_overlap_2() {
     }
 }
 
+// FIXME: This test currently doesn't pass because overlapping
+// impls for marker traits are currently not supported
 #[test]
-#[ignore]
 fn partial_overlap_3() {
     test! {
         program {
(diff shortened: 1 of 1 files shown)


nikomatsakis:
These tests look good, though as I said I think there is something of a bug here in terms of the first test. Right now the only reason that is passing is that `Foo` and `Bar` have no implementations, and we don't have the [open world reasoning stuff](http://aturon.github.io/blog/2017/04/24/negative-chalk/) implemented yet (that is, we don't have the parts of chalk that let it understand that more impls may come later).

Maybe we can add in the test case you had which was failing (i.e., where there was overlap) and we'll include a "FIXME" related to that?

mikeyhew:
@nikomatsakis OK I added in the failing test with an `#[ignore]` attribute. Were you saying that the `forall<T>` goal in the first test shouldn't actually be passing? If so, I can switch the expected result to "Ambiguous" and ignore it, and do the same for the `forall<T>` goal in the third test I just added.

src/solve/test.rs
```diff
@@ -1680,3 +1680,36 @@ fn partial_overlap_2() {
         }
     }
 }
+
+#[test]
+#[ignore]
```
nikomatsakis: I'd prefer not to have `#[ignore]` but rather include the current result that we see, with a comment that it is a bug. (I don't like to ignore tests, personally, it's hard to remember when to un-ignore them.)

nikomatsakis:
> Were you saying that the `forall<T>` goal in the first test shouldn't actually be passing?

No,  I wasn't saying that. I think it should pass.

mikeyhew:
@nikomatsakis updated. If you have any other requests, don't be shy, and if you can think of a better name for any of these tests, please let me know

nikomatsakis:
@mikeyhew Thanks! I see that the test fails -- I think this is because of a bug that I introduced in https://github.com/nikomatsakis/chalk/pull/60. I have since fixed that bug in https://github.com/nikomatsakis/chalk/pull/59, which I do hope to merge soon. Maybe we should wait for that?

Also, are you interested in taking a crack at some other chalk tasks? I've been thinking about a slew of new bugs to open. =)

mikeyhew:
@nikomatsakis AFAIK the test fails because marker traits aren't implemented yet, as you mentioned in #8. I'd be interested in implementing them. It doesn't seem like it would be too, too complicated, as long as someone gives me some pointers :)

nikomatsakis:
> AFAIK the test fails because marker traits aren't implemented yet

Ah, I see, right -- I forgot that we invoke `unwrap()` on the coherence results, so you can't actually write the sort of `FIXME` that I wanted. 

> It doesn't seem like it would be too, too complicated, as long as someone gives me some pointers :)

Indeed! Let me write up an issue on it.

nikomatsakis:
OK, I merged but added a `#[should_panic]`.

nikomatsakis:
Thanks @mikeyhew !

PR #64, opened 2017-10-22
Allow overlapping impls for marker traits in the coherence rules

### LINKED ISSUE #62: support overlap for marker traits
The current coherence rules do not have any special cases for "marker traits" -- that is, traits with zero items. As [discussed](https://github.com/rust-lang-nursery/chalk/issues/8#issuecomment-336644098) in #8, at least for now it'd be nice to have "marker-ness" be declared explicitly with `#[marker]`. 

Here are some steps to do this:

We want to model this on the support for `#[auto]` trait declarations. This means we would:

- add a field to the ast for trait definitions ([like `auto`](https://github.com/rust-lang-nursery/chalk/blob/c8971c18294a7310010806a7283cdd417bec5c50/chalk-parse/src/ast.rs#L39))
- add the keyword to the parser ([like this](https://github.com/rust-lang-nursery/chalk/blob/c8971c18294a7310010806a7283cdd417bec5c50/chalk-parse/src/parser.lalrpop#L51-L63))
- add a field to [the `TraitDatumBound` struct](https://github.com/rust-lang-nursery/chalk/blob/c8971c18294a7310010806a7283cdd417bec5c50/src/ir/mod.rs#L243-L248), again, like `auto`
- In [the `visit_specializations` function](https://github.com/rust-lang-nursery/chalk/blob/c8971c18294a7310010806a7283cdd417bec5c50/src/coherence/solve.rs#L9), we can basically just ignore marker traits, I believe

It might be nice, rather than just adding another boolean field, to introduce some sort of `TraitFlags` type (e.g., `struct TraitFlags { auto: bool, marker: bool }`) into the AST and copy that over into `ir::TraitDatumBound`? Or at least I don't love having a bunch of random boolean variables floating around.

lqd:
This introduces the explicit `#[marker]` attribute for marker traits, to allow overlapping impls for marker traits in coherence.

Fixes #62.

Marked WIP because of these questions/comments:
- in parsing, should we ensure `#[auto]` and `#[marker]` can or can't be used at the same time ? or that marker traits are "correct" eg have zero items ?
- I added the `#[marker]` attribute to the solver tests with traits named `Marker` but should we mark others as most of the tests do use marker traits ? or maybe only to the tests *requiring* it, ie `partial_overlap_3`
- this also introduces the requested `TraitFlags` struct but only on the ast side, with bools in  `ir::TraitDatumBound` — is that what was expected or did the instruction mean to also add a similar struct to the ir side ?

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -36,7 +36,12 @@ pub struct TraitDefn {
     pub parameter_kinds: Vec<ParameterKind>,
     pub where_clauses: Vec<WhereClause>,
     pub assoc_ty_defns: Vec<AssocTyDefn>,
+    pub flags: TraitFlags,
+}
+
+pub struct TraitFlags {
     pub auto: bool,
+    pub marker: bool,
 }
 
 pub struct AssocTyDefn {
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -49,16 +49,20 @@ StructDefn: StructDefn = {
 };
 
 AutoKeyword: () = "#" "[" "auto" "]";
+MarkerKeyword: () = "#" "[" "marker" "]";
 
 TraitDefn: TraitDefn = {
-    <auto:AutoKeyword?> "trait" <n:Id><p:Angle<ParameterKind>> <w:WhereClauses> "{"
+    <auto:AutoKeyword?> <marker:MarkerKeyword?> "trait" <n:Id><p:Angle<ParameterKind>> <w:WhereClauses> "{"
         <a:AssocTyDefn*> "}" =>
     TraitDefn {
         name: n,
         parameter_kinds: p,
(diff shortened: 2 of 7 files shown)


nikomatsakis:
> in parsing, should we ensure `#[auto]` and `#[marker]` can or can't be used at the same time ? or that marker traits are "correct" eg have zero items ?

It should be possible to use both at the same time.

> should we mark others as most of the tests do use marker traits

It probably doesn't matter too much, but we should not. The whole point of adding a distinct `#[marker]` annotation was that we wanted to make an opt-in choice as to which traits are considered markers. We may still want to align this with Rust's current syntax as well -- which is just to have zero items.

> did the instruction mean to also add a similar struct to the ir side

I meant to use a struct on both sides. The `ir` could maybe just use the `ast` struct, or it could have its own copy.

nikomatsakis:
@lqd nice, merged =)

PR #66, opened 2017-10-31
Recursive tables

nikomatsakis:
Branch that does two things:

- Make Chalk code agnostic over what solver it uses. It now threads around a `SolverChoice` enum that carries all the parameters and can be used to solve root goals.
    - I'd like to modify this interface so that `SolverChoice` instantiates a `dyn Solver` which carries around an internal cache, but I didn't get around to that.
- Improve the recursive solver to be less inefficient. It now has a notion of tables as well as a cache for use across root goals (and within root goals). It can still wind up doing a lot of work because it must reach an internal fixed point for each of the subqueries.
    - I think this current version is kind of the simplest "naive" solution, though I'd love to be proven wrong.
    - To do better, one must track the interdependencies between tables more intelligently.

One thing that's missing: good benchmarks! I made some `#[bench]` infrastructure with the simple test I've been using so far, but more work is needed. (Also, it seems that SLG gets the wrong result on this!)


nikomatsakis:
Gonna land this.

PR #67, opened 2017-11-01
Edit Glossary

fabric-and-ink:
I made some edits to the GLOSSARY.md.

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -36,9 +36,9 @@ canonical form but `?0: A && ?1: B` is.
 (See [*DeBruijn Index*](#debrujin-index).)
 
 ## Clause
-In the A clause is the disjunction of several expressions. For example the clause
-`condition_1 || condition_2 || ...` states that at least one of the conditions
-holds.
+In the A clause is the disjunction of several expressions. For example the
+clause `condition_1 || condition_2 || ...` states that at least one of the
+conditions holds.
 
 There are two notable special cases of clauses. A *Horn clause* has at most one
 positive literal. A *Definite clause* has exactly one positive literal.
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -30,11 +30,10 @@ Examples for binders:
 - A sum `\sum_n x_n` binds the index variable `n`.
 
 ## Canonical Form
-A formula in canonical form has the property that its DeBruijn indices are
-minimized. For example when the formula `forall<0, 1> { 0: A && 1: B }` is
-processed, both "branches" `0: A` and `1: B` are processed individually. Then
-the first branch would be in canonical form, the second branch not since the
-occurring DeBruijn index `1` could be replaced with `0`.
+For a formula in canonical form all its variables are assigned integer IDs with
+increasing values starting at zero. For example `?3: A && ?4: B` is not in
+canonical form but `?0: A && ?1: B` is.
+(See [*DeBruijn Index*](#debrujin-index).)
 
 ## Clause
 In the A clause is the disjunction of several expressions. For example the clause
@@ -79,6 +78,8 @@ innermost binder increasing from the inside out.
 Given the example `forall<T> { exists<U> { T: Foo<Item=U> } }` the
 literal names `U` and `T` are replaced with `0` and `1` respectively: `forall<0>
 { exists<1> { 0: Foo<Item=1> } }`.
+
+See also [*Canonical form*](#canonical-form).
 
 ## Formula
 A formula is a logical expression consisting of literals and constants connected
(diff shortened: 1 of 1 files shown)


GLOSSARY.md
```diff
@@ -32,8 +32,8 @@ Examples for binders:
 ## Canonical Form
 A formula in canonical form has the property that its DeBruijn indices are
 minimized. For example when the formula `forall<0, 1> { 0: A && 1: B }` is
```
nikomatsakis: Pre-existing, but I'm not sure this is how I would phrase it. It's not so much that the debruijn indices are *minimized*, but rather that free variables in the formula are assigned numbers starting from 0, based on the order in which they appear.

Also, this example is just not correct, or at least confusing to me. That is, this formula *is* in canonical form -- and, indeed, in closed form. But a formula like this `?3: A && ?4: B` is in neither -- the canonical form of it would be `?0: A && ?1: B`. 

I sometimes think of canonical form as converting into closed form -- that is, adding some binders at the outermost level, sort of like going from `?X: A && ?Y: B` to `exists<X, Y> { X: A && Y: B }`, but really that's not right. Canonicalization for example preserves the "universe" of each variable, and the solver acts a bit differently in that the variables that are enumerated in the canonicalization become part of the final solution that is returned.

GLOSSARY.md
```diff
@@ -47,7 +47,8 @@ positive literal. A *Definite clause* has exactly one positive literal.
 *Horn clauses* can be written in the form `A || !B || !C || ...` with `A` being
 the optional positive literal. Due to the equivalence `(P => Q) <=> (!P || Q)`
 the clause can be expressed as `B && C && ... => A` which means that A is true
-if `B`, `C`, etc. are all true. All rules in chalk are in this form. For example
+if `B`, `C`, etc. are all true. All rules in *chalk* are in this form. For
```
nikomatsakis: This is not, strictly speaking, true. The rules in Chalk are actually "hereditary harrop" clauses, which is an extension of Horn clauses. For example, Horn clauses do not support `forall<T> { ... }` goals or `A => B` goals.

GLOSSARY.md
```diff
@@ -181,12 +188,16 @@ if and only if the subformula `P(x)` is true for all possible values for x.
...
 quantifiers to a statement without them. The resulting statement is in general
 not equivalent to the original statement but equisatisfiable.
 
+The idea to accomplish this transformation by adding the type variable inside an
+existential quantifier to the current universe. This new type variable has to be
+unified with another type later on, otherwise the goal cannot be satisfied.
+
 ## Validity
-An argument (*premisses* therefore *conclusion*) is valid iff there is no
+An argument ("*premisses* therefore *conclusion*") is valid iff there is no
```
nikomatsakis: Spelling: "premises"

fabric-and-ink:
(Will continue to work on this next week. Really busy right now...)

GLOSSARY.md
fabric-and-ink: I added a short paragraph about hereditary Harrop clauses. But I wonder about the second type of goal that you mention. It is not yet clear to me why it is not supported by Horn clauses.

GLOSSARY.md
fabric-and-ink: I see!

GLOSSARY.md
nikomatsakis: >  It is not yet clear to me why it is not supported by Horn clauses.

A `A => B` goal means "assume `A` is true, then prove `B`". Horn clauses simply have no such construct. You have to create *global* rules saying what is and is not provable, and then apply them everywhere (vs having a local hypothesis that `A` is true which is only usable while proving `B`).

nikomatsakis:
I've been into making mdbooks lately. Maybe we should make this one of those. =)

fabric-and-ink:
Good idea! I will look into it.

GLOSSARY.md
fabric-and-ink: (Having some trouble with the GitHub UI... accidentally deleted my comment from 15 minutes ago.) I think, I have still a bit trouble understanding. I'm quite sure that  `(P => Q) <=> (!P || Q)`. Therefore the "effect" with respect to the unification for both kinds of clauses has to be  different, right? Do you mean that for *Horn* clauses you have to create *global* rules, whereas for *HH* clauses you can formulate *local* rules? What does that mean exactly? Could you please elaborate a bit more? Thanks :)

nikomatsakis:
I now wonder if we should merge this into the rustc-guide, which has some extensive coverage of traits:

https://rust-lang-nursery.github.io/rustc-guide/traits.html

fabric-and-ink:
Ok!

PR #68, opened 2017-11-01
Add a CONTRIBUTING.md file

fabric-and-ink:
I added a `CONTRIBUTING.md` file which is more or less a copy of the Rust projects' file. Only I ripped out most of the things that seem to be unnecessary (currently).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,147 @@
+*Note that this `Contribution.md` document is more or less a copy of the file
+with the same name of the [Rust compiler](https://github.com/rust-lang/rust)
+project.*
+
+# Contributing to chalk
+
+Thank you for your interest in contributing to chalk! There are many ways to
+contribute, and we appreciate all of them.
+
+* [Feature Requests](#feature-requests)
+* [Bug Reports](#bug-reports)
+* [Pull Requests](#pull-requests)
+* [Writing Documentation](#writing-documentation)
+* [Issue Triage](#issue-triage)
+* [Helpful Links and Information](#helpful-links-and-information)
+
+If you have questions, please join [our gitter channel](https://gitter.im/chalk-rs/Lobby).
+
+As a reminder, all contributors are expected to follow our [Code of Conduct][coc].
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+Sometimes, an issue will stay open, even though the bug has been fixed. And
+sometimes, the original bug may go stale because something has changed in the
+meantime.
+
+It can be helpful to go through older bug reports and make sure that they are
+still valid. Load up an older issue, double check that it's still true, and
+leave a comment letting us know if it is or is not. The [least recently
+updated sort][lru] is good for finding issues like this.
+
+Contributors with sufficient permissions on the Rust repo can help by adding
```
nikomatsakis: I don't think we're really using these same issue labels...

README.md
```diff
@@ -16,24 +16,6 @@ suitable reference.
...
-- [Cyclic queries in chalk](http://smallcultfollowing.com/babysteps/blog/2017/09/12/tabling-handling-cyclic-queries-in-chalk/)
-    - Handling cyclic relations and enabling the implementation of implied bounds and other long-desired features in an elegant way
-
-## Documentation
-
-In addition to the blog posts there is a [glossary](GLOSSARY.md) explaining some
-of the terminology used in *chalk*.
+## Contributing
+If you are interested in contributing to chalk, please take a look at the
+[Contribution guide](CONTRIBUTING.md).
```
nikomatsakis: Hmm, I think it's kind of sad to bury this information in the CONTRIBUTING.md file, which I suspect many will not read, and likely not read in full. I'm not sure if it needs to live in the README, but at least we should link directly to the section in the contributing file where stuff can be found, and keep the link to the glossary etc.

fabric-and-ink: Agreed. I had the same thought later on.

CONTRIBUTING.md
fabric-and-ink: I know, but we could! 🙂  Not sure if that really helps at all though... Three possibilities:
1) Align current tags with the tags from rustc
2) Change the tags in this file
3) Delete this section altogether

Which do you prefer?

nikomatsakis:
I suspect we can simplify this greatly, but it's a good start! Thanks @fabric-and-ink

Issue #58, opened 2017-09-27
model the algorithm for normalizing when trait is unknown


nikomatsakis:
Right now in Chalk we don't have support for types like `T::Item`; we always require fully explicit types like `<T as Iterator>::Item`. In particular, the trait is always known. So for example  [in the AST node, we know the trait](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/chalk-parse/src/ast.rs#L132-L136). In the lowered IR, a projection [references the id of a particular trait item](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/src/ir/mod.rs#L350-L354), and hence implicitly specifies a particular trait.

The rules in Rust are a bit hackily implemented but the *idea* is that we should permit `T::Item` only if there is a unique trait that could work, selected from the set of in-scope traits. I think the first step for this issue therefore is to model in-scope traits somehow. I'd probably want to do this with a new kind of [`DomainGoal`](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/src/ir/mod.rs#L388), something like `InScope(ItemId)`, where `ItemId` is the id of a trait. We would have to give this some syntax too (e.g., `InScope(Iterator)`). Then we can model which traits are in scope by adding those facts into the environment (e.g., `if (InScope(Iterator)) { .. })`).

Here are the steps to add `InScope`:

- Extend [the AST `WhereClause` struct](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/chalk-parse/src/ast.rs#L164-L171) to add something for `InScope(...)`.
- Extend [the parser](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/chalk-parse/src/parser.lalrpop#L144) to parse it.
- Extend [`DomainGoal` to include the new case `InScope(ItemId)`](https://github.com/nikomatsakis/chalk/blob/72e655355e237ff1b9fdaeb024e8e46569178e3c/src/ir/mod.rs#L388), as described before.
[...]

PR #69, opened 2017-11-01
Normalizing items when trait is unknown ("unselected" projections)

Garogolun:
Added `(Trait::Item)` to the grammar (note the parentheses) instead of `Trait::Item` to prevent a left recursive grammar. (I am too lazy.)

Is it okay that the `SLG` solver returns `No possible solution: no answers`? Or does it indicate that I need to adjust that solver?

See also #58.

### DIFF
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -106,13 +106,19 @@ AssocTyValue: AssocTyValue = {
 };
 
 pub Ty: Ty = {
-    <n:Id> => Ty::Id { name: n},
-    <n:Id> "<" <a:Comma<Parameter>> ">" => Ty::Apply { name: n, args: a },
-    <p:ProjectionTy> => Ty::Projection { proj: p },
     "for" "<" <l:Comma<LifetimeId>> ">" <t:Ty> => Ty::ForAll {
         lifetime_names: l,
         ty: Box::new(t)
     },
+    TyWithoutFor,
+};
+
+TyWithoutFor: Ty = {
+    <n:Id> => Ty::Id { name: n},
+    <n:Id> "<" <a:Comma<Parameter>> ">" => Ty::Apply { name: n, args: a },
+    <p:ProjectionTy> => Ty::Projection { proj: p },
+    <proj:UnselectedProjectionTy> => Ty::UnselectedProjection { <> },
+    "(" <Ty> ")",
 };
 
 Lifetime: Lifetime = {
@@ -127,6 +133,17 @@ Parameter: Parameter = {
 ProjectionTy: ProjectionTy = {
     "<" <t:TraitRef<"as">> ">" "::" <n:Id> <a:Angle<Parameter>> => ProjectionTy {
         trait_ref: t, name: n, args: a
+    },
+};
+
+UnselectedProjectionTy: UnselectedProjectionTy = {
+    <ty:TyWithoutFor> "::" <name:Id> <a:Angle<Parameter>> => {
+        let mut args = vec![Parameter::Ty(ty)];
+        args.extend(a);
+        UnselectedProjectionTy {
+            name: name,
+            args: args,
+        }
     },
 };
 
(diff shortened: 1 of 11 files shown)


nikomatsakis:
@Garogolun let me take a look =)

Oh, I see some merge conflicts. That's probably from the PR I just landed. I hope rebasing isn't too tricky.

nikomatsakis:
Re: the SLG solver, there are still some bugs left, we should open some FIXME issues (I found one too).

assoc-ty.chalk
```diff
@@ -0,0 +1,29 @@
...
+//     type Item;
+//     type IntoIter: Iterator;
+// }
+// 
+// struct IntoIter<T> { }
+// 
+// impl<T> IntoIterator for Vec<T> {
+//     type Item = T;
+//     type IntoIter = IntoIter<T>;
+// }
```
nikomatsakis: we don't usually add random `.chalk` files into the repo (maybe I should make an `examples` directory or something though)

chalk-parse/src/parser.lalrpop
```diff
@@ -130,6 +131,17 @@ ProjectionTy: ProjectionTy = {
     },
 };
 
+UnselectedProjectionTy: UnselectedProjectionTy = {
+    "(" <ty:Ty> "::" <name:Id> <a:Angle<Parameter>> ")" => {
```
nikomatsakis: Left recursion itself should not be a problem -- do you mean that it reports an ambiguity or other error to you?

src/ir/mod.rs
```diff
@@ -526,6 +555,12 @@ pub struct Normalize {
     pub ty: Ty,
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
nikomatsakis: This feels like a nice place for a comment that explains the difference between *normalizing* and *unselected normalize* and perhaps gives a brief example (of Rust code).

src/solve/test.rs
```diff
@@ -1872,8 +1872,10 @@ fn inscope() {
 
         goal {
             InScope(Foo)
-        } yields {
+        } yields[SolverChoice::Default] {
             "No possible solution: no applicable candidates"
+        } yields[SolverChoice::SLG(MAX_SIZE)] {
+            "No possible solution: no answers"
```
nikomatsakis: If you just make this `"No possible solution"` it will work for both solvers (we only check for a prefix).

src/solve/test.rs
```diff
@@ -1885,3 +1887,66 @@ fn inscope() {
...
+
+            impl Iterator2 for Chars {
+                type Item = char2;
+            }
+        }
+
+        goal {
+            (Chars::Item) = char
+        } yields[SolverChoice::Default] {
+            "No possible solution: no applicable candidates"
```
nikomatsakis: Same here

assoc-ty.chalk
Garogolun: Whoops, forgot to remove that from the commit...

src/solve/test.rs
Garogolun: I didn't see this distinction being made in any of the other tests, so that's why I thought something wasn't entirely right here. 😛

chalk-parse/src/parser.lalrpop
Garogolun: Without the parentheses LALRPOP crashes with:
```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/lalrpop-0.13.1/src/lr1/error/mod.rs:545:33
```
Somehow I thought left recursion and LR parsers didn't like each other... (Apparently, this isn't true at all.) I don't know what the issue is. I have noticed before that LALRPOP crashes when trying to report certain conflicts. (I should have opened a bug report.)

Garogolun: Found it! `Ty` already has a right recursive rule (`"for" "<" Comma<LifetimeId> ">" Ty`). I'll give the `for` rule a lower precedence.

Garogolun:
Rebased.

I just found the "no answers" instead of "no applicable candidates" suspicious, especially because no other test had this difference, but that was before I knew the tester only looks at the prefix.

nikomatsakis:
Nice!

PR #76, opened 2018-01-29
introduce new "on demand" SLG solver

nikomatsakis:
This work is incomplete, but it seems very promising. It introduces a new variant on the SLG solver that operates 'on demand'. This variant also does a kind of 'breadth-first' search (loosely inspired by [minikanren](http://minikanren.org/)). The current version does not yet handle negative cycles or delaying, but *should* otherwise be more-or-less complete. You can see in the included test that, when asked to solve `?T: Sized`, it yields `i32, Slice<i32>, Vec<i32>, Slice<Slice<i32>>, Vec<Slice<i32>>` etc -- i.e., it is solving for many types, rather than yielding up (e.g.) `Slice<Slice<i32>>` before even examining `Vec`.

The model is that each table has a vector of stored answers. It also has a queue of active "strands". Each strand represents an active "search thread" -- basically some promising way to find the next answer. The main operation is asking a table to ensure that it has N answers -- e.g., you start out by saying "ensure_answer(table, 0)" and it will then produce its first answer. You can then say "ensure_answer(table, 1)", and it will produce the next answer, and so forth. 

Each of these `ensure_answer` calls can yield one of the following results:

- Ok -- answer is available.
- Err(QuantumExceeded) -- didn't find the answer, but still might. Try again. (Internally, we switch to another strand here.)
- Err(NoMoreSolutions) -- there are not that many answers.
    - Note that thanks to subgoal abstraction and so forth, we are guaranteed to reach this state eventually (in contrast to Minikanren).
- Err(Cycle) -- this ensure_answer call depends on something higher up in the stack; cycles can result internally, but should never occur for the top-most call.

I believe rustc could use this solver essentially directly (but time will tell). The vision for rustc integration is as follows:

- When we need to solve something, we find the table. 
- We can then pull off as many answers **as we need**. In other words, one flaw with the old scheme was that processing something like `?T: Sized` would produce an infinite flow of answers. But with the on-demand scheme, **and in particular the breadth-first walk**, we can stop after we see the first few -- we see `i32`, `Slice<i32>`, and `Vec<i32>`, and we know that there is no advice we can give.

One thing not yet implemented is the subgoal selection order. Currently we are quite dumb. So if we have a goal like `?T: Sized, i32: Eq<?T>`, we might `?T: Sized` first. This would be a poor choice, because we will be enumerating all sized things then filtering them through `i32: Eq<?T>`. It'd be better to go the other way. I'd like to add the heuristic that we pick "more constrained" subgoals first (e.g., more inputs that are not pure type variables). That would solve cases like this, though ultimately it's just a heuristic. (We might supplement with some other heuristics, for example sorting traits based on what kinds of "open-ended" impls they have, and so forth.)

(A key property of the system though is that the order of answers produced for a given trait is independent of who is asking for them; this is needed for deterministic results. It precludes *some* optimizations I can imagine.)


nikomatsakis:
cc @scalexm @arielb1

nikomatsakis:
> (A key property of the system though is that the order of answers produced for a given trait is independent of who is asking for them; this is needed for deterministic results. It precludes some optimizations I can imagine.)

Actually, this is not *obviously* true in the face of cycles (though not obviously false either, or at least not to me at this moment 😄 ). It is even less obvious when delaying comes into the picture; though I also have to go remember just what we wanted negative literals for anyway. I think they had a fairly limited role...? (I suppose we've talked on and off about enabling more general negative reasoning...)

Anyway, this "deterministic" property is not *that* key, just handy. And it's not really about being *deterministic* so much as "order independent" or "stable" (that is, for the same queries in same order, the result is deterministic either way, which is the property that I imagine rustc wants the most).

src/solve/slg/on_demand/logic.rs
```diff
@@ -222,16 +232,30 @@ impl Forest {
...
+                        subgoal_index,
+                        subgoal_table,
+                        universe_map,
+                        answer_index: AnswerIndex::ZERO,
+                    });
+                }
+
+                None => {
+                    // If we failed to create a table for the subgoal,
+                    // then the execution has "flounded" (cannot yield
```
varkor: Typo: "floundered".

nikomatsakis: Fixed

nikomatsakis:
OK, some progress. I added a wrapper for `Forest` to return an iterator over simplified solutions. This is then used to create a `Option<Solution>` for a query, drawing only as many answers as we need. This in turn lets us integrate the on-demand solver into the testing framework (**it passes all the standard Chalk tests**). This means this PR is basically ready to land, I think, though I might like to write a few more docs and tests first.

There is definitely room for improvement. For example, currently, we only stop drawing solutions when the "guidance" we can give becomes "trivial" -- so e.g. `?T: Sized` will not be exhaustively solved, because pretty quickly we'll come to the conclusion that we can't give useful advice about what `?T` ought to be.

However, in some cases, there might be an infinite family of solutions, all of the form `Vec<?T>: Foo` (e.g., imagine `impl<T> Foo for Vec<T>`). In this case, given the query `?U: Foo`, we would *still* pull all solutions (until we hit the max-size limits), even though they will quickly stop being useful for the purposes of reaching a steady state. **The on-demand solver could actually figure this out fairly readily, I think** -- but we'd have to go beyond the simple "give me next answer plz" interface, and instead have some way to test "are any of the upcoming answers going to affect this guidance". This would be a great area for refinement but seems fairly straightforward.

The other major shortcoming of the on-demand solver as-is is that it does not yet handle delayed subgoals. Since we don't have negative reasoning in Rust as a first-class thing yet, this doesn't affect any of our tests, but we ought to be able to do better here. I'm sort of inclined to handle that as a follow-up too.

nikomatsakis:
For those trying to figure out how the !@$!$! the solver works, I wrote up a README. It's not complete, but it works through what happens from an initial request and maybe gives enough clues to figure out the rest. =)

nikomatsakis:
Argh. I pushed this to master accidentally. Reverted that push, but will open a new PR.

PR #78, opened 2018-02-05
add answer abstraction to on-demand solver; remove eager SLG solver

nikomatsakis:
- First, I moved over all the tests from the eager SLG solver to newer, on-demand one.
    - This revealed that answer abstraction was missing, so I implemented that.
- Then I killed the SLG solver.
- Then I tightened up privacy from `pub` to `crate`, revealing a bunch of dead code, which I removed.


nikomatsakis:
One thing I did *not* do is extend the repl with an "first N answers" or "give all answers" mode for on-demand SLG, though such a thing would certainly be possible.

Also, we still need to implement answer abstraction for SLG.

Issue #70, opened 2017-11-10
StreamingIterator code panics with "zipping things of mixed type"


mikeyhew:
The following program panics with the error message "zipping things of mixed type", after outputting `a = '?0, b = Array3<i32>`.

```rust
trait StreamingIterator {
    type Item<'a>;
}

trait IntoStreamingIterator where Self::IntoIter: StreamingIterator {
    type Item<'a>;
    type IntoIter;
}

impl<S> IntoStreamingIterator for S where S: StreamingIterator {
    type Item<'a> = S::Item<'a>;
    type IntoIter = S;
}

struct i32 {}
struct usize {}

struct Ref<'a, T> {}

struct Array3<T> {_1: T, _2: T, _3: T}

struct Array3IntoIter<T> {
    array: Array3<T>,
    index: usize
}

impl<T> StreamingIterator for Array3IntoIter<T> {
    type Item<'a> = Ref<'a, T>;
}

impl<T> IntoStreamingIterator for Array3<T> {
    type Item<'a> = Ref<'a, T>;
    type IntoIter = Array3IntoIter<T>;
}

// goal: Array3<i32>: IntoStreamingIterator

```

output including backtrace

```text
$ RUST_BACKTRACE=1 cargo run -- --program=streaming_iterator.chalk --goal='Array3<i32>: IntoStreamingIterator'
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/chalki --program=streaming_iterator.chalk '--goal=Array3<i32>: IntoStreamingIterator'`
a = '?0, b = Array3<i32>
thread 'main' panicked at 'zipping things of mixed kind', src/zip.rs:260:16
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:69
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:381
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:397
   4: std::panicking::begin_panic
             at src/libstd/panicking.rs:577
   5: std::panicking::begin_panic
             at /Users/travis/build/rust-lang/rust/src/libstd/panicking.rs:538
[...]

PR #82, opened 2018-02-05
Implied bounds, second take

### LINKED ISSUE #11: Well-formedness checking of declarations
We don't currently do any WF checking. Specifically, it'd be nice to lower each struct/trait/impl into a predicate that, if provable, defines whether the declaration is well-formed. In *most* cases this is relatively straightforward, though depending how far we go we might need to extend the IR.

For example, structs don't currently list their fields, but the WF rules for structs require that the type of each field is WF. So if we had a struct:

```rust
struct Foo<T> {
    b: Bar<T>
}

struct Bar<T> where T: Eq { t: T }
```

Then the WF predicate for `Foo` might look like:

```
StructFooWF :-
    forall<T> {
        WF(Bar<T>)
    }.
```

This relies on us having the rules for when *types* are well-formed (which...I think we do? if not, should open a bug on that), which would look like:

```rust
WF(Bar<T>) :- T: Eq.
```

In this case, that would be unprovable, and hence the `struct Foo` is not considered well-formed (not without a `T: Eq` where-clause, at least). If we added the `T: Eq` where clause:

```rust
struct Foo<T> where T: Eq { b: Bar<T> }
```

then we would have:

```
StructFooWF :-
    forall<T> {
        if (T: Eq) {
            WF(Bar<T>)
        }
    }.
```

and everything is provable again.

### LINKED ISSUE #12: Elaborating where clauses on traits, structs

scalexm:
This is a full implementation of implied bounds as well as enforcing WF requirements on impls and type declarations.

The rules for WF requirements on traits are detailed in #12 (and in the code). The rules for types are much more simple (no need for co-recursion but we still have two predicates, `WellFormed` and `FromEnv`).

I was not greatly inspired for the tests, so one can feel free to add some more.

Fixes #11 and incidentally some form of bug related to #70. And eventually #12 will be resolved too.


src/ir/mod.rs
```diff
@@ -747,13 +790,19 @@ impl<T> UCanonical<T> {
 }
 
 impl UCanonical<InEnvironment<Goal>> {
-    /// A goal has coinductive semantics if it is of the form `T: AutoTrait`.
+    /// A goal has coinductive semantics if it is of the form `T: AutoTrait`, or if it is of the
+    /// form `WellFormed(T: Trait)` where `Trait` is any trait. The latter is needed for dealing
```
nikomatsakis: I am mildly surprised that we don't want to just make *all* `WellFormed` goals coinductive. But I guess it's only needed for traits...

src/lower/mod.rs
```diff
@@ -445,6 +447,10 @@ impl LowerWhereClause<ir::DomainGoal> for WhereClause {
             WhereClause::TraitRefWellFormed { ref trait_ref } => {
                 ir::WellFormed::TraitRef(trait_ref.lower(env)?).cast()
             }
+            WhereClause::TyFromEnv { ref ty } => ir::FromEnv::Ty(ty.lower(env)?).cast(),
```
nikomatsakis: I am torn on the question of what this should mean:

```
if (T: Trait) { .. }
```

I am thinking maybe we should make that equivalent to `if (FromEnv(T: Trait)) { .. }` and then have an explicit form `if (Implemented(T: Trait)) { .. }`. Basically `T: Trait` would parse to an AST form whose expansion depends on whether it appears in a "clause" (in which case it is `FromEnv`) or a Goal (in which case it is "Implemented").

nikomatsakis: (That said, I'm not really sure when we would want to say `if (Implemented(T: Trait)) { ... }`.. but maybe it's useful when crafting unit tests or something.)

src/lower/mod.rs
```diff
@@ -1242,103 +1240,107 @@ impl Anonymize for [ir::ParameterKind<ir::Identifier>] {
...
+                })
+            });
+        }
+
+        clauses
     }
 }
 
 impl ir::TraitDatum {
-    fn to_program_clauses(&self, program: &ir::Program) -> Vec<ir::ProgramClause> {
```
nikomatsakis: awesome that we don't need this parameter anymore

src/lower/mod.rs
```diff
@@ -1407,64 +1412,126 @@ impl ir::AssociatedTyDatum {
...
             }
-        };
+        });
 
-        vec![fallback_clause, normalize_clause]
+        // And the other one being:
+        //
+        //    forall<T> {
+        //        Normalize(<T as Foo>::Assoc -> U) :- FromEnv(T: Foo<Assoc = U>)
+        //    }
```
nikomatsakis: As an aside, and not for this PR, I wonder if we can use macros to make this code less tedious to write.

src/lower/test.rs
```diff
@@ -542,3 +544,221 @@ fn overlapping_negative_impls() {
...
+            trait Copy { }
+
+            trait A where Self: B, Self: Copy {}
+            trait B where Self: A { }
+
+            impl<T> B for T {}
+            impl<T> A for T where T: B {}
+        } error_msg {
+            "trait impl for \"B\" does not meet well-formedness requirements"
+        }
```
nikomatsakis: nice

nikomatsakis: (maybe we should add a test where this succeeds?)

src/lower/test.rs
```diff
@@ -542,3 +544,221 @@ fn overlapping_negative_impls() {
...
+            trait Foo { }
+            struct OnlyFoo<T> where T: Foo { }
+
+            struct i32 { }
+
+            trait Bar {
+                type Value;
+            }
+
+            impl Bar for i32 {
```
nikomatsakis: Maybe add a comment here: this fails because `i32: Foo` does not hold

src/lower/test.rs
```diff
@@ -542,3 +544,221 @@ fn overlapping_negative_impls() {
...
+
+            struct Set<K> where K: Hash { }
+
+            struct OnlyEq<T> where T: Eq { }
+
+            trait Foo {
+                type Value;
+            }
+
+            impl<K> Foo for Set<K> {
```
nikomatsakis: Here, `Wf(Set<K>)` implies `K: Hash` and hence `OnlyEq<K>` is WF

src/lower/wf.rs
```diff
@@ -0,0 +1,272 @@
...
+                proj.parameters.fold(accumulator);
+            }
+            Ty::UnselectedProjection(ref proj) => {
+                accumulator.push(self.clone());
+                proj.parameters.fold(accumulator);
+            }
+
+            // Type parameters and higher-kinded types do not carry any input types (so we can sort
+            // of assume they are always WF).
+            Ty::Var(..) | Ty::ForAll(..) => (),
```
nikomatsakis: Hmm, `Ty::ForAll` corresponds to `for<'a> fn(&'a u32)` or some such. I don't know that we can assume they are always WF. I think this accumulator probably doesn't want to accumulate into a vector, but rather to invoke a callback, because in the case of `for<'a>` we may want to create subuniverses? I have to read this code a bit more carefully I guess.

src/ir/mod.rs
scalexm: Well types only have where clauses of the form `T: Trait`, there are no `WF(Type)` where clauses so you never have the kinds of cycles encountered with traits.

src/lower/mod.rs
scalexm: Yeah not really sure either, but I think converting to `FromEnv(T: Trait)` in clauses is a good thing, and the explicit goal `Implemented` would replace the former `if_raw` statement in a much nicer way.

src/lower/wf.rs
scalexm: Actually we discussed that and for some reason we decided that higher-kinded types are always WF. This is similar to the fact that if we have this kind of function:
```rust
fn foo<T>(arg: HashSet<NotHash<T>>) { ... }
```
we would not issue any error until the function is called. For lifetime parametric function types, again an error would be issued only if a function of this type is ever called. We discussed that in our dropbox paper (see 2017.07.25).

src/solve/test/bench.rs
```diff
@@ -108,6 +109,7 @@ fn cycley_recursive_cached(b: &mut Bencher) {
         CYCLEY_GOAL,
         b,
         "Unique",
+        true
```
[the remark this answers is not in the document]
scalexm: The recursive solver fails to answer uniquely goals of the form `<Bar as Foo>::Assoc: SomeTrait` (or `WellFormed(<Bar as Foo>::Assoc: SomeTrait)`). So this is a temporary hack so that the benchmark tests pass, but I should add a `FIXME` or something because this is really not ought to stay.

src/lower/wf.rs
nikomatsakis: Oh, right, I forgot about that. Basically we were assuming that for all types introduce their *own* implied bounds, right? Makes sense.

nikomatsakis: Can we put a more explicit comment then? =)

nikomatsakis:
@scalexm ok so the changes to make then are:

- [ ] add an explicit `Implemented(T: Trait)` in the AST
- [ ] change `T: Trait` to lower to `FromEnv` in a clause
- [ ] add a few comments:
    - some in the tests
    - but most importantly on the treatment of `ForAll`

Do you think you'll have time for that? Do you want me to do it? We should probably make more tests but I'm not inclined to block on that.

nikomatsakis:
@scalexm 

I read over the changes more thoroughly. Everything looks great. I looked briefly at the on-demand test was failing; for some reason it is now creating `FromEnv` tables (which fail to get proven). This is not unexpected, since you changed the program clauses. Seems fine.

However, I did notice one thing. It seems like you are treating `T: Iterator<Item = U>` as a "normalize" rule, but in fact it parsed as a "projection equality" rule:

https://github.com/rust-lang-nursery/chalk/blob/4defa33befbffd9239c5c293f3e56ec4e8fd810e/chalk-parse/src/parser.lalrpop#L201-L210

The intention is that "projection eq" is the "normal" rule, with `Normalize` being something that only comes from impls (i.e., trans could use it to find the "true" type).

PR #87, opened 2018-03-05
Extract SLG solver into its own crate

nikomatsakis:
The history is long and tortured, but it worked out pretty well!

The `chalk_src/context/mod.rs` crate defines the interface to the rest of the world. Can probably be simplified somewhat further. Should be reusable in rustc though.

cc @scalexm


chalk-slg/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+impl<C: Context> PartialEq for Literal<C> {
+    fn eq(&self, other: &Literal<C>) -> bool {
+        match (self, other) {
+            (Literal::Positive(goal1), Literal::Positive(goal2))
+            | (Literal::Negative(goal1), Literal::Negative(goal2)) => goal1 == goal2,
+
+            _ => false,
+        }
+    }
+}
```
scalexm: Is there a reason for not `#[derive]`-ing these impls?

nikomatsakis: Gah. I'm torn. On the one hand, they are not *needed*, and it requires deriving a bunch of ord and crap elsewhere that are also not *needed*.

On the *other* hand, writing the impls by hand is a real pain. After having tried that for long enough, I'm maybe inclined to just add the derives.

nikomatsakis: actually I was being silly. Fixed.

src/fold/mod.rs
```diff
@@ -365,7 +365,7 @@ macro_rules! copy_fold {
         impl ::fold::Fold for $t {
             type Result = Self;
             fn fold_with(&self,
-                         _folder: &mut ::fold::Folder,
+                         _folder: &mut dyn (::fold::Folder),
```
nikomatsakis: @Manishearth -- well, I hit the edge case now =)

Manishearth: Oh, that edge case is fine, it's the other way around. Will people ever be using `dyn::foo` as a path?

nikomatsakis: > Will people ever be using dyn::foo as a path?

Who knows...

Issue #81, opened 2018-02-05
On-demand solver can often stop sooner


nikomatsakis:
Currently, when asked to `solve`, the on-demand solver requests answers until the aggregate result becomes "trivial" (i.e., no information is gained). This occurs in `make_aggregate`:

https://github.com/rust-lang-nursery/chalk/blob/b0d1417104ccdc52a72b96d2fe89e56a1e284e13/src/solve/slg/aggregate.rs#L15-L20

Notably right here:

https://github.com/rust-lang-nursery/chalk/blob/b0d1417104ccdc52a72b96d2fe89e56a1e284e13/src/solve/slg/aggregate.rs#L57-L59

This is the key to e.g. why the [`only_demand_so_many` test](https://github.com/rust-lang-nursery/chalk/blob/b0d1417104ccdc52a72b96d2fe89e56a1e284e13/src/solve/slg/on_demand/test.rs#L227-L261) only, well, demands "so many" answers. But the current check is not that smart. 

For example, the goal `exists<T> { T: Foo }` on this input will still enumerate a ton of answers:

```rust
trait Sized { }
trait Foo { }

struct Vec<T> { }
impl<T> Sized for Vec<T> where T: Sized { }
impl<T> Foo for Vec<T> where T: Sized { }

struct i32 { }
impl Sized for i32 { }

struct Slice<T> { }
impl<T> Sized for Slice<T> where T: Sized { }
```

The problem here is that the stream of answers looks like:

- `Vec<i32>`
- `Vec<Vec<i32>>`
- `Vec<Slice<i32>>`
- `Vec<Vec<Vec<i32>>`
- ...

Hence our aggregate suggestion will quickly reach `Vec<_>` and stay there. It will never become truly trivial.

But we can easily test this! We could for example, before requesting the next solution, check whether there exist *any* strands that could violate the current aggregate (which would be `Vec<_>`). If not, we don't need to request the next answer, it won't give us anything new we've not already seen.

nikomatsakis:
### Background notes

Here are some notes on what we would want to modify. The basic idea is to extend the interface that the forest uses to give back answers: right now it's just an iterator. I'd like to add an extra method that lets us provide a substitution and find out if there are any available strands that might lead to a result that is not unifiable with the current substitution.

The current interface for getting answers is to invoke `solve`:

[...]

PR #93, opened 2018-03-13
Implement early stop for solver when no strands can invalidate the current substitution.

aatxe:
This should hopefully be a working implementation of #81.

### DIFF
--- a/chalk-engine/src/context/mod.rs
+++ b/chalk-engine/src/context/mod.rs
@@ -157,7 +157,7 @@ pub trait AggregateOps<C: Context> {
     fn make_solution(
         &self,
         root_goal: &C::CanonicalGoalInEnvironment,
-        simplified_answers: impl IntoIterator<Item = SimplifiedAnswer<C>>,
+        simplified_answers: impl AnswerStream<C>,
     ) -> Option<C::Solution>;
 }
 
@@ -336,4 +336,16 @@ pub trait UnificationResult<C: Context, I: InferenceContext<C>> {
     /// Add the residual subgoals as new subgoals of the ex-clause.
     /// Also add region constraints.
     fn into_ex_clause(self, ex_clause: &mut ExClause<C, I>);
+}
+
+pub trait AnswerStream<C: Context> {
+    fn peek_answer(&mut self) -> Option<SimplifiedAnswer<C>>;
+    fn next_answer(&mut self) -> Option<SimplifiedAnswer<C>>;
+
+    /// Invokes `test` with each possible future answer, returning true immediately
+    /// if we find any answer for which `test` returns true.
+    fn any_future_answer(
+        &mut self,
+        test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+    ) -> bool;
 }
--- a/chalk-engine/src/context/mod.rs
+++ b/chalk-engine/src/context/mod.rs
@@ -291,7 +291,7 @@ pub trait Environment<C: Context, I: InferenceContext<C>>: Debug + Clone {
 
 pub trait CanonicalExClause<C: Context>: Debug {
     /// Extracts the inner normalized substitution.
-    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
+    fn inference_normalized_subst(&self) -> &C::InferenceNormalizedSubst;
 }
 
 pub trait CanonicalConstrainedSubst<C: Context>: Clone + Debug + Eq + Hash + Ord {
@@ -299,7 +299,7 @@ pub trait CanonicalConstrainedSubst<C: Context>: Clone + Debug + Eq + Hash + Ord
     fn empty_constraints(&self) -> bool;
 
     /// Extracts the inner normalized substitution.
-    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
+    fn inference_normalized_subst(&self) -> &C::InferenceNormalizedSubst;
 }
 
 pub trait DomainGoal<C: Context, I: InferenceContext<C>>: Debug {
@@ -358,6 +358,6 @@ pub trait AnswerStream<C: Context> {
     /// if we find any answer for which `test` returns true.
     fn any_future_answer(
         &mut self,
-        test: impl FnMut(&mut C::InferenceNormalizedSubst) -> bool,
+        test: impl FnMut(&C::InferenceNormalizedSubst) -> bool,
     ) -> bool;
 }
--- a/chalk-engine/src/context/mod.rs
+++ b/chalk-engine/src/context/mod.rs
@@ -8,7 +8,7 @@ crate mod prelude;
 
 /// The "context" in which the SLG solver operates.
 pub trait Context: Sized + Clone + Debug + ContextOps<Self> + AggregateOps<Self> {
-    type CanonicalExClause: Debug;
+    type CanonicalExClause: CanonicalExClause<Self>;
 
     /// A map between universes. These are produced when
     /// u-canonicalizing something; they map canonical results back to
@@ -20,6 +20,10 @@ pub trait Context: Sized + Clone + Debug + ContextOps<Self> + AggregateOps<Self>
     ///
     /// [the rustc-guide]: https://rust-lang-nursery.github.io/rustc-guide/traits-canonicalization.html#canonicalizing-the-query-result
     type CanonicalConstrainedSubst: CanonicalConstrainedSubst<Self>;
+
+    /// Extracted from a canonicalized substitution or canonicalized ex clause, this is the type of
+    /// substitution that is fully normalized with respect to inference variables.
+    type InferenceNormalizedSubst: Debug;
 
     /// A canonicalized `GoalInEnvironment` -- that is, one where all
     /// free inference variables have been bound into the canonical
@@ -285,9 +289,17 @@ pub trait Environment<C: Context, I: InferenceContext<C>>: Debug + Clone {
     fn add_clauses(&self, clauses: impl IntoIterator<Item = I::ProgramClause>) -> Self;
 }
 
+pub trait CanonicalExClause<C: Context>: Debug {
+    /// Extracts the inner normalized substitution.
+    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
+}
+
 pub trait CanonicalConstrainedSubst<C: Context>: Clone + Debug + Eq + Hash + Ord {
     /// True if this solution has no region constraints.
     fn empty_constraints(&self) -> bool;
+
+    /// Extracts the inner normalized substitution.
+    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
 }
 
 pub trait DomainGoal<C: Context, I: InferenceContext<C>>: Debug {
@@ -346,6 +358,6 @@ pub trait AnswerStream<C: Context> {
     /// if we find any answer for which `test` returns true.
     fn any_future_answer(
         &mut self,
-        test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+        test: impl FnMut(&mut C::InferenceNormalizedSubst) -> bool,
     ) -> bool;
 }
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -107,15 +107,15 @@ impl<C: Context> Forest<C> {
         &mut self,
         table: TableIndex,
         answer: AnswerIndex,
-        mut test: impl FnMut(&mut C::InferenceNormalizedSubst) -> bool,
+        mut test: impl FnMut(&C::InferenceNormalizedSubst) -> bool,
     ) -> bool {
         if let Some(answer) = self.tables[table].answer(answer) {
             info!("answer cached = {:?}", answer);
-            return test(&mut answer.subst.inference_normalized_subst());
+            return test(answer.subst.inference_normalized_subst());
         }
 
         self.tables[table].strands_mut().any(|strand| {
-            test(&mut strand.canonical_ex_clause.inference_normalized_subst())
+            test(strand.canonical_ex_clause.inference_normalized_subst())
         })
     }
 
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -106,20 +106,17 @@ impl<C: Context> Forest<C> {
     pub(super) fn any_future_answer(
         &mut self,
         table: TableIndex,
-        mut test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+        answer: AnswerIndex,
+        mut test: impl FnMut(&mut C::InferenceNormalizedSubst) -> bool,
     ) -> bool {
-        let mut strands = self.tables[table].strands_mut();
-
-        // If we don't have any strands at all, then we are
-        // intentionally conservative and say it may_invalidate.
-        match strands.next() {
-            None => return true,
-            Some(strand) => if test(&mut strand.canonical_ex_clause) {
-                return true
-            },
+        if let Some(answer) = self.tables[table].answer(answer) {
+            info!("answer cached = {:?}", answer);
+            return test(&mut answer.subst.inference_normalized_subst());
         }
 
-        strands.any(|strand| test(&mut strand.canonical_ex_clause))
+        self.tables[table].strands_mut().any(|strand| {
+            test(&mut strand.canonical_ex_clause.inference_normalized_subst())
+        })
     }
 
     /// Ensures that answer with the given index is available from the
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -103,6 +103,25 @@ impl<C: Context> Forest<C> {
         }
     }
 
+    pub(super) fn any_future_answer(
+        &mut self,
+        table: TableIndex,
+        mut test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+    ) -> bool {
+        let mut strands = self.tables[table].strands_mut();
+
+        // If we don't have any strands at all, then we are
+        // intentionally conservative and say it may_invalidate.
+        match strands.next() {
+            None => return true,
+            Some(strand) => if test(&mut strand.canonical_ex_clause) {
+                return true
+            },
+        }
+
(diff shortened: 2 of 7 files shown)


src/solve/slg/implementation/aggregate.rs
```diff
@@ -15,18 +15,16 @@ impl context::Aggregate<SlgContext> for SlgContext {
...
         &self,
         root_goal: &Canonical<InEnvironment<Goal>>,
-        simplified_answers: impl IntoIterator<Item = SimplifiedAnswer<SlgContext>>,
+        mut simplified_answers: impl context::AnswerStream<SlgContext>,
     ) -> Option<Solution> {
-        let mut simplified_answers = simplified_answers.into_iter().peekable();
-
         // No answers at all?
-        if simplified_answers.peek().is_none() {
+        if simplified_answers.peek_answer().is_none() {
```
nikomatsakis: ah, I forgot about `peek`. Nice.

src/solve/slg/implementation/mod.rs
```diff
@@ -233,7 +233,14 @@ impl context::Environment<SlgContext> for Arc<Environment> {
     }
 }
 
-impl context::Substitution<SlgContext> for Substitution {}
+impl context::Substitution<SlgContext> for Substitution {
+    fn may_invalidate(&mut self, subst: &Substitution) -> bool {
+        self.parameters
+            .iter()
+            .zip(&subst.parameters)
+            .any(|(param1, param2)| param1 != param2)
```
nikomatsakis: OK, so, yeah this doesn't feel right. `!=`  is always kinda curious =)

nikomatsakis:
Oh, so, first thing before anything else: we need a test.

nikomatsakis:
As for the test, I think we want a variant of the `only_draw_so_many` test:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/solve/slg/test.rs#L251-L285

If we adapt the [program text](https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/solve/slg/test.rs#L255-L266) to something like this:

```rust
trait Sized { }
trait Foo { }

struct Vec<T> { }
impl<T> Sized for Vec<T> where T: Sized { }
impl<T> Foo for Vec<T> where T: Sized { }

struct i32 { }
impl Sized for i32 { }

struct Slice<T> { }
impl<T> Sized for Slice<T> where T: Sized { }
```

with the goal `exists<T> { T: Foo }`, then we ought to see (without your changes, that is) that it draws way too many items. With your changes it ought to draw some small-ish number =)

nikomatsakis:
We probably want to factor out some common infrastructure there, btw -- e.g., a helper where you give the program text, a goal, and the number of answers we expect to draw. That way we can add more tests.

nikomatsakis:
Regarding the inequality test, the anti-unification code here kind of does a similar thing:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/solve/slg/implementation/aggregate.rs#L120-L125

That said, I don't have a specific example where your code "goes wrong" -- maybe by being more conservative than necessary? -- but it feels like we should be able to create some. Let's start by getting the tests =)

src/solve/slg/implementation/aggregate.rs
```diff
@@ -53,11 +51,12 @@ impl context::Aggregate<SlgContext> for SlgContext {
         // limited for that, but the on-demand SLG solver probably could
         // give us that information.
         let guidance = loop {
-            if subst.value.is_empty() || is_trivial(&subst) {
+            if subst.value.is_empty() || is_trivial(&subst)
+                || !simplified_answers.may_invalidate(&subst.value) {
```
aatxe: @nikomatsakis this seems to have been a bug! I believe we should return `Guidance::Definite(subst)` when `may_invalidate` is false.

(pinged on gitter but leaving it here too as documentation)

aatxe: @nikomatsakis this seems to have been a bug! I believe we should return `Guidance::Definite(subst)` when `may_invalidate` is false.

(pinged on gitter but leaving it here too as documentation)

chalk-slg/src/logic.rs
```diff
@@ -103,6 +103,25 @@ impl<C: Context> Forest<C> {
...
 
+    pub(super) fn may_invalidate(
+        &mut self,
+        table: TableIndex,
+        subst: &C::Substitution,
+    ) -> bool {
+        let mut strands = self.tables[table].strands_mut();
+
+        // If we don't have any strands at all, then we are
+        // intentionally conservative and say it may_invalidate.
```
nikomatsakis: Say more about this? It surprises me a bit. Well, I guess it doesn't hurt -- they will request the next answer then, and get back nothing -- but I'm not sure why it's worth making a special case for it.

src/solve/slg/implementation/aggregate.rs
```diff
@@ -57,7 +55,11 @@ impl context::Aggregate<SlgContext> for SlgContext {
                 break Guidance::Unknown;
```
nikomatsakis: Yep, returning `Unknown` when nothing could invalidate was definitely wrong

chalk-slg/src/context/mod.rs
```diff
@@ -226,7 +226,9 @@ pub trait InferenceTable<C: Context>: Clone {
     ) -> Fallible<Self::UnificationResult>;
 }
 
-pub trait Substitution<C: Context>: Clone + Debug {}
+pub trait Substitution<C: Context>: Clone + Debug {
+    fn may_invalidate(&mut self, subst: &C::Substitution) -> bool;
```
nikomatsakis: I think it's worth commenting this a bit. Something like this:

---

Tests whether this substitution -- if it were to become a confirmed answer -- would affect `subst`, which represents the aggregated results of all answers thus far. If this returns true, then the outer solving loop has a reason to continue requesting answers -- but if all strands return false, it does not, as `subst` will not be further refined. 

Be careful here: `self` and `subst` exist in distinct inference tables.

---

Actually, the latter point is making me wonder: I think we probably have to thread through the inference tables, and this is making me think I may prefer the "callback variant", in which this method doesn't exist (more below).

chalk-slg/src/context/mod.rs
```diff
@@ -264,3 +266,11 @@ pub trait UniverseMap<C: Context>: Clone + Debug {
 pub trait UnificationResult<C: Context> {
     fn into_ex_clause(self, ex_clause: &mut ExClause<C>);
 }
+
+pub trait AnswerStream<C: Context> {
+    fn peek_answer(&mut self) -> Option<SimplifiedAnswer<C>>;
+    fn next_answer(&mut self) -> Option<SimplifiedAnswer<C>>;
+
+    /// True if there are any strands that could yield an answer invalidating `subst`
+    fn may_invalidate(&mut self, subst: &C::Substitution) -> bool;
```
nikomatsakis: In the callback variant, this would become:

```rust
/// Invokes `test` with each possible future answer, returning true immediately
/// if we find any answer for which `test` returns true.
fn any_future_answer(
    &mut self, 
    test: impl FnMut(&C::Substitution, &mut C::InferenceTable) -> bool,
) -> bool;
```

src/solve/slg/implementation/mod.rs
```diff
@@ -233,7 +233,176 @@ impl context::Environment<SlgContext> for Arc<Environment> {
...
+            }
+
+            (Ty::Var(_), ty) | (ty, Ty::Var(_)) => {
+                !self.recurse_ty(ty)
+            }
+
+            (Ty::ForAll(_), _)
+                | (Ty::Apply(_), _)
+                | (Ty::Projection(_), _)
+                | (Ty::UnselectedProjection(_), _) => true,
```
nikomatsakis: PS, run rustfmt, I *think* this is not the style it produces?

src/solve/slg/implementation/mod.rs
```diff
@@ -233,7 +233,176 @@ impl context::Environment<SlgContext> for Arc<Environment> {
...
+            .zip(&subst.parameters)
+            .any(|(p1, p2)| MayInvalidate.aggregate_parameters(p1, p2))
+    }
+}
+
+// This is a struct in case we need to add state at any point like in AntiUnifier
+struct MayInvalidate;
+
+impl MayInvalidate {
+    fn aggregate_parameters(&mut self, p1: &Parameter, p2: &Parameter) -> bool {
```
nikomatsakis: I think we should give the left- and right-hand-side more meaningful names. In particular:

- `p1` represents a potential new answer
- `p2` represents the aggregated answer thus far

right?

So maybe `new` and `current` as names (we could use same name no matter the type). Or else `new_` and `current_` as prefixes.

src/solve/slg/implementation/mod.rs
```diff
@@ -233,7 +233,176 @@ impl context::Environment<SlgContext> for Arc<Environment> {
...
+            (ParameterKind::Ty(_), _) | (ParameterKind::Lifetime(_), _) => {
+                panic!("mismatched parameter kinds: p1={:?} p2={:?}", p1, p2)
+            }
+        }
+    }
+
+    // Returns true if the two types could be unequal.
+    fn aggregate_tys(&mut self, ty0: &Ty, ty1: &Ty) -> bool {
+        match (ty0, ty1) {
+            (Ty::Var(_), Ty::Var(_)) => false,
```
nikomatsakis: A couple of things here. First, I think we need to thread the inference table for the new answer to this code -- and maybe we ought to thread the inference table for the current answer too, for uniformity (though I think the aggregator never unifies, so it's not strictly needed in that case). That way we can begin with a "normalizing" step to get rid of inference variables that have bound values.
Then we can start this function with a "shallow normalize" step like so:

```rust
if let Some(ty0) = self.infer0.normalize_shallow(ty0) {
    return self.aggregate_tys(ty0, ty1);
} else if let Some(ty1) = self.infer1.normalize_shallow(ty1) {
    return self.aggregate_tys(ty0, ty1);
}
```

just like what you see here in unification (and you will find elsewhere):

https://github.com/rust-lang-nursery/chalk/blob/d03d8f9e7bfb90d2fdfec86b5f15a7ce432547e5/src/solve/infer/unify.rs#L92-L95

Next, once we've done that, we can tweak this a bit further:

```rust
match (new, current) { // using the names I suggested above =)
    (_, Ty::Var(_)) => {
        // If the aggregate solution already has an inference variable here,
        // then no matter what type we produce, the aggregate cannot get 'more generalized'
        // than it already is. So return false, we cannot invalidate.
        false
    }

    (Ty::Var(_), _) => {
        // If we see a type variable in the potential future solution, we have to be conservative.
        // We don't know what type variable will wind up being! Remember that the future solution could
        // be any instantiation of `ty0` -- or it could leave this variable unbound, if the result is true for
        // all types.
        true
    }

    ...
}
```

nikomatsakis: I have to think though about how to make a test that will cause the existing code to go wrong =)

chalk-slg/src/context/mod.rs
nikomatsakis: Er I forgot to follow up on this. The main thing is that we would invoke with a closure which *still* calls your existing "invalidate" code, but that closure has access to the expected substitution from its environment (along with its inference table). It seems a bit more "neat" to thread down just a closure though.

src/solve/slg/implementation/mod.rs
aatxe: Will do! I don't normally use it, so I had just copied existing style (this format was done in the AntiUnifier).

chalk-slg/src/logic.rs
aatxe: I'm not yet entirely sure why, but without this check, there were two tests that failed, `solve::test::prove_infer` and `solve::test::suggested_subst`, which appears to be because they expected to get `Unknown` but were instead giving `subst`. I think the other option is a change in `make_solution`, but I'm not positive what that looks like.

src/solve/slg/implementation/mod.rs
nikomatsakis: I don't think it's indented quite as much there; but yeah I prefr to run rustfmt...

chalk-slg/src/logic.rs
nikomatsakis: curious

nikomatsakis: OK, I did a bit of investigating. I found the problem. The problem is that our main loop invokes `peek_answer` and then `may_invalidate` -- but if `peek_answer` gets an answer already, then there is no strand left for `may_invalidate` to examine. So it is correct that no strands will invalidate, but incorrect that no *future answers* may do so. 

To fix this, we should pass in the current answer index to `may_invalidate`. This is also needed -- now that I think about -- for the caching scenario I expect in the future, where the answers have already been computed as part of some other query and we are just replaying them.

nikomatsakis:
ping @aatxe -- any thoughts on this?

aatxe:
Sorry, I'll work more on this this weekend!

chalk-engine/src/logic.rs
```diff
@@ -103,6 +103,25 @@ impl<C: Context> Forest<C> {
         }
     }
 
+    pub(super) fn any_future_answer(
+        &mut self,
+        table: TableIndex,
```
nikomatsakis: so I think this would take an `answer: AnswerIndex`, to start

chalk-engine/src/logic.rs
```diff
@@ -103,6 +103,25 @@ impl<C: Context> Forest<C> {
...
+        mut test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+    ) -> bool {
+        let mut strands = self.tables[table].strands_mut();
+
```
nikomatsakis: and here if there is an answer with that index we would invoke `test`:

```rust
         // First, check for a tabled answer.
         if self.tables[table].answer(answer).is_some() {
             info!("answer cached = {:?}", self.tables[table].answer(answer));
             return test(...);
    }
```

nikomatsakis: but there is a problem because it is not a canonical ex clause, but rather a substitution

chalk-engine/src/context/mod.rs
```diff
@@ -346,6 +358,6 @@ pub trait AnswerStream<C: Context> {
     /// if we find any answer for which `test` returns true.
     fn any_future_answer(
         &mut self,
-        test: impl FnMut(&mut C::CanonicalExClause) -> bool,
+        test: impl FnMut(&mut C::InferenceNormalizedSubst) -> bool,
```
nikomatsakis: Probably `&C::InferenceNormalizedSubst` would suffice, right?

chalk-engine/src/context/mod.rs
```diff
@@ -285,9 +289,17 @@ pub trait Environment<C: Context, I: InferenceContext<C>>: Debug + Clone {
     fn add_clauses(&self, clauses: impl IntoIterator<Item = I::ProgramClause>) -> Self;
 }
 
+pub trait CanonicalExClause<C: Context>: Debug {
+    /// Extracts the inner normalized substitution.
+    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
```
nikomatsakis: This should return `&C::InferenceNormalizedSubst`

chalk-engine/src/context/mod.rs
```diff
@@ -285,9 +289,17 @@ pub trait Environment<C: Context, I: InferenceContext<C>>: Debug + Clone {
...
+}
+
 pub trait CanonicalConstrainedSubst<C: Context>: Clone + Debug + Eq + Hash + Ord {
     /// True if this solution has no region constraints.
     fn empty_constraints(&self) -> bool;
+
+    /// Extracts the inner normalized substitution.
+    fn inference_normalized_subst(&self) -> C::InferenceNormalizedSubst;
```
nikomatsakis: Same here

src/solve/slg/implementation/mod.rs
```diff
@@ -413,10 +412,20 @@ impl context::DomainGoal<SlgContext, SlgContext> for DomainGoal {
     }
 }
 
+impl context::CanonicalExClause<SlgContext> for Canonical<ExClause<SlgContext, SlgContext>> {
+    fn inference_normalized_subst(&self) -> Substitution {
+        self.value.subst.clone()
```
nikomatsakis: If you changed signature in trait, this could be `&self.value.subst`

src/solve/slg/implementation/mod.rs
```diff
@@ -413,10 +412,20 @@ impl context::DomainGoal<SlgContext, SlgContext> for DomainGoal {
...
+    }
+}
+
 impl context::CanonicalConstrainedSubst<SlgContext> for Canonical<ConstrainedSubst> {
     fn empty_constraints(&self) -> bool {
         self.value.constraints.is_empty()
     }
+
+    fn inference_normalized_subst(&self) -> Substitution {
+        self.value.subst.clone()
```
nikomatsakis: Same here

chalk-engine/src/logic.rs
```diff
@@ -106,20 +106,17 @@ impl<C: Context> Forest<C> {
...
-        // If we don't have any strands at all, then we are
-        // intentionally conservative and say it may_invalidate.
-        match strands.next() {
-            None => return true,
-            Some(strand) => if test(&mut strand.canonical_ex_clause) {
-                return true
-            },
+        if let Some(answer) = self.tables[table].answer(answer) {
+            info!("answer cached = {:?}", answer);
+            return test(&mut answer.subst.inference_normalized_subst());
```
nikomatsakis: If you change trait, this `&mut` can go away.

chalk-engine/src/logic.rs
```diff
@@ -106,20 +106,17 @@ impl<C: Context> Forest<C> {
...
         }
 
-        strands.any(|strand| test(&mut strand.canonical_ex_clause))
+        self.tables[table].strands_mut().any(|strand| {
+            test(&mut strand.canonical_ex_clause.inference_normalized_subst())
```
nikomatsakis: And this one.

PR #95, opened 2018-03-14
Delete the recursive solver

### LINKED ISSUE #90: remove recursive solver
It seems like it's time to get rid of Chalk's "recursive" solver and commit to the on-demand SLG solver. It's basically strictly better at this point.

The recursive solver code is found here:

https://github.com/rust-lang-nursery/chalk/tree/master/src/solve/recursive

We'll also want to remove this `SolverChoice` variant:

https://github.com/rust-lang-nursery/chalk/blob/82a8472ca9f6756d160b639dfc68f6a69076b716/src/solve/mod.rs#L183-L187

Which will entail then modifying the unit tests. They mostly run on both solvers, as controlled here:

https://github.com/rust-lang-nursery/chalk/blob/82a8472ca9f6756d160b639dfc68f6a69076b716/src/solve/test/mod.rs#L72

But sometimes the tests have distinct results for the recursive vs SLG solvers:

https://github.com/rust-lang-nursery/chalk/blob/82a8472ca9f6756d160b639dfc68f6a69076b716/src/solve/test/mod.rs#L932-L937

And of course we'll have to modify `chalki`, the command-line interpreter too. We can just remove this arm:

https://github.com/rust-lang-nursery/chalk/blob/82a8472ca9f6756d160b639dfc68f6a69076b716/src/bin/chalki.rs#L226-L228

though we may just want to remove the [whole `--solver` option](https://github.com/rust-lang-nursery/chalk/blob/82a8472ca9f6756d160b639dfc68f6a69076b716/src/bin/chalki.rs#L37).

aravind-pg:
Now uses the SLG solver for everything. Closes #90.

I was a bit surprised at the amount of code that became dead as a result of the deletion (see the second commit, 13f6ffb937ca6cc7d23444602fcae089413dcda7) -- this is definitely worth double-checking.


src/solve/mod.rs
```diff
@@ -180,12 +67,6 @@ impl fmt::Display for Solution {
 
 #[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq, Hash)]
 pub enum SolverChoice {
```
aravind-pg: This is now a single-variant enum, which is a bit funky, but I left it the way it is since it actually reads pretty well IMO.

src/solve/mod.rs
```diff
@@ -42,118 +42,6 @@ pub enum Guidance {
 }
 
 impl Solution {
-    /// There are multiple candidate solutions, which may or may not agree on
```
nikomatsakis: Yeah, this is dead because the chalk solver doesn't (yet?) support this kind of "advice". That is, the recursive solver would sometimes give advise to prefer answers derived from where clauses where there was ambiguity -- the slg solver doesn't draw this distinction. That's ok (imo), I'd prefer to start without it and add it back as we need to for compatibility.

(Still, worth noting, cc @scalexm)

src/solve/mod.rs
nikomatsakis: Seems ok. We may add more options later or something. (Or we can always clean it up later.)

src/solve/mod.rs
scalexm: Yes, there are still some `FIXME` anyway, e.g. https://github.com/rust-lang-nursery/chalk/blob/master/src/solve/test/mod.rs#L1213

PR #100, opened 2018-03-22
Publish and simplify bounds

nikomatsakis:
I published [`chalk-engine` v0.1.0 to `crates.io` ](https://crates.io/crates/chalk-engine). The branch adds various bits of metadata for that.

Also, because I was being sloppy with branch hygiene, I made various simplifications to the bounds in the context trait, paring down the bounds to the minimal things that it seemed like chalk-engine needs. This is to be published as v0.2.0.

Actually, I'm a bit unsure about this -- for example, a number of things are now just `Debug`, but I see no reason that they *shouldn't* be cloned or hashed, we just don't happen to have a need for it. I was debating about setting `Debug + Clone + Eq + Ord + Hash` as the default bounds everywhere, but ultimately decided against it for now. We can always add those later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,7 +77,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "chalk"
 version = "0.1.0"
 dependencies = [
- "chalk-engine 0.1.0",
+ "chalk-engine 0.2.0",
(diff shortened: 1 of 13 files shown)


nikomatsakis:
gonna merge :)

PR #102, opened 2018-03-26
Add `WhereClauseAtom` abstraction

scalexm:
Actually I'll start with this small change and see if I can help @PramodBisht on their side.

### DIFF
--- a/src/fold/mod.rs
+++ b/src/fold/mod.rs
@@ -229,9 +229,9 @@ tuple_fold!(A, B, C, D, E);
 impl<T: Fold> Fold for Option<T> {
     type Result = Option<T::Result>;
     fn fold_with(&self, folder: &mut dyn Folder, binders: usize) -> Fallible<Self::Result> {
-        match *self {
+        match self {
             None => Ok(None),
-            Some(ref e) => Ok(Some(e.fold_with(folder, binders)?)),
+            Some(e) => Ok(Some(e.fold_with(folder, binders)?)),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/fold/mod.rs
```diff
@@ -423,9 +423,9 @@ macro_rules! enum_fold {
 
 enum_fold!(PolarizedTraitRef[] { Positive(a), Negative(a) });
 enum_fold!(ParameterKind[T,L] { Ty(a), Lifetime(a) } where T: Fold, L: Fold);
-enum_fold!(DomainGoal[] { Implemented(a), ProjectionEq(a), Normalize(a), UnselectedNormalize(a), WellFormed(a), FromEnv(a), InScope(a) });
-enum_fold!(WellFormed[] { Ty(a), TraitRef(a), ProjectionEq(a) });
-enum_fold!(FromEnv[] { Ty(a), TraitRef(a), ProjectionEq(a) });
+enum_fold!(WhereClauseAtom[] { Implemented(a), ProjectionEq(a) });
+enum_fold!(DomainGoal[] { Holds(a), WellFormed(a), FromEnv(a), Normalize(a), UnselectedNormalize(a),
```
nikomatsakis: as an aside, I think it was @varkor who mentioned that we should make the rustc-guide match up with this setup more closely

src/ir/mod.rs
```diff
@@ -447,17 +447,74 @@ impl PolarizedTraitRef {
...
+    /// For example, given the following type definition:
+    ///
+    /// ```notrust
+    /// struct Set<K> where K: Hash {
+    ///     ...
+    /// }
+    /// ```
+    ///
+    /// then we have the following rule: `WellFormedTy(Set<K>) :- Implemented(K: Hash)`.
+    WellFormedTy(Ty),
```
nikomatsakis: Why is this .. oh, it's not a `WhereClauseAtom` because `Holds(Ty)` makes no sense? I wonder if it'd be worth adding it just for symmetry. We just happen to have no rules that establish `HoldsTy` =)

scalexm: Yeah I don't know, actually `WF(ty)` is a bit different from `WF(Trait)`: it is not co-inductive and it does need to "recursively" prove bounds (just the bounds directly expressed on the type). I guess: let's keep it separated for the moment.

PR #103, opened 2018-03-27
Extend `ProgramClause` definition

### LINKED ISSUE #94: extend the definition of "clause" that chalk uses
The first-order hereditary harrop predicates that Chalk uses allow for a fairly flexible definition of program clause:

    Clause = DomainGoal | Clause && Clause | Goal => DomainGoal | ForAll { Clause }

The intention then is that implications in `Goal` can reference these clauses in their full generality:

    Goal = ... | Clause => Goal | ...

However, Chalk currently uses a rather simpler definition:

https://github.com/rust-lang-nursery/chalk/blob/master/src/ir/mod.rs#L776

This is equivalent to the following:

    Clause = DomainGoal | Clause && Clause

We should generalize this to the full form -- or at least include `ForAll` binders. This will require a few changes. For one thing, we'll have to change environments to store clauses, rather than just domain goals:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/ir/mod.rs#L74-L79

Then we have to update the `chalk-slg` `HhGoal` type in an analogous fashion:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/chalk-slg/src/hh.rs#L10

Presumably extending the `Context` trait with the notion of a clause as well:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/chalk-slg/src/context/mod.rs#L9-L11

Finally, we have to modify the `Context::program_clauses` implementation:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/solve/slg/implementation/mod.rs#L69-L73

In particular, this is the code that finds hypotheses from the environment:

https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/solve/slg/implementation/mod.rs#L74-L78

Actually, this code is probably fine as is, we just have to (a) implement [`CouldMatch`](https://github.com/rust-lang-nursery/chalk/blob/2575275d42a94589b9ab4263e6b62d076b66d166/src/ir/could_match.rs#L5-L7) for `Clause` and (b) implement `into_program_clause` for `CouldMatch`.

(Note: It might be good to do https://github.com/rust-lang-nursery/chalk/issues/90 first, just to limit the amount of code affected.)

scalexm:
Add support for fully general clauses in the environment. Fixes #94.

### DIFF
--- a/src/solve/slg/implementation/resolvent.rs
+++ b/src/solve/slg/implementation/resolvent.rs
@@ -84,7 +84,12 @@ impl context::ResolventOps<SlgContext, SlgContext> for TruncatingInferenceTable 
         let ProgramClauseImplication {
             consequence,
             conditions,
-        } = self.infer.instantiate_binders_existentially(&clause.implication);
+        } = match clause {
+            ProgramClause::Implies(implication) => implication.clone(),
+            ProgramClause::ForAll(implication) => {
+                self.infer.instantiate_binders_existentially(implication)
+            }
+        };
         debug!("consequence = {:?}", consequence);
         debug!("conditions = {:?}", conditions);
 
--- a/src/solve/slg/implementation/resolvent.rs
+++ b/src/solve/slg/implementation/resolvent.rs
@@ -305,10 +305,10 @@ impl<'t> AnswerSubstitutor<'t> {
     /// case.
     fn assert_matching_vars(&mut self, answer_depth: usize, pending_depth: usize) -> Fallible<()> {
         assert!(answer_depth < self.answer_binders);
-        assert!(pending_depth < self.answer_binders);
+        assert!(pending_depth < self.pending_binders);
         assert_eq!(
-            answer_depth - self.answer_binders,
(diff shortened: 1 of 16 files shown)


src/solve/slg/implementation/resolvent.rs
```diff
@@ -300,10 +305,10 @@ impl<'t> AnswerSubstitutor<'t> {
     /// case.
     fn assert_matching_vars(&mut self, answer_depth: usize, pending_depth: usize) -> Fallible<()> {
```
scalexm: I think I fixed a bug just below (overflow when substracting the two `usize` args). I don't know why this did not trigger earlier though.

nikomatsakis: curious

PR #89, opened 2018-03-08
Start documenting the crate.

Phlosioneer:
Also doing small refactorings where it makes sense, and leaving TODO's for larger refactorings or documentation-related features.

Any files I add to this PR are finished. Feel free to comment on them!

I can either document a bunch of files, and then we merge them in one big PR; or I can open a PR for each file; or somewhere in-between. Any preferences?

I'm basically going to be documenting this as I read through and understand the code. Let me know if I get anything wrong, or if there's some subtle points that I've missed. I try my best to be thorough and clear.

PR #108, opened 2018-03-31
Parser documentation

Phlosioneer:
Added lots of documentation to the parser-generator code and the ast structures.

Continuation of #89 

As before, I left some TODO's for parts that I was unsure about, or where I need to reference documentation that hasn't been written yet.

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -13,10 +13,12 @@ impl Span {
     }
 }
 
+/// The top-level container for a parsed .chalk file.
 pub struct Program {
     pub items: Vec<Item>
 }
 
+/// The `Item` enum represents a statement in a .chalk file. 
 pub enum Item {
     StructDefn(StructDefn),
     TraitDefn(TraitDefn),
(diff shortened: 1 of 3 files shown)


chalk-parse/src/ast.rs
```diff
@@ -1,70 +1,140 @@
...
     pub assoc_ty_defns: Vec<AssocTyDefn>,
+
+    /// Any modifiers on this trait.
     pub flags: TraitFlags,
 }
 
+/// This structure contains any flags attached to a rust trait.
 pub struct TraitFlags {
+
+    /// This is true if the trait is `#[auto]`.
```
nikomatsakis: It'd be good to explain a bit what these mean. Maybe something like "Auto traits are automatically implemented for a type based on its contents (think `Send`)."

chalk-parse/src/ast.rs
```diff
@@ -1,70 +1,140 @@
...
     pub auto: bool,
+
+    /// This is true if the trait is `#[marker]`.
```
nikomatsakis: "Marker traits permit overlapping implementations (see https://github.com/rust-lang/rust/issues/29864)"

chalk-parse/src/ast.rs
```diff
@@ -1,70 +1,140 @@
...
     pub marker: bool,
+
+    /// This is true if the trait is `extern`.
```
nikomatsakis: "That means it's defined outside the 'current crate', which can be significant to coherence. See <https://rust-lang.github.io/rfcs/1023-rebalancing-coherence.html>."

chalk-parse/src/ast.rs
```diff
@@ -1,70 +1,140 @@
...
+/// }
+/// ```
+/// 
+/// The `T` in `impl<T>` is a generic type parameter.
 pub enum ParameterKind {
     Ty(Identifier),
     Lifetime(Identifier),
 }
 
+/// A concrete type parameter.
```
nikomatsakis: This is really the *value* that can be given for a generic parameter. We should probably rename it.

chalk-parse/src/ast.rs
```diff
@@ -104,68 +176,123 @@ impl Kinded for Parameter {
...
 
+/// This is the same as a projection, except without an explicit `as`. For
+/// example, `Bar::Item`.
 pub struct UnselectedProjectionTy {
     pub name: Identifier,
     pub args: Vec<Parameter>,
 }
 
+/// Represents a Trait applied to a type, separated by some string. The type that
+/// is "recieving" the trait is the first element of the args array. For example,
```
nikomatsakis: Nit: "receiving"

chalk-parse/src/ast.rs
```diff
@@ -175,30 +302,60 @@ impl PolarizedTraitRef {
...
+/// A valid Rust identifier, along with a `Span` specifying line numbers.
 #[derive(Copy, Clone, Debug)]
 pub struct Identifier {
     pub str: InternedString,
     pub span: Span,
 }
 
+/// A where clause; either a normal Rust trait bound, or a Chalk-specific clause.
+// TODO: Add a "See" that points to a list of Chalk's where clauses, and a
+// description of each.
```
nikomatsakis: Maybe link here?

https://rust-lang-nursery.github.io/rustc-guide/traits-lowering-rules.html#lowering-where-clauses

chalk-parse/src/ast.rs
```diff
@@ -175,30 +302,60 @@ impl PolarizedTraitRef {
...
+    /// Example: `<Foo<T> as Bar>::Item == T`
     ProjectionEq { projection: ProjectionTy, ty: Ty },
+
+    // TODO: What does "well formed" mean?
     TyWellFormed { ty: Ty },
+
+    // TODO: What does "well formed" mean?
     TraitRefWellFormed { trait_ref: TraitRef },
+
+    // TODO: What does "from env" mean?
```
nikomatsakis: hmm, Well Formed and FromEnv ought to be explained here

https://rust-lang-nursery.github.io/rustc-guide/traits-implied-bounds.html

but seems like I didn't write it yet =)

nikomatsakis: there is some coverage here: 

https://rust-lang-nursery.github.io/rustc-guide/traits-lowering-rules.html#implied-bounds

chalk-parse/src/ast.rs
varkor: Probably a good idea to go with `Arg` or similar, following the drift in naming convention in `rustc`.

chalk-parse/src/ast.rs
Phlosioneer: Alright. I honestly don't know how they work myself, I'll add that.

chalk-parse/src/ast.rs
Phlosioneer: Yeah. I struggled with these two at first, and tried my best to give it a name. I second renaming it `Arg` instead of `Parameter`.

chalk-parse/src/ast.rs
Phlosioneer: Woah! I've never seen that guide before!

nikomatsakis:
ping @Phlosioneer =) looks like I had some minor suggestions? Not sure. At minimum, this PR needs a rebase...

nikomatsakis:
So... this Pr is probably good, but it's been a long time and I think I'm just going to close it. Feel free to re-open.

PR #109, opened 2018-03-31
Fix executable permissions

Phlosioneer:
For some reason, src/lib.rs and src/lower/test.rs were marked with
executable permissions. Thit PR removes those permissions.


nikomatsakis:
Weird. Thanks =)

Issue #105, opened 2018-03-27
A deref obligation


leodasvacas:
Deref is not well integrated into typeck, all derefs must be eagerly checked causing problems even in a simpe program such as:
```rust
use std::sync::Arc;

fn main() {
    let mut a = Default::default();
    // Works if the lines are swapped.
    let b = *a;
    a = Arc::new(0);
}
```
The type of `a` will not be inferred because derefs error if they find a type variable. There are probably good reasons of why this difficult to support, but hopefully chalk has a better chance at supporting derefs as obligations. I'll start by implementing `Deref` as domain goal.

nikomatsakis:
Closing this for now. I think handling coercions is probably pretty squarely future work.

PR #112, opened 2018-04-16
Raw pointer deref.

leodasvacas:
Introduce `*mut` and `*const` to the parser which for now are semantically equivalent. Add fact `Derefs(T, U) :- T == *const/mut U`. Part of #105.


nikomatsakis:
I think this is sort of hopelessly outdated at this point. Sorry @leodasvacas =)

Issue #116, opened 2018-05-01
Implement GATs


scalexm:
## Step 1
The grammar needs to be extended to support bounds and (quantified) where clauses on associated types like in:
```rust
trait Foo {
    type Item<'a, T>: Clone where Self: Sized;
}
```

The grammar is located [parser.lalrpop](https://github.com/rust-lang-nursery/chalk/blob/master/chalk-parse/src/parser.lalrpop), and the AST in [ast.rs](https://github.com/rust-lang-nursery/chalk/blob/master/chalk-parse/src/ast.rs).

When this is done, this item:
https://github.com/rust-lang-nursery/chalk/blob/eeb21829c1e847921261ad820b5b2ec35b649c76/chalk-parse/src/parser.lalrpop#L204-L207
won't be needed anymore because we will be using `QuantifiedWhereClauses` everywhere.

We would also remove where clauses on associated type values because they are useless:
https://github.com/rust-lang-nursery/chalk/blob/eeb21829c1e847921261ad820b5b2ec35b649c76/chalk-parse/src/parser.lalrpop#L104-L111

## Step 2
Currently, a where clause of the form `where T: Foo<Item = U>` translated to a single domain goal `ProjectionEq(<T as Foo>::Item = U)`. We would need this clause to be translated to two distinct domain goals:
* `ProjectionEq(<T as Foo>::Item = U)` as before
* `Implemented(T: Foo)`

Currently, we cannot translate one AST where clause to multiple IR domain goals. This means that the following trait:
https://github.com/rust-lang-nursery/chalk/blob/eeb21829c1e847921261ad820b5b2ec35b649c76/src/lower/mod.rs#L437-L439
should be changed into something like:
```rust
trait LowerWhereClause<T> {
    fn lower(&self, env: &Env) -> Result<Vec<T>>;
}
```

## Step 3
Now we should be in a position where we can implement the various rules written in [this comment](https://github.com/rust-lang-nursery/chalk/issues/12#issuecomment-377258347) (niko's comment just below can be ignored, it was basically merged into the main one). This includes:
* program clauses which should be implemented in [lower/mod.rs](https://github.com/rust-lang-nursery/chalk/blob/master/src/lower/mod.rs)
* WF requirements which should be implemented in [lower/wf.rs](https://github.com/rust-lang-nursery/chalk/blob/master/src/lower/wf.rs)

Note: the organization of these two files will change after #114 is fixed, so the preceding text may become obsolete.

## Step 4
[...]

PR #118, opened 2018-05-07
Implement GATs, step 1

tmandry:
This adds parsing rules for bounds and where clauses on associated type definitions for #116.

It also removes where clauses on associated type values.

The way I handled bounds directly in the parser felt a bit messy, open to suggestions. Not sure if using the associated type name directly in the `TraitRef` is going to work downstream.

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -54,7 +54,7 @@ pub struct TraitFlags {
 pub struct AssocTyDefn {
     pub name: Identifier,
     pub parameter_kinds: Vec<ParameterKind>,
-    pub bounds: Vec<TraitBound>,
+    pub bounds: Vec<InlineBound>,
     pub where_clauses: Vec<QuantifiedWhereClause>,
 }
 
@@ -68,20 +68,23 @@ pub enum Parameter {
     Lifetime(Lifetime),
 }
 
+/// An inline bound, e.g. `: Foo<K>` in `impl<K, T: Foo<K>> SomeType<T>`.
+pub enum InlineBound {
+    TraitBound(TraitBound),
+    ProjectionEqBound(ProjectionEqBound),
+}
+
 /// Represents a trait bound on e.g. a type or type parameter.
 /// Does not know anything about what it's binding.
 pub struct TraitBound {
     pub trait_name: Identifier,
-    pub args_no_self: Vec<TraitBoundParameter>,
+    pub args_no_self: Vec<Parameter>,
 }
 
-pub enum TraitBoundParameter {
-    Ty(Ty),
-    Lifetime(Lifetime),
-    ProjectionEq(ProjectionEqBound),
-}
-
+/// Represents a projection equality bound on e.g. a type or type parameter.
+/// Does not know anything about what it's binding.
 pub struct ProjectionEqBound {
+    pub trait_bound: TraitBound,
     pub name: Identifier,
     pub parameters: Vec<Parameter>,
     pub value: Ty,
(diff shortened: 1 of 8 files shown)


chalk-parse/src/ast.rs
```diff
@@ -54,6 +54,8 @@ pub struct TraitFlags {
 pub struct AssocTyDefn {
     pub name: Identifier,
     pub parameter_kinds: Vec<ParameterKind>,
+    pub bound: Option<TraitRef>,
```
scalexm: We would need multiple bounds there (see below)

chalk-parse/src/parser.lalrpop
```diff
@@ -72,9 +72,26 @@ TraitDefn: TraitDefn = {
 };
 
 AssocTyDefn: AssocTyDefn = {
-    "type" <name:Id> <p:Angle<ParameterKind>> ";" => AssocTyDefn {
-        name: name,
-        parameter_kinds: p
+    "type" <name:Id> <p:Angle<ParameterKind>> <b:(":" <Id> <Angle<Parameter>>)?>
+        <w:QuantifiedWhereClauses> ";" =>
+    {
```
scalexm: Ok so currently we do not handle at all the `+` separator which expresses multiple bounds as in `where T: Clone + Sized`. This is not a problem for e.g. where clauses on traits because you can still write `where T: Clone, T: Sized` but for associated types since there is no other way to express these multiple bounds, we need to be able to parse them with the `+` acting as a separator.

chalk-parse/src/ast.rs
scalexm: We would need multiple bounds there (see below), and using `TraitRef` is too rigid as we cannot parse things like `type Foo<T>: Iterator<Item = T>`.

scalexm: I'm thinking this would maybe need some slight refactoring for how we handle where clauses in the grammar. Something which may work would be to split existing where clauses like `TraitRef` into two parts: a left part (which would be `T` in `T: Foo<K>`) and a right part (which would be `Foo<K>`), separated by the `:` symbol. This would prevent code duplication I think because in the case of associated types, you would only ask for the right part and then build the whole `TraitRef` by hand as you did. It would be especially useful when we add lifetime requirements on types as well.

tmandry: Yep, what do you think of something like `TraitBound` and `ProjectionEqBound` that I added to ast? (Currently they're only being used for GATs.)

These can contain methods to convert them to `TraitRef`, `WhereClause::ProjectionEq`, etc. if the proper information is obtained.

chalk-parse/src/parser.lalrpop
tmandry: done

chalk-parse/src/ast.rs
```diff
@@ -177,6 +177,11 @@ impl PolarizedTraitRef {
     }
 }
 
+pub struct TraitBound {
```
nikomatsakis: I think it would be useful to add a comment here explaining what this represents and how it is used.

chalk-parse/src/ast.rs
```diff
@@ -68,6 +68,23 @@ pub enum Parameter {
     Lifetime(Lifetime),
 }
 
+pub struct TraitBound {
```
nikomatsakis: Maybe this should not be called `TraitBound`? I would think something like `TypeParameterBound`

chalk-parse/src/ast.rs
tmandry: done!

PR #119, opened 2018-05-07
Implement GATs, step 2

tmandry:
This refactors LowerWhereClause to permit multiple domain goals from a single AST where clause.

For #116.

For AST Goals and Clauses, I added some assertions that there was only one domain goal, to keep the one-to-many relationship from "leaking" into those traits. I think this is correct, but not too pretty.

### DIFF
--- a/src/ir/lowering/mod.rs
+++ b/src/ir/lowering/mod.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use chalk_parse::ast::*;
 use lalrpop_intern::intern;
 
-use cast::Cast;
+use cast::{Cast, Caster};
 use errors::*;
 use ir::{self, Anonymize, ToParameter};
 use solve::SolverChoice;
@@ -205,7 +205,7 @@ impl LowerProgram for Program {
                     impl_data.insert(item_id, d.lower_impl(&empty_env)?);
                 }
                 Item::Clause(ref clause) => {
-                    custom_clauses.push(clause.lower_clause(&empty_env)?);
+                    custom_clauses.extend(clause.lower_clause(&empty_env)?);
                 }
             }
         }
@@ -412,12 +412,7 @@ trait LowerWhereClauseVec<T> {
 
 impl LowerWhereClauseVec<ir::DomainGoal> for [WhereClause] {
     fn lower(&self, env: &Env) -> Result<Vec<ir::DomainGoal>> {
-        self.iter()
-            .flat_map(|wc| match wc.lower(env) {
-                Ok(v) => v.into_iter().map(Ok).collect(),
-                Err(e) => vec![Err(e)],
-            })
-            .collect()
+        self.iter().flat_map(|wc| wc.lower(env).apply_result()).collect()
     }
 }
 
@@ -519,7 +514,7 @@ impl LowerWhereClause<ir::LeafGoal> for WhereClause {
             | WhereClause::TraitRefFromEnv { .. }
             | WhereClause::Derefs { .. } => {
                 let goals: Vec<ir::DomainGoal> = self.lower(env)?;
-                goals.into_iter().map(|g| g.cast()).collect()
+                goals.into_iter().casted().collect()
             }
             WhereClause::UnifyTys { ref a, ref b } => vec![ir::EqGoal {
                 a: ir::ParameterKind::Ty(a.lower(env)?),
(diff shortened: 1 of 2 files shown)


src/ir/lowering/mod.rs
```diff
@@ -412,26 +412,36 @@ trait LowerWhereClauseVec<T> {
 
 impl LowerWhereClauseVec<ir::DomainGoal> for [WhereClause] {
     fn lower(&self, env: &Env) -> Result<Vec<ir::DomainGoal>> {
-        self.iter().map(|wc| wc.lower(env)).collect()
+        self.iter()
+            .flat_map(|wc| match wc.lower(env) {
+                Ok(v) => v.into_iter().map(Ok).collect(),
+                Err(e) => vec![Err(e)],
```
scalexm: mmh can't it be just `=> Err(e)`? (I know there is convertibility from `Vec<Result<T, E>>` to `Result<Vec<T, E>>` but it feels simpler this way in that case)

src/ir/lowering/mod.rs
```diff
@@ -507,17 +518,17 @@ impl LowerWhereClause<ir::LeafGoal> for WhereClause {
             | WhereClause::TyFromEnv { .. }
             | WhereClause::TraitRefFromEnv { .. }
             | WhereClause::Derefs { .. } => {
-                let g: ir::DomainGoal = self.lower(env)?;
-                g.cast()
+                let goals: Vec<ir::DomainGoal> = self.lower(env)?;
+                goals.into_iter().map(|g| g.cast()).collect()
```
scalexm: You can use the wonderful `Caster` iterator adaptor I wrote:
```rust
goals.into_iter().casted().collect()
```
=)

src/ir/lowering/mod.rs
```diff
@@ -837,7 +848,10 @@ trait LowerClause {
 impl LowerClause for Clause {
     fn lower_clause(&self, env: &Env) -> Result<ir::ProgramClause> {
         let implication = env.in_binders(self.all_parameters(), |env| {
-            let consequence: ir::DomainGoal = self.consequence.lower(env)?;
+            let consequences: Vec<ir::DomainGoal> = self.consequence.lower(env)?;
+            assert_eq!(1, consequences.len());
+            let consequence = consequences.into_iter().next().unwrap();
```
scalexm: Ah I understand the problem. So this may not be the right way to do it, because at one point you might want to lower an "AST"-rule which would look like:
```
T: Iterator<Item = i32> :- blabla
```
but the problem is that the consequence part (left part) would be translated to: `ProjectionEq(<T as Iterator>::Item = i32), T: Iterator` so your assertion would fail.

Instead, we would need to output a vector of `ProgramClause` (again), and we would generate two clauses in my example:
```
ProjectionEq(<T as Iterator>::Item = i32) :- blabla
T: Iterator :- blabla
```

I think this is the right way, you might want to get @nikomatsakis's confirmation though.

src/ir/lowering/mod.rs
tmandry: Yeah there's a problem with match arms having different types. If I used `v.into_iter().map(Ok)` then I'd have a `Map` object which of course wouldn't work. So the least common denominator is `Vec`. I may be missing something though.

scalexm: Ah right sorry, ok this is fine the way you wrote I guess.

src/ir/lowering/mod.rs
tmandry: Good point. For `Clause` this is pretty easy.

For `Goal` it's.. a bit more involved. Bubbling up multiple leaf goals in `LowerGoal` would complicate the code quite a bit. I _think_ I could handle it with a `Goal::And` at the lowest level, but I'm not certain if that's sound.

src/ir/lowering/mod.rs
tmandry: I added `ApplyResult` to make it more ergonomic at least.

src/ir/lowering/mod.rs
```diff
@@ -968,15 +985,27 @@ impl<'k> LowerGoal<Env<'k>> for Goal {
...
                       .collect();
                 Ok(Box::new(ir::Goal::Implies(where_clauses?, g.lower(env)?)))
             }
             Goal::And(g1, g2) => {
                 Ok(Box::new(ir::Goal::And(g1.lower(env)?, g2.lower(env)?)))
             }
             Goal::Not(g) => Ok(Box::new(ir::Goal::Not(g.lower(env)?))),
-            Goal::Leaf(wc) => Ok(Box::new(ir::Goal::Leaf(wc.lower(env)?))),
+            Goal::Leaf(wc) => {
+                // A where clause can lower to multiple leaf goals; wrap these in Goal::And.
```
tmandry: Not certain about this yet

scalexm: Seems good to me! Also you can use `itertools::Itertools` and write something like:
```rust
leaves.fold1(|goal, leaf| ir::Goal::And(Box::new(goal), Box::new(leaf)))
      .expect("at least one goal");
```
(see for example [here](https://github.com/rust-lang-nursery/chalk/blob/master/src/rules/wf/mod.rs#L265))

scalexm:
@tmandry See my comment about `itertools::Itertools`, otherwise everything seems ok.

src/ir/lowering/mod.rs
nikomatsakis: Meh. This "custom clause" stuff is mostly useful for writing unit tests of the SLG engine etc. I think it's fine to assert things here, though it'd be even nicer to move it to the grammar, so that a `Clause` always has a limited set of "domain goals" as its "consequence".

src/ir/lowering/mod.rs
nikomatsakis: Bah, itertools should have called this method `reduce`. =)

PR #122, opened 2018-05-09
Even fewer engine traits 

leodasvacas:
Continuation of #120


leodasvacas:
The `Ord` bound was factored away from `CanonicalConstrainedSubst`. This means that we can no longer dedup by ordering, so I went for a larger refactoring, making `DelayedLiteralSets` be a Vec of HashSets rather than a HashSet of Vecs. The uniqueness invariant is maintained by making the Vec private, insertion is done through the `insert_if_minimal` method.

Changes to tests are just cosmetic or ordering related, there should be no functional change.

PR #129, opened 2018-05-14
Split off ContextOps from Context

leodasvacas:
Remove it as a super trait, leaving ContextOps as a completely separate trait.

The motivation is that in rustc the `ChalkContext` which impls `ContextOps` carries an actual `TyCtx<'a, 'tcx, 'gcx>` and needs the extra `'a` lifetime parameter. However types that take just `Context` need at most `'tcx` and `'gcx`, so to prevent the extra lifetime from leaking everywhere `ContextOps` is no longer a super trait of `Context`.


nikomatsakis:
Thanks @leodasvacas !

PR #131, opened 2018-05-21
One big happy family

nikomatsakis:
Refactor the context traits to allow rustc to implement them. The key change is that all the types go into one `Context` trait which offers nothing but types (well, and a few very simple helper methods). This means that, on the rustc side, we can implement `Context` with a type like

```rust
struct ChalkContext<'tcx> { PhantomData<&'tcx ()> }
```

This in turn means that types like `ExClause<ChalkContext<'tcx>>` are effectively the same (contravariant in the lifetime `'tcx`, etc) as other rustc types like `Ty<'tcx>` and so forth. This unlocks the ability to implement the `Lift` and other traits we need. 

Plus, I think this setup is ultimately simpler on the chalk side, since we don't have annoying things like `ExClause<C, I>`. The main downside is that we need some lift and conversion methods here and there.

r? @scalexm or @leodasvacas

### DIFF
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -12,7 +12,7 @@ crate mod prelude;
 ///
 /// At any given time, the SLG solver may have more than one context
 /// active. First, there is always the *global* context, but when we
-/// are in the midst of pursuing some particular strange, we will
+/// are in the midst of pursuing some particular strand, we will
 /// instantiate a second context just for that work, via the
 /// `instantiate_ucanonical_goal` and `instantiate_ex_clause` methods.
 ///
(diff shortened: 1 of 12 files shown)


chalk-engine/src/context.rs
```diff
@@ -1,14 +1,30 @@
...
-/// The "context" in which the SLG solver operates.
-// FIXME(leodasvacas): Clone and Debug bounds are just for easy derive,
-//                     they are not actually necessary.
+/// The "context" in which the SLG solver operates. It defines all the
+/// types that the SLG solver may need to refer to, as well as a few
+/// very simple interconversion methods.
+///
+/// At any given time, the SLG solver may have more than one context
+/// active. First, there is always the *global* context, but when we
+/// are in the midst of pursuing some particular strange, we will
```
lqd: tiny typo: "strange" => "strand"

PR #134, opened 2018-05-21
Finish implementing GATs

### LINKED ISSUE #116: Implement GATs

tmandry:
Implement and test logic around Generic Associated Types.

This required shuffling around the logic rules we used. This PR implements the set of rules described in https://github.com/rust-lang-nursery/chalk/issues/12#issuecomment-377258347.

Closes #116.

### DIFF
--- a/src/rules/wf.rs
+++ b/src/rules/wf.rs
@@ -256,17 +256,52 @@ impl WfSolver {
         // ```
         // we would issue the following subgoal: `forall<'a> { WellFormed(Box<&'a T>) }`.
         let compute_assoc_ty_goal = |assoc_ty: &AssociatedTyValue| {
+            let assoc_ty_datum = &self.env.associated_ty_data[&assoc_ty.associated_ty_id];
+            let bounds = &assoc_ty_datum.bounds;
+
+            let trait_datum = &self.env.trait_data[&assoc_ty_datum.trait_id];
+
             let mut input_types = Vec::new();
             assoc_ty.value.value.ty.fold(&mut input_types);
 
-            if input_types.is_empty() {
-                return None;
-            }
+            let goals = input_types.into_iter()
+                                   .map(|ty| DomainGoal::WellFormedTy(ty).cast());
+                                   //.chain(bounds.iter()
+                                   //             .flat_map(|b| b.clone()
+                                   //                            .lower_with_self(assoc_ty.value.value.ty.clone()))
+                                   //             .map(|g| g.into_well_formed_goal().cast()));
+            let goal = goals.fold1(|goal, leaf| Goal::And(Box::new(goal), Box::new(leaf)));
+                            //.expect("at least one goal");
+            let goal = goal //Goal::Implies(hypotheses, Box::new(goal))
+                .map(|goal| goal.quantify(QuantifierKind::ForAll, assoc_ty.value.binders.clone()));//binders);
 
-            let goals = input_types.into_iter().map(|ty| DomainGoal::WellFormedTy(ty).cast());
-            let goal = goals.fold1(|goal, leaf| Goal::And(Box::new(goal), Box::new(leaf)))
-                            .expect("at least one goal");
-            Some(goal.quantify(QuantifierKind::ForAll, assoc_ty.value.binders.clone()))
+            // FIXME this is wrong (and test)!
+            let mut bound_binders = assoc_ty.value.binders.clone();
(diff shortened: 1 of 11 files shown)


tmandry:
Ready for review! Various disclaimers..

This doesn't yet support higher-ranked inline bounds.

The following clause is not tested:
```
FromEnv(Self: Foo) :- FromEnv((Foo::Item<'a,T>)<Self>).
```
I'm not sure it's possible to test this in the current grammar.

I just remembered I want to clean up some repetition in `lowering.rs` later.

src/rules/wf.rs
```diff
@@ -256,16 +240,50 @@ impl WfSolver {
...
+                      .flat_map(|b| b.lower_with_self(assoc_ty.value.value.ty.clone()))
+                      .map(|g| g.into_well_formed_goal());
+            
+            let goals = wf_goals.chain(bound_goals).casted();
+            let goal = match goals.fold1(|goal, leaf| Goal::And(Box::new(goal), Box::new(leaf))) {
+                Some(goal) => goal,
+                None => return None,
+            };
+
+            let hypotheses =
```
scalexm: and saying that we do the same for where clauses here, regarding the comment above

src/rules/wf/test.rs
```diff
@@ -217,7 +217,7 @@ fn projection_type_in_header() {
 
             // Projection types in an impl header are not assumed to be well-formed,
             // an explicit where clause is needed (see below).
```
scalexm: The comment is obsolete and can be removed.

src/solve/test.rs
```diff
@@ -567,6 +760,46 @@ fn normalize_implied_bound() {
     }
 }
 
+#[test]
+fn normalize_implied_bound_gat() {
```
scalexm: Pre-existing: could this be renamed to just "gat_implied_bounds" or something like this?

src/rules/wf/test.rs
tmandry: I think it's still mostly correct, but I changed the wording some.

PR #136, opened 2018-05-22
Adding IsLocal DomainGoal and Generating IsLocal clause for all non-external types

sunjay:
This PR is relatively small but does several things that all needed to happen at once in order for tests to continue to pass:

1. Added the `IsLocalTy` and `IsLocalTraitRef` domain goals
2. Added parsing for `IsLocal` (just the `Ty` version for now)
   This means that parsing something like `forall<> { IsLocal(Foo) }` works, though this is not tested anywhere currently (other than by me manually) because we do not have parse tests
3. Automatically generates `forall<T> { IsLocal(Foo<T>) }` for `struct Foo<T: Eq>` and generates nothing extra for `extern struct Bar<T: Eq>` because it is marked `extern`

### DIFF
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -508,7 +508,7 @@ pub enum DomainGoal {
 
     /// A predicate which is true is some trait ref is well-formed.
     /// For example, given the following trait definitions:
-    /// 
+    ///
     /// ```notrust
     /// trait Clone { ... }
     /// trait Copy where Self: Clone { ... }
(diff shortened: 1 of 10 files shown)


src/ir.rs
```diff
@@ -567,7 +567,10 @@ pub enum DomainGoal {
     /// Derefs(T, U) :- Implemented(T: Deref<Target = U>)
     /// ```
     /// In Rust there are also raw pointers which can be deref'd but do not implement Deref.
-    Derefs(Derefs)
+    Derefs(Derefs),
+
+    IsLocalTy(Ty),
```
nikomatsakis: Comment here would be nice. At least something minimal like:

---

True if a type is considered to have been "defined" by the current crate. This is true for a `struct Foo { }` but false for a `extern struct Foo { }`. However, for fundamental types like `Box<T>`, it is true if `T` is local.

PR #138, opened 2018-05-22
Added #[fundamental] for structs

sunjay:
This adds the ability to have `#[fundamental]` on structs in chalk. I recommend looking at the test that I added for complete examples of the behaviour that this enables. Some parts of that test are a bit redundant because of how `forall` works, but I did it like that so I could check every case and so that we know that each of the cases I covered work as we would expect them to.

Usually, if you have a type `extern struct External<T>`, you want `IsLocal(External<T>)` to be false regardless of the type of `T`. There are certain "fundamental" types however which benefit from a slightly modified rule. Take `Box<T>` for example. Given a crate that defines `MyInternalType`, we want that crate to be able to implement traits on `Box<MyInternalType>`. To allow this, we make `Box<T>` a "fundamental" type with the `#[fundamental]` attribute. This makes it so that `IsLocal(Box<T>)` can be true if and only if `IsLocal(T)` is true. For the crate mentioned previously, `IsLocal(MyInternalType)` would be true and that would allow them to define a trait for `Box<MyInternalType>` because `IsLocal(Box<MyInternalType>)` would also become true thanks to `Box<T>` being fundamental.

`#[fundamental]` **only** changes the behaviour of types that are marked `extern`. If a type is not marked `extern`, then it should always satisfy `IsLocal` regardless of anything else. An example where this comes up is for `Box<T>` when we are compiling `std`. This allows us to implement any trait for any `Box<T>` regardless of whether `T` is local or extern. This is okay because in the context of compiling `std`, `Box<T>` is local. This is the same behaviour you would expect with any non-fundamental type too.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -320,7 +320,7 @@ impl ir::StructDatum {
             // one type parameter, nor do we know what the behaviour for that should be. Thus, we
             // are asserting here that there is only a single type parameter until the day when
             // someone makes a decision about how that should behave.
-            assert_eq!(self.binders.len_type_parameters(), 1,
+            assert_eq!(self.binders.value.self_ty.len_type_parameters(), 1,
                 "Only fundamental types with a single parameter are supported");
 
             let local_fundamental = self.binders.map_ref(|bound_datum| ir::ProgramClauseImplication {
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -276,6 +276,15 @@ impl ir::StructDatum {
         // If the type Foo is not marked `extern`, we also generate:
         //
         //    forall<T> { IsLocalTy(Foo<T>) }
+        //
+        // Given an `extern` type that is also fundamental:
+        //
+        //    #[fundamental]
+        //    extern struct Box<T> {}
+        //
+        // We generate the following clause:
+        //
+        //    forall<T> { IsLocalTy(Box<T>) :- IsLocalTy(T) }
 
         let wf = self.binders.map_ref(|bound_datum| {
             ir::ProgramClauseImplication {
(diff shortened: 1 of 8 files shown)


src/rules.rs
```diff
@@ -303,6 +312,30 @@ impl ir::StructDatum {
 
             clauses.push(is_local);
         }
+        // If a type is `extern`, but is also `#[fundamental]`, it satisfies IsLocalTy
+        // if and only if its parameters satisfy IsLocalTy
+        else if self.binders.value.flags.fundamental {
```
nikomatsakis: Nit: formatting wise, i'd prefer the comment to be inside the `else if` and to have `} else if .. {` on one line. This is because, when I am reading code, if the `else` is not on the same line as the closing block, I tend to assume there isn't one (I've noticed this pattern before...)

src/solve/test.rs
```diff
@@ -2121,3 +2121,110 @@ fn local_and_external_types() {
...
+
+        goal { IsLocal(Ref<Box<Internal>>) } yields { "Unique" }
+    }
+
+    // If a type is not extern, it is always local regardless of its parameters or #[fundamental]
+    test! {
+        program {
+            // if we were compiling std, Box would never be extern
+            #[fundamental]
+            struct Box<T> { }
```
nikomatsakis: This is an interesting case! Comparing [to the `ty_is_local` fn in rustc](https://github.com/rust-lang/rust/blob/9a4e5df12685e58ad42ce0152cdbe266373c6bf3/src/librustc/traits/coherence.rs#L415-L418), I think this is the same behavior (and it makes sense). In particular, in rustc, [a `Box` is an `AdtDef` and it will have a local `DefId`](https://github.com/rust-lang/rust/blob/9a4e5df12685e58ad42ce0152cdbe266373c6bf3/src/librustc/traits/coherence.rs#L470).

src/solve/test.rs
```diff
@@ -2121,3 +2121,110 @@ fn local_and_external_types() {
...
+        program {
+            #[fundamental]
+            extern struct Box<T> { }
+
+            extern struct External { }
+            struct Internal { }
+        }
+
+        // With fundamental, Box can be local for certain types
+        goal { forall<T> { not { IsLocal(Box<T>) } } } yields { "No possible solution" }
```
nikomatsakis: Just for kicks, let's try to prove:

```
forall<T> { IsLocal(Box<T>) }
```

and

```
not { forall<T> { IsLocal(Box<T>) } }
```

as well. I would expect all of those to fail.

PR #140, opened 2018-05-24
Rename IsLocal to ImplAllowed

sunjay:
After some discussion on Discord, it became clear that the name `IsLocal` is a little confusing and misleading. The interpretation that @nikomatsakis intended for `IsLocal(T: Trait)` was that a "trait ref being local" basically says "this crate is allowed to write that impl" under the orphan rules. This is different (but related) to the description we had before in the documentation for `IsLocal`:

> True if a type is considered to have been "defined" by the current crate.

If all we're looking at is whether a type or trait is "defined" in the current crate, then having `IsLocal(T: Trait)` doesn't really make sense. You would probably want something like `IsLocal(Trait)` instead.

The goal of this work is to model the orphan rules in chalk as part of coherence. That means that the information we're really looking for is actually whether implementing traits for a given type is allowed based on those rules.

Instead of `IsLocal(T)` and `IsLocal(T: Trait)`, this PR renames that domain goal to be `ImplAllowed(T)` and `ImplAllowed(T: Trait)` to signify when the current crate is allowed to implement traits for a type.

* We use `ImplAllowed(T)` for each non-extern type `T` to signify that implementing any trait is fine for that type.
* We use `forall<T> { ImplAllowed(T: Trait) }` when a trait is local to signify that implementing that trait for any type is fine under the orphan rules. If a trait is extern, we have a slightly different rule that says `forall<T> { ImplAllowed(T: Trait) :- ImplAllowed(T) }` to allow you to implement external traits for your internal types.

I think this new name will help clear up a lot and generally make this concept easier to understand.

*Note:* The `IsLocal(T: Trait)` syntax and rules are **not** implemented yet but you should expect a follow-up PR very soon to add that.

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -218,7 +218,7 @@ pub enum WhereClause {
     UnifyLifetimes { a: Lifetime, b: Lifetime },
     TraitInScope { trait_name: Identifier },
     Derefs { source: Ty, target: Ty },
-    TyIsLocal { ty: Ty },
+    IsLocal { ty: Ty },
 }
 
 pub struct QuantifiedWhereClause {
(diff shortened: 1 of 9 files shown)


sunjay:
@nikomatsakis Renamed `ImplAllowed` to `LocalImplAllowed` per our discussion.

sunjay:
@nikomatsakis Per our conversation, we now have two domain goals: `IsLocal(Ty)` and `LocalImplAllowed(T: Trait)`.

PR #141, opened 2018-05-26
Implementing LocalImplAllowed(T: Trait) for local and external traits

sunjay:
This currently contains a bunch of the changes from #140. Once that is merged, the extra changes should disappear automatically. Even with those extra changes, this isn't too difficult to review.

This PR automatically adds the appropriate program clauses for local and external traits. Those program clauses use a new domain goal `LocalImplAllowed(T: Trait)`. That domain goal will be used to implement the coherence + orphan rules.

### DIFF
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -452,17 +452,23 @@ impl TraitDatum {
         let mut clauses = vec![wf];
 
         if !self.binders.value.flags.external {
-            let impl_allowed = self.binders.map_ref(|bound_datum| ProgramClauseImplication {
-                consequence: DomainGoal::LocalImplAllowed(bound_datum.self_ty.clone().cast()),
-                conditions: Vec::new(),
-            }).cast();
+            let impl_allowed = self.binders.map_ref(|bound_datum|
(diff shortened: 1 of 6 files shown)


src/rules.rs
```diff
@@ -437,6 +450,29 @@ impl TraitDatum {
...
+                ProgramClauseImplication {
+                    consequence: DomainGoal::LocalImplAllowed(bound_datum.trait_ref.clone()),
+                    conditions: vec![
+                        DomainGoal::IsLocal(bound_datum.trait_ref.parameters[0].assert_ty_ref().clone()).cast(),
+                    ],
+                }
+            ).cast();
+
+            clauses.push(impl_maybe_allowed);
+        }
```
leodasvacas: nit: These `if` branches have some code duplication, you could DRY it up by doing:
```rust
let mut conditions = Vec::new();
if !self.binders.value.flags.external {
    conditions.push(DomainGoal::IsLocal(....).cast());
}
// ...make ProgramClause and push it
```

sunjay:
Superseded by #150

Issue #144, opened 2018-06-03
Weird bug around projections


scalexm:
Given the following traits:
```rust
trait Bar { }
trait Foo { type Item<T>: Bar; }
```
and the following goal:
```
forall<T, U> {
    if (T: Foo) {
        <T as Foo>::Item<U>: Bar
    }
}
```
we give `No solution` whereas this should be `Unique`. Note that if we remove the generic parameter from `Item`, this works fine.

Basically at some point, the projection type `<T as Foo>::Item<U>` fails to unify with the skolemized type, as we can see in the debug log:
```
| resolvent_clause(
:     goal=ProjectionEq(<!1 as Foo>::Item<!2> = (Foo::Item)<?0, !1>),
:     clause=ForAll(for<type, type> ProgramClauseImplication { consequence: ProjectionEq(<?1 as Foo>::Item<?0> = (Foo::Item)<?0, ?1>), conditions: [] })) {
: | instantiate(arg=ProgramClauseImplication { consequence: ProjectionEq(<?1 as Foo>::Item<?0> = (Foo::Item)<?0, ?1>), conditions: [] })
: | new_variable: var=?1 ui=U2
: | new_variable: var=?2 ui=U2
: | instantiate: vars=[?1, ?2]
: | consequence = ProjectionEq(<?2 as Foo>::Item<?1> = (Foo::Item)<?1, ?2>)
: | conditions = []
: | unify(a=ProjectionEq(<!1 as Foo>::Item<!2> = (Foo::Item)<?0, !1>),
: :       b=ProjectionEq(<?2 as Foo>::Item<?1> = (Foo::Item)<?1, ?2>)) {
: : | unify_ty_ty(a=!2,
: : :             b=?1) {
: : : | unify_var_ty(var=?1, ty=!2)
: : : | unify_var_ty: var ?1 set to !2
: : | }
: : | unify_ty_ty(a=!1,
: : :             b=?2) {
: : : | unify_var_ty(var=?2, ty=!1)
: : : | unify_var_ty: var ?2 set to !1
: : | }
: : | unify_ty_ty(a=(Foo::Item)<?0, !1>,
: : :             b=(Foo::Item)<?1, ?2>) {
: : : | unify_ty_ty(a=?0,
: : : :             b=!2) {
: : : : | unify_var_ty(var=?0, ty=!2)
: : : | }
: : | }
: | }
| }
```

Note that the following answers `Unique` as expected:
```
forall<T, U, V> {
[...]

PR #145, opened 2018-06-03
Add support for quantified bounds on GATs

### LINKED ISSUE #116: Implement GATs

scalexm:
Closes #116 modulo #144.

### DIFF
--- a/src/rules/wf/test.rs
+++ b/src/rules/wf/test.rs
@@ -519,3 +519,45 @@ fn higher_ranked_cyclic_requirements() {
         }
     }
 }
+
+#[test]
+fn higher_ranked_inline_bound_on_gat() {
+    lowering_success! {
+        program {
+            trait Fn<T> { }
+            struct Ref<'a, T> { }
+            struct i32 {}
+
+            struct fn<T> { }
+
+            impl<'a, T> Fn<Ref<'a, T>> for for<'b> fn<Ref<'b, T>> { }
+
+            trait Bar {
+                type Item<T>: forall<'a> Fn<Ref<'a, T>>;
+            }
+
+            impl Bar for i32 {
+                type Item<T> = for<'a> fn<Ref<'a, T>>;
+            }
+        }
+    }
+
+    lowering_error! {
+        program {
+            trait Fn<T, U> { }
+            struct i32 {}
+
+            struct fn<T, U> { }
+
+            trait Bar {
+                type Item<T>: forall<U> Fn<T, U>;
+            }
+
+            impl Bar for i32 {
+                type Item<T> = fn<T, i32>;
+            }
+        } error_msg {
+            "trait impl for \"Bar\" does not meet well-formedness requirements"
+        }
+    }
+}
--- a/src/rules/wf/test.rs
+++ b/src/rules/wf/test.rs
@@ -549,6 +549,8 @@ fn higher_ranked_inline_bound_on_gat() {
 
             struct fn<T, U> { }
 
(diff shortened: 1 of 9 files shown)


src/rules/wf/test.rs
```diff
@@ -519,3 +519,45 @@ fn higher_ranked_cyclic_requirements() {
...
+
+    lowering_error! {
+        program {
+            trait Fn<T, U> { }
+            struct i32 {}
+
+            struct fn<T, U> { }
+
+            trait Bar {
+                type Item<T>: forall<U> Fn<T, U>;
```
nikomatsakis: Can we get a test where the bound *would've* worked, but for the higher-ranked nature of it? I am thinking of something like:

```
struct foo;
impl Fn<i32, i32> for foo { }

impl Bar for i32 {
  type Item<T> = foo;
}
```

scalexm: @nikomatsakis I forgot to add an `impl<T, U> Fn<T, U> for fn<T, U> { }` in that test, I think the (now) corrected test is what you want

PR #149, opened 2018-06-05
Many lifetimes one universe

### LINKED ISSUE #147: Allow a universe to bind many names
In rustc we allow an universal binder to bind more than one name, such as `for<'a, 'b> fn(&'a u32, &'b u32)` were both `'a` and `'b` have a DeBruijn index of `0`. In chalk we also allow you to write that, but when moved into the environment `'a` and `'b` will have distinct universes (see `fn instantiate_binders_universally` and `unify_forall_tys`).

Currently chalk represents universal type in the environment directly by the universe index, so `forall<T> { Goal(T) }` becomes `Goal(Ui)` once the binder is into the environment. We might want to instead represent them with an index into a table in the environment, like we do for existential variables today, this would be one way to allow us to associate the universal type not only with a universe but with a particular name within that universe. Or maybe we could inline it by carrying two indexes `Uij` where `j` is the relative index of the name within the universe.

A motivation would be to u-canonicalize `forall<T, U> { Goal(T, U) }` and  `forall<T, U> { Goal(U, T) }` into the same thing, as it should be easier to put the universals in a canonical order when you know they are at the universe.

Or at least that's how I understood it, ping @nikomatsakis

leodasvacas:
This makes it so that `for<'a, 'b>` or `forall<'a, 'b>` are represented as two lifetimes in the same universe, denoted `!1'0` and `!1'1`. This should not change any semantics.

The end goal is to later make u-canonicalization more powerful by putting parameters bound in the same binder in a canonical order (e.g. order of appearance) so that `forall<T, U> { Goal(T, U) }` and `forall<T, U> { Goal(U, T) }` get u-canonicalized to the same thing, so hopefully that means this change makes sense.

This fixes #147, I think, though there is followup work to be done for types to also use a `UniversalIndex` and improving u-canonicalization.

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -30,7 +30,7 @@ impl Debug for TypeName {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
         match self {
             TypeName::ItemId(id) => write!(fmt, "{:?}", id),
-            TypeName::ForAll(universe) => write!(fmt, "!{}", universe.counter),
+            TypeName::ForAll(universe) => write!(fmt, "!{}_{}", universe.ui.counter, universe.idx),
             TypeName::AssociatedType(assoc_ty) => write!(fmt, "{:?}", assoc_ty),
         }
     }
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -60,7 +60,7 @@ impl Debug for Lifetime {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
         match self {
             Lifetime::Var(depth) => write!(fmt, "'?{}", depth),
-            Lifetime::ForAll(universe) => write!(fmt, "'!{}", universe.counter),
+            Lifetime::ForAll(UniversalIndex { ui, idx }) => write!(fmt, "'!{}'{}", ui.counter, idx),
         }
     }
 }
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -60,7 +60,7 @@ impl Debug for Lifetime {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
         match self {
             Lifetime::Var(depth) => write!(fmt, "'?{}", depth),
-            Lifetime::ForAll(UniversalIndex { ui, idx }) => write!(fmt, "'!{}'{}", ui.counter, idx),
+            Lifetime::ForAll(UniversalIndex { ui, idx }) => write!(fmt, "'!{}_{}", ui.counter, idx),
         }
     }
 }
--- a/chalk-ir/src/fold.rs
+++ b/chalk-ir/src/fold.rs
@@ -135,7 +135,7 @@ pub trait UniversalFolder {
     ///
     /// - `universe` is the universe of the `TypeName::ForAll` that was found
     /// - `binders` is the number of binders in scope
-    fn fold_free_universal_ty(&mut self, universe: UniverseIndex, binders: usize) -> Fallible<Ty>;
+    fn fold_free_universal_ty(&mut self, universe: UniversalIndex, binders: usize) -> Fallible<Ty>;
 
     /// As with `fold_free_universal_ty`, but for lifetimes.
     fn fold_free_universal_lifetime(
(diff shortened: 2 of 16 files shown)


src/solve/infer/unify.rs
```diff
@@ -166,19 +166,22 @@ impl<'t> Unifier<'t> {
...
-        // for<'a...> T == for<'b...> U where 'a != 'b
+        // for<'a...> T == for<'b...> U
         //
         // if:
         //
         // for<'a...> exists<'b...> T == U &&
         // for<'b...> exists<'a...> T == U
+        //
+        // Here we only check for<'a...> exists<'b...> T == U,
+        // can someone smart comment why this is sufficient?
```
leodasvacas: @nikomatsakis I got curious about this check, I feel like there are assumptions related to regions that are not being spelled out, do you remember what's going on here?

leodasvacas:
Thanks for the quick review! I've changed the debug output (new one is really better) and used the same representation for types.

src/solve/infer/unify.rs
nikomatsakis: Huh, it sure doesn't look right to me. We should try to concoct some test. I imagine that this equality would come out as true:

```
for<'a, 'b> fn(&'a u32, &'b u32) = for<'a> fn(&'a u32, &'a u32)
```

It's actually debatable whether this is correct -- as described in https://github.com/rust-lang/rust/issues/33684, I think we actually **should** consider those two types as equal, which may require changes to the unification algorithm implemented here.

But something like *this* is clearly false, and I imagine that some ordering of it may succeed today?

```
for<'a, 'b> fn(&'a u32) -> &'b u32 = for<'a> fn(&'a u32) -> &'a u32
```

leodasvacas: I don't think either of those equalities would hold under this algorithm since they require `'a = 'b`, and didn't understand why we'd want the first one to hold.

nikomatsakis: The first one is tricky. I at first thought it should not hold, but I think that this wrong. The starting point is that if `T <: U` and `U <: T` then `T = U`. And "clearly"  the more general form should be a "subtype" of the less general form (since you can trivially instantiate it that way):

```
for<'a, 'b> fn(&'a u32, &'b u32) <: for<'c> fn(&'c u32, &'c u32)
```

But what about the reverse? Here, because `'a` and `'b` only appear in contravariant positions, we can actually consider `for<'c> fn(&'c, &'c)` to be a subtype of `for<'a, 'b> fn(&'a, &'b)` because, no matter what regions `'a` and `'b` are instantated with when the function is called, `'c` will just be the intersection of them. Put another way, consider that functions with those two signatures can freely call one another:

```rust
fn foo<'a>(x: &'a u32, y: &'a u32) {
    bar(x, y) // OK: 'b and 'c can both be bound to `'a` (or, in fact, the call site)
}

fn bar<'b, 'c>(x: &'b u32, y: &'c u32) {
    foo(x, y) // OK: 'a is inferred to the call site
}
```

([This is the same argument I made before here](https://github.com/rust-lang/rust/issues/32330#issuecomment-202536977))

leodasvacas: Ah thanks for reviving that comment, the unification is too restrictive then, do we even have that differentiation between a late-bound and early-bound universal in chalk? Should I open an issue to track this?

nikomatsakis:
Heh:

> Thanks for the quick review! 

and *then* I dropped it on the floor for 22 days. Sorry about that.

src/solve/infer/unify.rs
nikomatsakis: OK so there are two separate things, right @leodasvacas? The first is the case where we *want* equality to be true and the second is one where we do not... I still think this algorithm is incomplete as impl'd, but it'd be nice to find examples of both where it fails.

> do we even have that differentiation between a late-bound and early-bound universal in chalk?

We do, yes:

https://github.com/rust-lang-nursery/chalk/blob/62038e04b1cd19a77b3dd2a51903435617c1c333/src/ir.rs#L435

At least if I understood your question.

leodasvacas: I'm not sure how to call it, but I mean that difference between an universal lifetime in inputs vs in outputs of a function. Can we construct a type in chalk that represents that difference?

nikomatsakis: Ah, I see. Well, the difference there arises from **variance**, and you're right that Chalk just doesn't model that at all, which is part of the problem here I suppose. We may have to start.

nikomatsakis:
So... I kind of destroying this PR with my latest one ;) but I am planning to rebase the heart of the changes, I think.

nikomatsakis:
Wow, I'm impressed with git here. A rebase actually worked.

PR #150, opened 2018-06-06
Complete Implementation of LocalImplAllowed(T: Trait)

sunjay:
If it makes it easier to review, I can potentially split this PR up into smaller PRs. That being said, I think I have provided enough background in the rest of this description to give you enough context to be able to review all of this with no issues.

(If any part of what I describe isn't totally accurate, I would love your feedback on it. :smile:)

This PR adds `LocalImplAllowed(T: Trait)`. This goal will be used to implement the "orphan check". Basically, if a type `Foo` satisifies `LocalImplAllowed(Foo: Trait)`, it is allowed by the orphan rules to implement `Trait`. Niko and I are working on [another part](https://gist.github.com/nikomatsakis/d82d18fb08453074457c906f8198e3e3) ([context](http://aturon.github.io/blog/2017/04/24/negative-chalk/)) of this which will be incorporated into the overlap check to ensure that impls can't violate the orphan rules.

For a local trait, `LocalImplAllowed` is pretty simple. It allows you to implement that trait for any type since you were the one who defined it.

```rust
forall<P0, P1..Pn> { LocalImplAllowed(P0: LocalTrait<P1..Pn>) }
```

For an external trait, the rules are a bit more complicated. The orphan rules dictate that an implementation `ExternTrait<P1..Pn> for P0` is allowed if all of the following is met:

1. There is some parameter `Pi` that is a local type (according to `IsLocal` and considering fundamental types)
2. Within `Pi`, all type parameters on the impl are covered by this local type -- This rule probably isn't that relevant anymore because tuples aren't fundamental
3. Each prior parameter `Pj` where `j < i` is a **deeply external type** that does not contain any of the parameters `P0..Pn`

`LocalImplAllowed` manages to check for all of this using a new predicate `IsDeeplyExternal`.

## Example

Let's say you had:

```rust
extern trait Foo<T, U, V> where Self: Eq<T> { ... }
```

This would generate the following program clauses:

```rust
forall<Self, T, U, V> {
  LocalImplAllowed(Self: Foo<T, U, V>) :- IsLocal(Self)
}
forall<Self, T, U, V> {
  LocalImplAllowed(Self: Foo<T, U, V>) :-
      IsDeeplyExternal(Self),
      IsLocal(T)
}
forall<Self, T, U, V> {
  LocalImplAllowed(Self: Foo<T, U, V>) :-
      IsDeeplyExternal(Self),
      IsDeeplyExternal(T),
      IsLocal(U)
}
forall<Self, T, U, V> {
  LocalImplAllowed(Self: Foo<T, U, V>) :-
      IsDeeplyExternal(Self),
      IsDeeplyExternal(T),
      IsDeeplyExternal(U),
      IsLocal(V)
}
```
We generate every possible case for each `i` from `0` to `n`. This correctly models the orphan rules as defined above.

`IsDeeplyExternal` is pretty simple. Given a type `S<T1..Tn>`, we generate:

```rust
forall<T1..Tn> { IsDeeplyExternal(S<T1..Tn>) :- IsDeeplyExternal(T1), ..., IsDeeplyExternal(Tn) }
```

This encapsulates the "does not contain any of the parameters `P0..Pn`" part of the rules because type parameters are neither local nor external. If we had `impl<T> ExternalTrait<LocalType> for T`, we would not be able to prove `IsLocal(T)` or `IsDeeplyExternal(T)`.

The reason this PR is rather large is because it contains:

- [x] the implementation of `IsDeeplyExternal`
    - [x] for normal types
        - [x] code documentation
        - [x] tests
    - [x] for fundamental types
        - [x] code documentation
        - [x] tests
- [x] the implementation of `LocalImplAllowed`
    - [x] for normal types
        - [x] code documentation
        - [x] tests
    - [x] for fundamental types
        - [x] code documentation
        - [x] tests

### DIFF
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -709,6 +709,29 @@ pub enum DomainGoal {
     /// like `Box<T>`, it is true if `T` is local.
     IsLocal(Ty),
 
+    /// True if a type is *not* considered to have been "defined" by the current crate. This is
+    /// false for a `struct Foo { }` but true for a `extern struct Foo { }`. However, for
+    /// fundamental types like `Box<T>`, it is true if `T` is external.
+    IsExternal(Ty),
+
+    /// True if a type both external and its type parameters are recursively external
+    ///
+    /// More formally, for each non-fundamental struct S<P0..Pn> that is external:
+    /// forall<P0..Pn> {
+    ///     IsDeeplyExternal(S<P0...Pn>) :-
+    ///         IsDeeplyExternal(P0),
+    ///         ...
+    ///         IsDeeplyExternal(Pn)
+    /// }
+    ///
+    /// For each fundamental struct P<P0>,
+    ///
+    /// forall<P0> { IsDeeplyExternal(S<P0>) :- IsDeeplyExternal(P0) }
+    ///
+    /// Note that any of these types can have lifetimes in their parameters too, but we only
+    /// consider type parameters.
+    IsDeeplyExternal(Ty),
+
     /// Used to dictate when trait impls are allowed in the current (local) crate based on the
     /// orphan rules.
     ///
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -596,6 +596,13 @@ pub struct TraitRef {
     crate parameters: Vec<Parameter>,
 }
 
+impl TraitRef {
+    crate fn type_parameters<'a>(&'a self) -> impl Iterator<Item=Ty> + 'a {
+        // This unwrap() is safe because is_ty ensures that we definitely have a Ty
+        self.parameters.iter().filter(|p| p.is_ty()).map(|p| p.clone().ty().unwrap())
+    }
+}
+
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
 pub enum PolarizedTraitRef {
     Positive(TraitRef),
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -453,13 +453,17 @@ pub struct ApplicationTy {
 }
 
 impl ApplicationTy {
-    crate fn first_type_parameter(&self) -> Option<Ty> {
+    crate fn type_parameters<'a>(&'a self) -> impl Iterator<Item=Ty> + 'a {
         // This unwrap() is safe because is_ty ensures that we definitely have a Ty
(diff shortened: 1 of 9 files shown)


src/ir.rs
```diff
@@ -709,6 +720,29 @@ pub enum DomainGoal {
...
+    ///
+    /// More formally, for each non-fundamental struct S<P0..Pn> that is external:
+    /// forall<P0..Pn> {
+    ///     IsDeeplyExternal(S<P0...Pn>) :-
+    ///         IsDeeplyExternal(P0),
+    ///         ...
+    ///         IsDeeplyExternal(Pn)
+    /// }
+    ///
+    /// For each fundamental struct P<P0>,
```
scalexm: `P<P0>` -> `S<P0>`

src/ir.rs
```diff
@@ -709,6 +720,29 @@ pub enum DomainGoal {
...
+    ///
+    /// forall<P0> { IsDeeplyExternal(S<P0>) :- IsDeeplyExternal(P0) }
```
scalexm: According to these comments, it seems to me that there is no difference between `IsDeeplyExternal` for non-fundamental-extern type and for fundamental-extern types.

src/rules.rs
```diff
@@ -350,18 +357,45 @@ impl StructDatum {
...
+                            bound_datum.self_ty.first_type_parameter().unwrap()
+                        ).cast(),
+                        ],
+                    }).cast());
+                };
+            }
+
+            fundamental_rule!(IsLocal);
+            fundamental_rule!(IsExternal);
+            fundamental_rule!(IsDeeplyExternal);
```
scalexm: Definitely looks like the `IsDeeplyExternal` rules for non-fundamental extern types and fundamental extern types could be merged :) (the macro expansion would result in the same code as in your `else` branch, since fundamental types have exactly 1 type parameter)

src/ir.rs
nikomatsakis: I think the right way to think about it is that structs are either local, extern, or fundamental, but not (e.g.) "extern fundamental".

sunjay:
@nikomatsakis yes. This PR was growing quite big, so I decided to put that in the next one.

PR #151, opened 2018-06-06
Refactoring external and fundamental struct flags into a single StructType enum

sunjay:
The insight was that structs aren't really external or not, they are either local, fundamental, or external. The parsing and AST are left the same, only the IR is changed. This does not affect the behaviour of any of the code at all, only the representation is slightly refactored.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,4 +1,4 @@
-#![allow(bare_trait_object)] // FIXME -- error-chain uses old-style trait objects
+#![allow(bare_trait_objects)] // FIXME -- error-chain uses old-style trait objects
 
 use chalk_parse::{self, ast};
 use ir;
--- a/src/ir/lowering.rs
+++ b/src/ir/lowering.rs
@@ -592,9 +592,12 @@ impl LowerStructDefn for StructDefn {
                 self_ty,
                 fields: fields?,
                 where_clauses,
-                flags: ir::StructFlags {
-                    external: self.flags.external,
-                    fundamental: self.flags.fundamental,
+                struct_type: match (self.flags.external, self.flags.fundamental) {
+                    // A local struct is local regardless of fundamental
+                    (false, _) => ir::StructType::Local,
+                    // An external struct can be fundamental or not
+                    (true, false) => ir::StructType::External,
+                    (true, true) => ir::StructType::Fundamental,
                 },
             })
         })?;
--- a/src/rules.rs
+++ b/src/rules.rs
@@ -337,65 +337,59 @@ impl StructDatum {
         let mut clauses = vec![wf];
 
         // Types that are not marked `extern` satisfy IsLocal(TypeName)
-        if !self.binders.value.flags.external {
-            // `IsLocalTy(Ty)` depends *only* on whether the type is marked extern and nothing else
-            let is_local = self.binders.map_ref(|bound_datum| ProgramClauseImplication {
-                consequence: DomainGoal::IsLocal(bound_datum.self_ty.clone().cast()),
-                conditions: Vec::new(),
-            }).cast();
+        match self.binders.value.struct_type {
+            StructType::External => clauses.extend(vec![
+                self.binders.map_ref(|bound_datum| ProgramClauseImplication {
+                    consequence: DomainGoal::IsExternal(bound_datum.self_ty.clone().cast()),
+                    conditions: Vec::new(),
+                }).cast(),
+                self.binders.map_ref(|bound_datum| ProgramClauseImplication {
+                    consequence: DomainGoal::IsDeeplyExternal(bound_datum.self_ty.clone().cast()),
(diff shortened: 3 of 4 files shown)


src/errors.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(bare_trait_object)] // FIXME -- error-chain uses old-style trait objects
+#![allow(bare_trait_objects)] // FIXME -- error-chain uses old-style trait objects
```
sunjay: This change has nothing to do with the refactor. The warning was just bothering me and I forgot to exclude it from my commit.

src/ir/lowering.rs
```diff
@@ -592,9 +592,12 @@ impl LowerStructDefn for StructDefn {
...
                 where_clauses,
-                flags: ir::StructFlags {
-                    external: self.flags.external,
-                    fundamental: self.flags.fundamental,
+                struct_type: match (self.flags.external, self.flags.fundamental) {
+                    // A local struct is local regardless of fundamental
+                    (false, _) => ir::StructType::Local,
+                    // An external struct can be fundamental or not
+                    (true, false) => ir::StructType::External,
+                    (true, true) => ir::StructType::Fundamental,
```
sunjay: This logic directly corresponds to the way things currently work in the code. Instead of doing it with the `if local { ... } else if fundamental { ... } else { ... }` like before, we do all of that branching here.

src/rules.rs
```diff
@@ -337,65 +337,59 @@ impl StructDatum {
...
-            // If a type is `extern`, but is also `#[fundamental]`, it satisfies IsLocal
-            // if and only if its parameters satisfy IsLocal
-
-            // Fundamental types must always have at least one type parameter for this rule to
-            // make any sense. We currently do not have have any fundamental types with more than
-            // one type parameter, nor do we know what the behaviour for that should be. Thus, we
-            // are asserting here that there is only a single type parameter until the day when
-            // someone makes a decision about how that should behave.
-            assert_eq!(self.binders.value.self_ty.len_type_parameters(), 1,
+        match self.binders.value.struct_type {
```
sunjay: None of this inner code changed, I just moved it all into the branches of the match.

src/ir/lowering.rs
```diff
@@ -592,9 +592,12 @@ impl LowerStructDefn for StructDefn {
                 self_ty,
                 fields: fields?,
                 where_clauses,
-                flags: ir::StructFlags {
-                    external: self.flags.external,
-                    fundamental: self.flags.fundamental,
+                struct_type: match (self.flags.external, self.flags.fundamental) {
+                    // A local struct is local regardless of fundamental
+                    (false, _) => ir::StructType::Local,
```
nikomatsakis: I think this is correct, but it was not immediately obvious to me, I have to admit. But I guess the point is that this program compiles (and clearly should), even though `Clone` is an external trait:

```rust
#![feature(fundamental)]

#[fundamental]
struct Foo<T> { t: T }

impl<T: Clone> Clone for Foo<T> {
    fn clone(&self) -> Self {
        Foo { t: self.t.clone() }
    }
}

fn main() { }
```

sunjay:
> if you have a "local fundamental" type Foo, then some downstream crate can implement something for Foo<DownstreamType>

@nikomatsakis Isn't this exactly what we want? That's the point of fundamental isn't it? If you have a "local fundamental" Foo, it is okay for the downstream crate to implement Foo<DownstreamType> because Foo is fundamental.

I don't think I understand your concern.

nikomatsakis:
My concern is that -- with this enum -- we have no way to represent the difference between a "Local fundamental" thing and an "external fundamental" thing. At first, I thought that both of those were just fundamental, but in fact we see that's not true:

- we need to distinguish local/external fundamental for [this example](https://github.com/rust-lang-nursery/chalk/pull/151#discussion_r193568207)
- but we can't equate local fundamental and local, because [they are different for downstream crates](https://github.com/rust-lang-nursery/chalk/pull/151#pullrequestreview-126566257)

...so really *do* want two booleans, I think.

sunjay:
Okay, so Niko and I had a long conversation about this and here's the conclusion about why we need to continue to distinguish between local and fundamental:

First, it's important to acknowledge that as far as the orphan check goes, the current code is correct (as far as we know currently). So no changes are needed there. The distinction up for discussion is important in another part of coherence where we start to query whether a certain impl would be possible in any possible current/future crate based on the orphan rules.

In that situation, it comes down to the following: (quoted from my conversation with Niko)

> It is important to distinguish Local+!Fundamental from Local+Fundamental because:
>
> - For a Local+!Fundamental type, downstream crates can have fewer impls using that type, which impacts the `compatible { .. }` rules.

The `compatible { .. }` rules being referred to are the rules representing the "current/future crate" concept I was talking about above. We are still in the process of figuring them out.

PR #153, opened 2018-06-08
Performing the Orphan Check on all local impls

sunjay:
This PR adds the orphan check during lowering. This checks that each local impl satisfies the orphan rules via the `LocalImplAllowed` goal.

If you have:

```rust
impl<T> Trait for MyType<T>
```

We generate:

```rust
forall<T> { LocalImplAllowed(MyType<T>: Trait) }
```

If that is not provable, your impl violates the orphan rules.

---

This PR adds:

- [x] The ability to mark an impl `extern` to allow us to differentiate between local and external impls (not actually used in tests yet, see below)
    - [x] parsing => AST
    - [x] lowering => IR
- [x] A new module for the orphan check similar to the modules we have for overlap/disjoint, wf, etc.
    - [x] call to the method this module adds during the lowering process
- [x] Tests for coherence violations (Looks like `LocalImplAllowed` works!)

The tests are largely taken/adapted from rustc's coherence tests. I am not totally satisfied with their coverage, so I will probably go add more tests tomorrow. In particular, I want to add some tests with successful outcomes so that the `extern impl` syntax I added can actually get used. This syntax, while not used just yet, is definitely necessary for lowering real Rust programs.

### DIFF
--- a/src/coherence/test.rs
+++ b/src/coherence/test.rs
@@ -226,3 +226,173 @@ fn overlapping_negative_impls() {
         }
     }
 }
+
+#[test]
+fn orphan_check() {
+    // These tests are largely adapted from the compile-fail coherence-*.rs tests from rustc
+
+    lowering_error! {
+        program {
+            extern trait Foo { }
+            extern struct Bar { }
+
+            impl Foo for Bar { }
+        } error_msg {
+            "impl for trait \"Foo\" violates the orphan rules"
+        }
+    }
+
+    lowering_error! {
+        program {
+            extern trait Foo { }
+
+            impl<T> Foo for T { }
+        } error_msg {
+            "impl for trait \"Foo\" violates the orphan rules"
+        }
+    }
+
+    lowering_error! {
+        program {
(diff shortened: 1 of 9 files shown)


src/coherence/test.rs
```diff
@@ -226,3 +226,173 @@ fn overlapping_negative_impls() {
...
+    lowering_error! {
+        program {
+            #[auto] extern trait Send { }
+            extern trait TheTrait<T> { }
+            extern struct isize { }
+            extern struct usize { }
+
+            struct TheType { }
+
+            impl TheTrait<TheType> for isize { }
```
nikomatsakis: putting these three impls together into one test makes it hard to see which one is getting the error...maybe break them up? if needed we could make a macro that lets you "combine" a prefix with a number of suffixes, for convenience.

nikomatsakis: or else we could change the error to report all the failed impls at once

nikomatsakis: (in this case, I guess we expect only one failure?)

PR #155, opened 2018-06-17
Fix link in README

daboross:
The old page linked to no longer exists. I think this is the best equivalent page that exists now?


nikomatsakis:
Thanks!

PR #157, opened 2018-06-18
Rename IsExternal to IsUpstream and extern keyword to #[upstream]

sunjay:
We are changing the terminology of chalk to use "upstream" instead of "external". This change is particularly useful in the upcoming coherence implementation as it precisely describes the structs and traits we want to be talking about.

This PR does not change `IsDeeplyExternal` as that is going to be removed in #153 anyway once I get around to it.

This is just a renaming, so no functionality is changed whatsoever. Hopefully that will make it very easy to review.


nikomatsakis:
Merge conflicts now =)

sunjay:
@nikomatsakis Resolved. :)

PR #162, opened 2018-07-06
Taking ambiguous into account in disjoint

sunjay:
We only want to consider impls disjoint if we can unambiguously say that they are that way. Previously, we weren't doing that because `is_some()` would allow both `Yes` and `Ambig` as results. This change makes the handling explicit.


sunjay:
Just going to include this as part of #165 since it is pretty related.

PR #164, opened 2018-07-06
More useful debug implementation for ProgramClause and ProgramClauseImplication

sunjay:
This is something I did while I was debugging why some of my code wasn't working. This makes it significantly easier to read program clauses. I am open to any feedback for how to make this better. 

One thing I could do if you think it makes sense is print out the conditions of the program clause implication as a list with `[` and `]` wrapped around it (there is a method for that in `Formatter`).

### DIFF
--- a/src/ir/debug.rs
+++ b/src/ir/debug.rs
@@ -291,6 +291,32 @@ impl<T: Debug> Debug for Binders<T> {
     }
 }
 
+impl Debug for ProgramClause {
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
+        match self {
+            ProgramClause::Implies(pc) => write!(fmt, "{:?}", pc),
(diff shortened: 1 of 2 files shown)


src/ir/debug.rs
```diff
@@ -291,6 +291,32 @@ impl<T: Debug> Debug for Binders<T> {
...
+        write!(fmt, "{:?}", self.consequence)?;
+
+        let conds = self.conditions.len();
+        if conds == 0 {
+            return Ok(());
+        }
+
+        write!(fmt, " :- ")?;
+        for cond in &self.conditions[..conds - 1] {
+            write!(fmt, "{:?}, ", cond)?;
```
nikomatsakis: ugh why is 2018 and formatting a comma-separated list is still a pain

PR #165, opened 2018-07-09
[WIP] Compatible Overlap Check in Chalk

sunjay:
This pull request completes the implementation of coherence in chalk. It adds additional lowering rules which will be used in "compatible mode" to allow the overlap check to reason about *potential* impls that may or may not exist yet somewhere. This is a key part of enforcing that there is up to one applicable impl at any given time. Each rule added during lowering models an impl that we forsee could be added in a semver compatible world.

Before reviewing and merging this PR, please merge the three **tiny** PRs I created for unrelated changes that are also included on this branch:

1. Taking ambiguous into account in disjoint - https://github.com/rust-lang-nursery/chalk/pull/162
2. Add parsing for Compatible and DownstreamType - https://github.com/rust-lang-nursery/chalk/pull/163
3. More useful debug implementation for ProgramClause and ProgramClauseImplication - https://github.com/rust-lang-nursery/chalk/pull/164

Once those are merged, many of the changes in this branch will automatically disappear and you'll be left with just the lowering rules I added.

**Please do not merge this until:**

- [ ] More tests are written
- [ ] The tests pass (need to fix the `not { ... }` problem or figure out a workaround)


sunjay:
@nikomatsakis This now compiles and works as we expected it to work. I basically hand-reverted https://github.com/rust-lang-nursery/chalk/pull/130 per what we discussed. I am very curious whether there are any cases that this won't work for now that we moved away from using `not` in `overlap`.

PR #170, opened 2018-08-14
Replace `::crate` with `crate`

varkor:
Fixes dependency issue in https://github.com/rust-lang/rust/pull/53347.


eddyb:
Huh, Travis CI failed because of `?` in macros, did we revert that or something?

eddyb:
cc @nikomatsakis we need a new version published

nikomatsakis:
done -- v0.7.0

PR #173, opened 2018-09-07
type alias bounds and other miscellany

nikomatsakis:
Add correct type alias bounds

cc @eddyb


nikomatsakis:
I also remove various unstable features — not all of them. Seems better to have fewer though.

nikomatsakis:
I would like to get down to zero in shared code.

PR #176, opened 2018-10-04
Fix an implied bound related to assoc types according to #12

scalexm:
See [this comment](https://github.com/rust-lang-nursery/chalk/issues/12#issuecomment-409497442). I explained why this is bad in the test I just added.

cc @nikomatsakis


nikomatsakis:
Ok, this is making my head hurt =)

I had no doubt it's correct, though, but it's not obvious to me yet.

PR #177, opened 2018-10-04
Move some methods in `context::InferenceTable` and add a `&self` arg

scalexm:
Should be good now.

cc @nikomatsakis

### DIFF
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -113,12 +113,6 @@ pub trait Context: Clone + Debug {
         environment: &Self::Environment,
         goal: Self::Goal,
     ) -> Self::GoalInEnvironment;
-
-    /// Upcast this domain goal into a more general goal.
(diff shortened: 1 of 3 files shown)


chalk-engine/src/context.rs
```diff
@@ -115,10 +115,10 @@ pub trait Context: Clone + Debug {
     ) -> Self::GoalInEnvironment;
 
     /// Upcast this domain goal into a more general goal.
-    fn into_goal(domain_goal: Self::DomainGoal) -> Self::Goal;
+    fn into_goal(&self, domain_goal: Self::DomainGoal) -> Self::Goal;
```
nikomatsakis: Why don't we just move these methods onto the `InferenceTable` trait? They only seem to be used from `simplify.rs`, and it has an `InferenceTable`. That gives us access to everything we need, no?

scalexm: Yes this is simpler indeed, I blindly wanted to somehow preserve the semantics of the `Context` trait :)

PR #178, opened 2018-10-04
distinguish bound vs inference variables

nikomatsakis:
This branch refactors Chalk's IR to separate *free inference variables* from *bound variables*. The goal is to generally clarify what is going on — for example, it removes the need to track binder depth during substitution and unification (I'm not sure if I got all the simplifications here, actually, maybe I missed some).

r? @scalexm

### DIFF
--- a/chalk-ir/src/fold.rs
+++ b/chalk-ir/src/fold.rs
@@ -38,10 +38,10 @@ pub use self::subst::Subst;
 ///
 /// To create a folder type F, one typically does one of two things:
 ///
-/// - Implement `ExistentialFolder` and `IdentityPlaceholderFolder`:
+/// - Implement `FreeVarFolder` and `IdentityPlaceholderFolder`:
 ///   - This ignores universally quantified variables but allows you to
 ///     replace existential variables with new values.
-/// - Implement `ExistentialFolder` and `PlaceholderFolder`:
+/// - Implement `FreeVarFolder` and `PlaceholderFolder`:
 ///   - This allows you to replace either existential or universal
 ///     variables with new types/lifetimes.
 ///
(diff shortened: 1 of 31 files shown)


chalk-ir/src/fold.rs
```diff
@@ -100,67 +106,138 @@ pub trait ExistentialFolder {
...
-        depth: usize,
-        binders: usize,
-    ) -> Fallible<Lifetime>;
+    /// As `fold_free_var_ty`, but for lifetimes.
+    fn fold_free_var_lifetime(&mut self, depth: usize, binders: usize) -> Fallible<Lifetime>;
 }
 
 /// A convenience trait. If you implement this, you get an
-/// implementation of `UniversalFolder` for free that simply ignores
+/// implementation of `PlaceholderFolder` for free that simply ignores
```
scalexm: `PlaceholderFolder` -> `FreeVarFolder` (pre-existing with the old names)

chalk-ir/src/fold.rs
```diff
@@ -100,67 +106,138 @@ pub trait ExistentialFolder {
...
 /// universal values (that is, it replaces them with themselves).
```
scalexm: `universal values` -> `free variables`

chalk-ir/src/fold.rs
```diff
@@ -87,11 +93,11 @@ impl<T: ExistentialFolder + UniversalFolder + DefaultTypeFolder> TypeFolder for
...
-/// variables**; for example, if you folded over `Foo<?T> }`, where `?T`
-/// is an inference variable, then this would let you replace `?T` with
-/// some other type.
-pub trait ExistentialFolder {
+/// The methods for folding **free variables**. These are `BoundVar`
+/// instances where the binder is not something we folded over.  This
+/// is used when you are instanting previously bound things with some
+/// replacement.
+pub trait FreeVarFolder {
     /// Invoked for `Ty::Var` instances that are not bound within the type being folded
```
scalexm: `Ty::Var` -> `Ty:: BoundVar`

chalk-ir/src/fold.rs
```diff
@@ -100,67 +106,138 @@ pub trait ExistentialFolder {
...
+    }
+}
+
+pub trait InferenceFolder {
+    /// Invoked for each occurence of a inference type; these are
+    /// used when we instantiate binders universally. Returns a type
+    /// to use instead, which should be suitably shifted to account
+    /// for `binders`.
+    ///
+    /// - `universe` is the universe of the `TypeName::ForAll` that was found
```
scalexm: Also fix comment here.

PR #195, opened 2018-12-31
Switch to Rust 2018

fabric-and-ink:
Not sure if this is worthwhile but I switched chalk to Rust 2018.


tmandry:
Looks great, thank you!

PR #196, opened 2019-01-04
Better context docs

tmandry:
Minor addition really, but couldn't think of anything else to add at the moment. Linking to the guide is at least a good place to start.


nikomatsakis:
Thanks @tmandry =)

PR #200, opened 2019-01-08
(very) WIP: trait objects

scalexm:
I've sketched here the rules for well-formedness of trait objects, prototyping what I tried to explain on Zulip.

cc @nikomatsakis

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -158,6 +158,8 @@ pub enum Ty {
 
     /// Inference variable.
     InferenceVar(InferenceVar),
+
+    Dynamic(Vec<TraitBound>),
 }
 
 impl Ty {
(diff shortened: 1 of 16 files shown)


chalk-ir/src/lib.rs
```diff
@@ -411,13 +413,53 @@ impl TraitRef {
...
     Implemented(TraitRef),
     ProjectionEq(ProjectionEq),
 }
 
+/// Represents a trait bound on e.g. a type or type parameter.
+/// Does not know anything about what it's binding.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TraitBound {
+    pub trait_id: ItemId,
+    pub args_no_self: Vec<Parameter>,
```
nikomatsakis: Random thought: we may want to introduce a `Binder` and use that to represent `Self`. I was sort of hoping that we could make a trait object something like

```
exists<Self> { Clause* }
```

chalk-solve/src/solve/slg/implementation/clauses.rs
```diff
@@ -0,0 +1,47 @@
+use chalk_ir::*;
+use chalk_ir::cast::{Cast, Caster};
+
+crate fn program_clauses_for_dynamic_types(goal: &DomainGoal) -> impl Iterator<Item = ProgramClause> {
```
nikomatsakis: Interesting. I was wondering how to handle the "open-ended" set of rules here. 

I was imagining perhaps a slightly more "meta" approach, but it would require deeper changes to the engine. 

e.g., I was imagining something where to solve `dyn (\Self => C): Trait<..>` we kind of convert it to:

```rust
forall<Self> {
  if (C[Self]) { // here C is assumed to be `Self => Clause`
    Self: Trait<..>
  }
}
```

chalk-ir/src/lib.rs
scalexm: I see. That could work, what advantages would you see in using this representation?

chalk-solve/src/solve/slg/implementation/clauses.rs
scalexm: In rustc we already have this approach where we return program clauses only when we have a bit more information about the goal, because of e.g. tuples and fn types, for which the arity cannot be encoded in the logic.

chalk-ir/src/lib.rs
nikomatsakis: I was thinking that there would be less duplication -- i.e., we wouldn't need a `TraitBound` that is distinct from other clauses.

chalk-solve/src/solve/slg/implementation/clauses.rs
nikomatsakis: Yep. I think I probably like this approach better than baking dyn types more deeply into the engine.

nikomatsakis: It feels to me like -- at some point -- we should definitely try to have more code sharing between chalk + rustc though. There is a fair amount of non-trivial logic that seems like it is basically the same between the two projects.

scalexm:
Closing for now. I'll probably open another PR when it's time.

PR #207, opened 2019-03-06
Build on stable

flodiebold:
Since we're thinking about integrating chalk with rust-analyzer, having chalk build on stable would be great :slightly_smiling_face: 

Getting there is mostly straightforward. Specialization was used to provide a nicer Debug implementation for `Parameter`; I chose the implementation that broke less tests ;) That's where the changes in the `slg` tests come from.


chalk-ir/src/debug.rs
```diff
@@ -344,10 +344,10 @@ impl<T: Display> Display for Canonical<T> {
 }
 
 impl<T: Debug, L: Debug> Debug for ParameterKind<T, L> {
-    default fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
         match *self {
-            ParameterKind::Ty(ref n) => write!(fmt, "Ty({:?})", n),
-            ParameterKind::Lifetime(ref n) => write!(fmt, "Lifetime({:?})", n),
```
nikomatsakis: Hmm, this is annoying. Sometimes we use `ParameterKind<T, T>` for the same `T`, in which case it's pretty important to know whether it is `Ty` or `Lifetime`. OTOH yeah the output otherwise is kind of annoyingly verbose.

We could probably just make a copy of this type for the special case of `Parameter`, even if it means copying some methods or something. Not sure how much work that would be, but it might be worth it.

chalk-macros/src/index.rs
```diff
@@ -22,6 +22,11 @@ macro_rules! index_struct {
             $v fn increment(&mut self) {
                 self.value += 1;
             }
+
+            // TODO: Once the Step trait is stabilized (https://github.com/rust-lang/rust/issues/42168), instead implement it and use the Iterator implementation of Range
+            pub fn iterate_range(range: ::std::ops::Range<Self>) -> impl Iterator<Item = $n> {
```
nikomatsakis: Nit: Can we make this a method with an extension trait? would read nicer.

nikomatsakis:
[Started a Zulip topic to discuss](https://github.com/rust-lang-nursery/chalk/pull/207#pullrequestreview-214105279)

chalk-macros/src/index.rs
flodiebold: Hm, I don't think I'd be able to use impl trait in that case?

flodiebold:
Pushed a change that makes `Parameter` a newtype with its own `Debug` implementation.

flodiebold:
Ok, I didn't run `cargo test --all`, will fix that...

chalk-macros/src/index.rs
eddyb: Yeah, not on stable, at least.

PR #208, opened 2019-03-14
Split ItemId into several id types

detrumi:
As part of [Chalk-RLS integration](https://paper.dropbox.com/doc/Chalk-RLS-integration-QCJelXDeaq7GV6jVQln4j), this attempts to split the `ItemId` type into several types like `TypeId` and `TraitId`. I feel that this also improves the code, since code that handles a subset of item kinds becomes more descriptive (which is currently often reflected in variable naming, like in `trait_id: ItemId`)

Similar to how rust-analyzer handles its ids, each id is a newtype wrapper around a single `RawId` type. To keep the solver general over different value types, `ItemId` becomes an enum with variants for all item-id types, with some more specific id enums like `TypeKindId`.

Note: this is an early WIP, I just got it to compile but the tests are still failing, so some types are probably incorrectly assigned.

### DIFF
--- a/chalk-ir/src/fold.rs
+++ b/chalk-ir/src/fold.rs
@@ -482,6 +482,9 @@ macro_rules! copy_fold {
 copy_fold!(Identifier);
 copy_fold!(UniverseIndex);
 copy_fold!(ItemId);
+copy_fold!(StructId);
+copy_fold!(TraitId);
+copy_fold!(TypeId);
 copy_fold!(usize);
 copy_fold!(QuantifierKind);
 copy_fold!(chalk_engine::TableIndex);
--- a/chalk-ir/src/fold.rs
+++ b/chalk-ir/src/fold.rs
@@ -358,7 +358,7 @@ pub fn super_fold_ty(folder: &mut dyn Folder, ty: &Ty, binders: usize) -> Fallib
                     folder.fold_free_placeholder_ty(ui, binders)
                 }
 
-                TypeName::ItemId(_) | TypeName::AssociatedType(_) => {
+                TypeName::TypeKindId(_) | TypeName::AssociatedType(_) => {
                     let parameters = parameters.fold_with(folder, binders)?;
                     Ok(ApplicationTy { name, parameters }.cast())
                 }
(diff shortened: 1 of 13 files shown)


src/rust_ir.rs
```diff
@@ -18,22 +19,22 @@ pub mod lowering;
...
-    crate type_ids: BTreeMap<Identifier, ItemId>,
+    crate type_ids: BTreeMap<Identifier, TypeId>,
 
     /// For each struct/trait:
-    crate type_kinds: BTreeMap<ItemId, TypeKind>,
+    crate type_kinds: BTreeMap<TypeKindId, TypeKind>,
 
     /// For each struct:
-    crate struct_data: BTreeMap<ItemId, StructDatum>,
+    crate struct_data: BTreeMap<StructId, StructDatum>,
```
matklad: Given that per-type ids are now dense, one *can* use a `Vec` or arena here, instead of `BTreeMap`. I don't think we should change this though: my understanding that we are going to refactor out this struct to a set of queries anyway.

chalk-ir/src/fold.rs
```diff
@@ -474,6 +474,9 @@ macro_rules! copy_fold {
 copy_fold!(Identifier);
 copy_fold!(UniverseIndex);
 copy_fold!(ItemId);
+copy_fold!(StructId);
```
flodiebold: note that rust-analyzer distinguishes further between structs and enums, and the various language-defined types (primitives, function pointers etc.). the distinction doesn't matter to Chalk, but it matters a lot in rust-analyzer :) we'll have to think about how to handle this.

nikomatsakis: Interesting.

nikomatsakis:
Any idea why CI is not active here? Is it because the PR is marked as draft?

nikomatsakis:
I marked as ready for review :P

src/rust_ir.rs
nikomatsakis: Yeah, leave it

detrumi:
Whoops, forgot to run `cargo test --all`!

PR #211, opened 2019-04-10
decompose chalk logic for query-ification and re-use

nikomatsakis:
This branch teases apart the chalk repo into more independent pieces and establishes clear boundaries between them, in the form of traits. The goal is to enable the RLS 2.0 work (and eventually rustc) to re-use the lowering code and more of the IR.

I think the branch has reached *almost* the state that I want, except that the names need work, and there needs to be clearer documentation. I'm hoping to produce some of that, but I thought it might be better to discuss the overall setup and naming conventions we want all together.

Key goals:

- the `Program` and `ProgramEnvironment` structs, which currently define the full set of traits/impls/lowered-clauses etc, are moved into the `chalk` crate
    - eventually these should just be the `ChalkDatabase` but that transition is not complete
- all the interesting bits (lowering, coherence, solver) are in other crates and interact with those program/environment structs only through traits -- this means they demand little bits of the information at a time
    - right now they often take multiple `&dyn Foo` pointers for these traits; I'd like to consolidate these too so that they effectively just take a `db` database that is bounded by various traits
- in particular the lowering logic and coherence rules are moved into a `chalk-rules` crate
    - with a bit more refactoring, I think we could make this crate independent of `chalk-ir`, which would probably permit this code to be shared with rustc
- the solver interface is cleaned up and simplified
    - but we're still instantiating a lot of one-off solvers in the WF rules etc, which I'd like to fix


nikomatsakis:
The failure is nightly only, and seems to be because of a change to the behavior of `{:#?}` to add trailing commas...

README.md
```diff
@@ -37,3 +37,4 @@ Here are some blog posts talking about chalk:
 ## Contributing
 
 If you're like to contribute, consider joining the [Traits Working Group](https://github.com/rust-lang/rust/issues/48416). We hang out on the [rust-lang discord](http://discord.gg/rust-lang) in the `#wg-traits` channel (Lang category).
+
```
bjorn3: Why this extra newline?

nikomatsakis: Yeah, I don't know.

Issue #214, opened 2019-04-16
simplified crate reorg


nikomatsakis:
After pondering the comments by @matklad and others in our discussion, I propose the following crate organization. The idea is to merge the chalk-solve and chalk-rules crate, so that outside code only needs to instantiate a `Solver` with a `RustIrDatabase` (sometimes that supplies rust-ir), and everything else is handled internally to the `chalk-solve` crate. This seems simpler to explain and deal with than today's setup, but it has one downside: if we ever wanted to have rustc incorporate the chalk-rules logic directly, we would either have to tease the two crates apart again (which requires some traits in between) or else port rustc to use chalk-ir.

## Overview and end-goal

The chalk-solve crate would expect to be given a `RustIrDatabase`, which would be roughly equivalent to today's [`RustIrSource`]. It would allow you to instantiate a [`Solver`], just as today. Unlike today, though, [the `solve` method][sm] would not take a `&dyn ChalkSolveDatabase` but rather a `&dyn RustIrDatabase`. The implementation of the methods that are in `ChalkSolveDatabase` today would be moved and hardcoded into chalk-solver itself. So basically the idea is "give me some rust source, and I'll give you a trait solver".

[`RustIrSource`]: https://github.com/rust-lang/chalk/blob/eb1ed91c944b8e4f12b36299b3692b357724f596/chalk-rules/src/lib.rs#L21-L64
[`Solver`]: https://github.com/rust-lang/chalk/blob/eb1ed91c944b8e4f12b36299b3692b357724f596/chalk-solve/src/solve.rs#L95-L101
[sm]: https://github.com/rust-lang/chalk/blob/eb1ed91c944b8e4f12b36299b3692b357724f596/chalk-solve/src/solve.rs#L104-L126

Currently, the chalk-solve code doesn't know about the `RustIrDatabase` directly -- it instead has a `ChalkSolveDatabase` trait with some callbacks. These are implemented by the `chalk` crate, which then delegates to the code in chalk-rules. The idea here is to "cut out the middleman" and have chalk-solve directly invoke chalk-rules. (To add extra fun, the chalk-rules code sometimes needs to invoke chalk-solve, which is why we want to merge them into one crate.)

## Intermediate steps

If we merge the two crates into one, I think the steps would be roughly like this:

- [x] Rename `chalk_rules::RustIrSource` to `RustIrDatabase`
[...]

PR #215, opened 2019-04-17
Simplify crate structure

detrumi:
Implements changes described in #214


chalk-solve/src/coherence/orphan.rs
```diff
@@ -12,7 +13,11 @@ use failure::Fallible;
 //     forall<T> { LocalImplAllowed(MyType<T>: Trait) }
 //
 // This must be provable in order to pass the orphan check.
-pub fn perform_orphan_check<DB>(db: &DB, impl_id: ImplId) -> Fallible<()>
+pub fn perform_orphan_check<DB>(
+    db: &DB,
```
nikomatsakis: Pre-existing, but maybe we should make this `&impl ChalkSolveDatabase`

chalk-solve/src/wf.rs
```diff
@@ -26,6 +27,7 @@ pub enum WfError {
 #[derive(new)]
 pub struct WfSolver<'db, DB: ChalkSolveDatabase> {
     db: &'db DB,
+    solver_choice: SolverChoice,
```
nikomatsakis: Maybe we should make this store the *solver*, not the solver-choice. This would allow for more caching (basically, we would invoke `into_solver` from within the `new` method)

chalk-solve/src/coherence.rs
```diff
@@ -16,6 +17,7 @@ where
     DB: ChalkSolveDatabase,
 {
     db: &'db DB,
+    solver_choice: SolverChoice,
```
nikomatsakis: Similar to my comment in `wf.rs`, I think we could store the solver *here* (and instantiate it in the `new` method) instead of the `solver_choice`. But this is purely an efficiency thing.

chalk-solve/src/wf.rs
detrumi: I was wondering why [the commit that originally removed this](https://github.com/rust-lang/chalk/commit/53614c9dbdb5546161007c0446fe59f2b3b39a5e#diff-a48af376562b9620a110c2c78c1da2bdL30) did this, but it turns out that doing `self.solver.solve()` would require the solver to be mutable, which was avoided by creating a new one every time.

Changing the functions to take `&mut self` would probably pollute the code a bit.

detrumi:
Leaving the `SolverChoice` parameters as is, as the functions would require mutability otherwise. I was planning to update `crates.md` with the changes, but let wait until `ChalkSolveDatabase` can be removed once `program_clauses_that_could_match` is replaced.

chalk-solve/src/wf.rs
nikomatsakis: OK, it's fine, we can leave it as is for now.

chalk-solve/src/coherence/orphan.rs
```diff
@@ -13,14 +13,11 @@ use failure::Fallible;
 //     forall<T> { LocalImplAllowed(MyType<T>: Trait) }
 //
 // This must be provable in order to pass the orphan check.
-pub fn perform_orphan_check<DB>(
-    db: &DB,
+pub fn perform_orphan_check(
+    db: &ChalkSolveDatabase,
```
nikomatsakis:
```suggestion
    db: &dyn ChalkSolveDatabase,
```

matklad: Might be a good idea and E-good-first-bug to add this check to CI

nikomatsakis: I suppose we should just add `#[deny(rust_2018_idioms)]` or whatever it is

PR #216, opened 2019-04-21
Port DomainGoal matching from rustc

detrumi:
See the part on `program_clauses_that_could_match` in #214 

Notes from comments that will probably be done outside of this PR:
- [Refactor out extra `could_match` check](https://github.com/rust-lang/chalk/pull/216#discussion_r277438825)
- [Remove `environment` function and its users](https://github.com/rust-lang/chalk/pull/216#discussion_r278868648), [remove `environment` usage in bench](https://github.com/rust-lang/chalk/pull/216#discussion_r278869627)


src/db.rs
```diff
@@ -69,14 +73,43 @@ impl ChalkDatabase {
...
+        };
+
         if let Ok(env) = self.environment() {
-            vec.extend(
-                env.program_clauses
-                    .iter()
-                    .filter(|&clause| clause.could_match(goal))
-                    .cloned(),
-            );
+            env.program_clauses_for_env(self, &mut clauses);
```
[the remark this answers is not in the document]
detrumi: Oops, easy mistake to make though, since they're identical structs modulo names :)

Good tip on inlining the logic, that also removed `ChalkSolveDatabase` in one go.

chalk-solve/src/solve/slg.rs
```diff
@@ -317,6 +322,142 @@ fn into_ex_clause(result: UnificationResult, ex_clause: &mut ExClause<SlgContext
     ex_clause.constraints.extend(result.constraints);
 }
 
+/// Returns a set of program clauses that could possibly match
+/// `goal`. This can be any superset of the correct set, but the
+/// more precise you can make it, the more efficient solving will
+/// be.
+fn program_clauses_that_could_match(
```
nikomatsakis: Nit: I think this logic should live in the `crate::clauses` module, actually. I suspect this is the only real consumer of the `ToProgramClauses` trait.

chalk-solve/src/solve/slg/clause_visitor.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl<'db, 'set> ClauseVisitor<'db, 'set> {
+    pub fn new(
+        program: &'db dyn RustIrDatabase,
+        round: &'set mut FxHashSet<ProgramClause>,
+    ) -> Self {
+        ClauseVisitor { program, round }
+    }
+
+    fn visit_ty(&mut self, ty: Ty) {
```
nikomatsakis: I guess these can be `&Ty` (and so forth for the other `visit` methods)

src/db.rs
```diff
@@ -67,19 +63,6 @@ impl ChalkDatabase {
     }
 }
 
-impl ChalkSolveDatabase for ChalkDatabase {
```
nikomatsakis: Is the `self.environment()` method dead code at this point? I would expect so.

chalk-solve/src/solve/slg.rs
```diff
@@ -322,112 +322,6 @@ fn into_ex_clause(result: UnificationResult, ex_clause: &mut ExClause<SlgContext
...
-        TypeKindId::TraitId(trait_id) => program
-            .trait_datum(trait_id)
-            .to_program_clauses(program, clauses),
-        TypeKindId::StructId(struct_id) => program
-            .struct_datum(struct_id)
-            .to_program_clauses(program, clauses),
-    }
-}
-
 fn program_clauses_for_env<'db>(
```
nikomatsakis: I think this should move to clauses too -- along with the "clause visitor"

chalk-solve/src/clauses.rs
```diff
@@ -952,3 +953,109 @@ impl ToProgramClauses for AssociatedTyDatum {
         );
```
nikomatsakis: I think we should move the `ToProgramClauses` trait and impls into a sub-module, and make it `pub(crate)` as it is never used outside this crate (or even `pub(super)`, to be more specific. The `clauses` module should probably just export a single function, like

```rust
/// Given some goal `goal` that must be proven, along with
/// its `environment`, figures out the program clauses that apply
/// to this goal from the Rust program. So for example if the goal
/// is `Implemented(T: Clone)`, then this function might return clauses
/// derived from the trait `Clone` and its impls.
pub fn program_clauses_for_goal(
    program: &dyn RustIrDatabase, // nit: I think we should rename this to `db`
    environment: &Environment, // I forget the exact name of this type
    goal: &DomainGoal,
) -> Vec<ProgramClause> {
    let mut vec = vec![];
    program_clauses_that_could_match(program, goal, &mut vec);
    program_clauses_for_env(program, environment, &mut vec);
    vec.retain(|c| c.could_match(goal));
    vec
}
```

chalk-solve/src/clauses.rs
```diff
@@ -952,3 +953,109 @@ impl ToProgramClauses for AssociatedTyDatum {
...
+    program: &dyn RustIrDatabase,
+    goal: &DomainGoal,
+    vec: &mut Vec<ProgramClause>,
+) {
+    let mut clauses = vec![];
+    match goal {
+        DomainGoal::Holds(WhereClause::Implemented(trait_ref)) => {
+            program
+                .trait_datum(trait_ref.trait_id)
+                .to_program_clauses(program, &mut clauses);
```
nikomatsakis: Something seems fishy here -- I'm surprised this never has to lower the *impls* of the given trait

chalk-solve/src/clauses.rs
```diff
@@ -952,3 +953,109 @@ impl ToProgramClauses for AssociatedTyDatum {
...
+/// Returns a set of program clauses that could possibly match
+/// `goal`. This can be any superset of the correct set, but the
+/// more precise you can make it, the more efficient solving will
+/// be.
+pub fn program_clauses_that_could_match(
+    program: &dyn RustIrDatabase,
+    goal: &DomainGoal,
+    vec: &mut Vec<ProgramClause>,
+) {
+    let mut clauses = vec![];
```
nikomatsakis: If we adopt the "outer, public" function I mentioned above, we can remove this intermediate vector and just write directly to `vec`

chalk-solve/src/clauses.rs
```diff
@@ -952,3 +953,109 @@ impl ToProgramClauses for AssociatedTyDatum {
...
+            match_ty(program, goal, &normalize.ty, &mut clauses)
+        }
+        DomainGoal::InScope(type_kind_id) => match_type_kind(program, *type_kind_id, &mut clauses),
+        DomainGoal::LocalImplAllowed(trait_ref) => program
+            .trait_datum(trait_ref.trait_id)
+            .to_program_clauses(program, &mut clauses),
+        DomainGoal::Compatible(()) => (),
+    };
+
+    vec.extend(
```
nikomatsakis: In my suggestion, I moved this logic into the outer function

chalk-solve/src/clauses.rs
nikomatsakis: Indeed, in the [rust source code](https://github.com/rust-lang/rust/blob/3d720d7e617442aaf65df0910e8c9f0fdb0cf31c/src/librustc_traits/chalk_context/program_clauses/mod.rs#L72-L76), we do seem to lower the rules from the impl, although in a mildly confusing way.

nikomatsakis: So I think we need something like

```rust
for impl_id in program.impls_for_trait(trait_ref.trait_id) {
    program.impl_datum(impl_id).to_program_clauses(program, &mut clauses);
}
```

nikomatsakis: Actually i'm a bit confused by the rustc source, maybe @scalexm can explain. The comment says:

```rust
                // These come from:
                // * implementations of the trait itself (rule `Implemented-From-Impl`)
                // * the trait decl (rule `Implemented-From-Env`)
```

but the source seems to (a) lower the associated items from an impl and (b) not the trait declaration itself. I haven't looked more deeply though and I'm sure it all works out.

chalk-solve/src/clauses.rs
detrumi: Makes sense!

detrumi: `ToProgramClauses` is still used here though: https://github.com/rust-lang/chalk/blob/master/src/query.rs#L126

src/db.rs
detrumi: It's still used in [`chalki`](https://github.com/rust-lang/chalk/blob/master/src/bin/chalki.rs#L199) and a [benchmark](https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/solve/test/bench.rs#L23), but maybe those can be changed...

nikomatsakis:
@detrumi I removed the "draft status" because i was hoping that would trigger travis -- but I still don't see travis checks, so i'm not sure what's up with that

chalk-solve/src/clauses/clause_visitor.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::clauses::match_type_kind;
+use crate::DomainGoal;
+use crate::FromEnv;
+use crate::ProgramClause;
+use crate::RustIrDatabase;
+use crate::Ty;
+use chalk_ir::TypeName;
+use rustc_hash::FxHashSet;
+
+pub struct ClauseVisitor<'db, 'set> {
```
nikomatsakis:
```suggestion
pub(super) struct ClauseVisitor<'db, 'set> {
```

nikomatsakis: Actually, i'd probably make the type private to this module, and just add a public function as the interface. Something like

```rust
pub(super) fn elaborate_env_clause(
    db: &dyn RustIrDatabase, 
    in_clause: &ProgramClause,
    out: &mut FxHashSet<ProgramClause>,
)
```

It'd be good to add a comment, too. The point of this is -- if we are trying to prove a `FromEnv` goal, then we have to elaborate all the from-env goals in our environment. i.e., if we have `T: Clone` in the environment, that might (indirectly) allow us to prove `T: Copy`. We find that out by adding in the clauses from the `trait Clone`, which include a rule

```rust
FromEnv(T: Copy) :- FromEnv(T: Clone)
```

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -0,0 +1,871 @@
...
+use chalk_ir::fold::shift::Shift;
+use chalk_ir::fold::Subst;
+use chalk_ir::*;
+use chalk_rust_ir::*;
+use std::iter;
+
+/// Trait for lowering a given piece of rust-ir source (e.g., an impl
+/// or struct definition) into its associated "program clauses" --
+/// that is, into the lowered, logical rules that it defines.
+pub trait ToProgramClauses {
```
nikomatsakis:
```suggestion
pub(super) trait ToProgramClauses {
```

doesn't matter, but it's nice to be precise

chalk-solve/src/clauses.rs
nikomatsakis: Oh, sorry, missed this -- yeah, we have to remove the `environment` query, and its users. We can leave it for later potentially.

src/db.rs
nikomatsakis: Hmm, actually, that `chalki` use is kind of handy as a debugging tool. Let's leave it as is for now. We might eventually move to allowing you to lower not *the whole program* but *some specific piece of IR* -- but either way the `ToProgramClauses` trait would have to stay public.

nikomatsakis: I think the benchmark however can be altered, we can remove `env` and just do `solve_root_goal(db, ...)`.

detrumi:
> do the tests pass?

Not yet, there's still a lot (42) of failing tests. Some "does not meet well-formedness requirements" errors, and a lot of diffs where no solutions were found.

There's also tests that fail around `LocalImplAllowed`, do we need to handle `DomainGoal::IsLocal` and such specially while matching? I wouldn't expect it, since that's not included in the rustc code either, but the tests are still failing...

```
failures:
    test::auto_trait_with_impls
    test::auto_trait_without_impls
    test::coherence::downstream_impl_of_fundamental_43355
    test::coherence::fundamental_traits
    test::coherence::overlapping_assoc_types_error
    test::coherence::two_blanket_impls_open_ended
    test::coinductive_semantics
    test::forall_projection
    test::forall_projection_gat
    test::fundamental_types
    test::gat_implied_bounds
    test::is_fully_visible
    test::local_and_upstream_types
    test::local_impl_allowed_for_traits
    test::mixed_indices_normalize_application
    test::mixed_indices_normalize_gat_application
    test::normalize_basic
    test::normalize_gat1
    test::normalize_gat2
    test::normalize_gat_with_higher_ranked_trait_bound
    test::normalize_gat_with_where_clause
    test::normalize_gat_with_where_clause2
    test::normalize_rev_infer
    test::normalize_rev_infer_gat
    test::normalize_under_binder
    test::projection_from_env
    test::projection_from_env_slow
    test::slg::cached_answers_1
    test::slg::cached_answers_2
    test::slg::cached_answers_3
    test::slg::contradiction
    test::slg::example_2_1_EWFS
    test::slg::example_2_2_EWFS
    test::slg::example_2_3_EWFS
    test::slg::example_3_3_EWFS
    test::slg::negative_answer_delayed_literal
    test::slg::negative_loop
    test::unselected_projection
    test::unselected_projection_with_gat
    test::unselected_projection_with_parametric_trait
    test::wf::assoc_type_recursive_bound
    test::wf::cyclic_wf_requirements
```

chalk-solve/src/clauses/program_clauses.rs
detrumi: `ToProgramClauses` is still used elsewhere, as noted in other comments.

nikomatsakis:
@detrumi ok, cool -- yeah I guess there's nothing for it but to start digging in closer into the code, I guess we're just missing the right rules. Have you looked at [the Lowering Rules rustc-guide chapter](https://rust-lang.github.io/rustc-guide/traits/lowering-rules.html)? Although, I suppose, it's probably not complete (sigh) -- I don't think we ever added the `IsLocal` and al lthat stuff to it.

chalk-solve/src/lib.rs
```diff
@@ -61,16 +61,6 @@ pub trait RustIrDatabase {
...
-/// solving a goal.
-pub trait ChalkSolveDatabase: RustIrDatabase + Debug {
-    /// Returns a set of program clauses that could possibly match
-    /// `goal`. This can be any superset of the correct set, but the
-    /// more precise you can make it, the more efficient solving will
-    /// be.
-    fn program_clauses_that_could_match(&self, goal: &DomainGoal, vec: &mut Vec<ProgramClause>);
-}
-
 pub use solve::Solution;
```
flodiebold: Would you mind exposing `solve::Guidance` here as well? It's rather hard to use `Solution` without being able to destructure `Guidance` ;) (I could of course create a PR, but since we're depending on this branch anyway, it's just the one line, and it would probably conflict with the deleted lines above...)

detrumi: Done :+1:

nikomatsakis:
This now passes tests on stable (fails on nightly due to `Debug` changes)

chalk-solve/src/clauses/clause_visitor.rs
```diff
@@ -0,0 +1,112 @@
...
+        match from_env {
+            FromEnv::Trait(trait_ref) => {
+                let mut clauses = vec![];
+                let trait_datum = self.db.trait_datum(trait_ref.trait_id);
+
+                trait_datum.to_program_clauses(self.db, &mut clauses);
+
+                // If we know that `T: Iterator`, then we also know
+                // things about `<T as Iterator>::Item`, so push those
+                // implied bounds too:
```
scalexm: Right, this isn’t done in rustc right now because bounds on associated items are converted in where clauses like in `where <Self as Trait>::Item: Foo`, but we’ll need this once we have GATs and cannot make this conversion.

chalk-solve/src/clauses/clause_visitor.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            FromEnv::Ty(ty) => self.visit_ty(ty),
+        }
+    }
+
+    fn visit_domain_goal(&mut self, domain_goal: &DomainGoal) {
+        match domain_goal {
+            DomainGoal::FromEnv(from_env) => self.visit_from_env(from_env),
+            DomainGoal::Holds(WhereClause::ProjectionEq(ProjectionEq { projection, ty: _ })) => {
+                self.visit_projection_ty(projection)
```
scalexm: I’m not sure this is useful; you would not have any useful implied bounds if you just have e.g. `where ProjectionEq(<T as Iterator>::Item = HashSet<K>)`

nikomatsakis: Maybe it's not needed. I'm experimenting with removing it.

Issue #220, opened 2019-05-03
get tests working with nightly again, strip commas?


nikomatsakis:
If you run `cargo +stable test --all`, all the tests pass. If you run `cargo +nightly test --all`, they do not. This is because the details of `Debug` formatting changed: there are now trailing commas where there didn't used to be. We could edit all the tests, but I think maybe it'd be better to just modify the code that compares debug values to strip trailing commas =)

Example test failure:

```
- constraints: []
+ constraints: [],
  },
- binders: []
+ binders: [],
  },
  delayed_literals: DelayedLiteralSet {
- delayed_literals: {}
- }
- }
+ delayed_literals: {},
+ },
+ },
```

nikomatsakis:
[Zulip thread for discussion](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/chalk.20tests.20on.20nightly.20chalk.23220)

PR #221, opened 2019-05-03
Adding the ability to parse 'default type Foo = Bar' when implementing a trait

sunjay:
Should be an easy to review PR. This implements the first step laid out by @nikomatsakis here: https://github.com/rust-lang/chalk/issues/219#issuecomment-489219272

> As @sunjay pointed out on Zulip, an obvious first step is:
> 
> Extend the parser to support `default type Foo = Bar` inside of an impl; that should be a flag on the [associated type value](https://github.com/rust-lang/chalk/blob/3238c6d0bc7d09b6eebe069ebd47dfabae787082/chalk-parse/src/ast.rs#L142-L147).

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -144,6 +144,7 @@ pub struct AssocTyValue {
     pub name: Identifier,
     pub parameter_kinds: Vec<ParameterKind>,
     pub value: Ty,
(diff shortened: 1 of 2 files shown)


sunjay:
As expected from #220 this passes on stable but not nightly.

PR #223, opened 2019-05-05
Slightly nicer Debug impls for TypeId, StructId and TraitId

flodiebold:
Without the TLS, all of these gave something like `RawId { index: 0 }`, which
omits the interesting information (whether it's a type, struct or trait) but
leaves the irrelevant `RawId`. This is more helpful, I think.


flodiebold:
The nightly test failure seems to only be the comma problem.

nikomatsakis:
@flodiebold can you rebase?

flodiebold:
@nikomatsakis Done!

Issue #224, opened 2019-05-06
improve test organization


detrumi:
Currently, the main Chalk is mainly an elaborate testing harness, which should be moved to a different crate. The tests themselves should then probably be moved to a `tests` directory.

First, clear out the main crate:
- [ ] Move the tests and test helpers to a new testing crate
- [ ] Move the repl (`chalki`) to its own crate
- [ ] Make `chalk-solve` the main crate (and call it `chalk`)

Then, organize the tests:
- [ ] Move tests to a `tests` directory
- [ ] Separate tests into categories
- [ ] Test improvements? (maybe use insta snapshots?)

matklad:
I'll also through in an idea of using test coverage marks: 

https://github.com/rust-analyzer/rust-analyzer/blob/ef782adc293deb287128f005dbab2038ba3ccdc1/crates/test_utils/src/marks.rs

nikomatsakis:
This is basically fixed by #271

nikomatsakis:
Closing this as fixed, I moved the test coverage idea to #278.

PR #225, opened 2019-05-07
Move tests out of main crate

detrumi:
First part of #224 

Some things might be moved around a bit more, since there's some problems with the current approach:
- `cargo test` is not finding the tests without `--all` (I might just move the tests to a `tests` folder already)
- the `chalki` binary isn't accessible from the main crate without mixing dependencies (though moving it to a `/bin` directory probably has the same problem)

Also, moving the `chalk-solve` contents to the root crate runs into all kinds of cyclical dependencies between crates, so that guides the design towards splitting up.


nikomatsakis:
I'm going to close this, sorry @detrumi, we failed to act on it fast enoug!

Issue #203, opened 2019-01-20
Trait objects


scalexm:
We need to design rules for the well-formedness of trait objects that address the soundness issue https://github.com/rust-lang/rust/issues/44454 and that are future-proof under implied bounds.

In what follows, we'll write `dyn Trait` to actually denote `dyn Trait + 'a` for a placeholder lifetime `'a`, the rules will work regardless of `'a`.

## Trait objects in rustc
Say we have the following trait:
```rust
trait Foo<T1...Tk> where WC1, ..., WCn { }
```
We will assume that this trait has no methods -- methods only change a bit the definition of "object-safety". Then we say that the trait is object-safe if:
* `Self` does not appear as a non-self type parameter in the where clauses `WC1`, ..., `WCn` or in the where clauses of any super-trait of `Foo<T1...Tk>`
* `Sized` is not a transitive super-trait of `Foo<T1...Tk>`

Then `dyn Foo<...>` is considered well-formed if `Foo` is object-safe. Example of object-safe traits:
```rust
trait Bar { } // Object-safe
trait Baz { } // Object-safe
trait Foo: Bar + Baz { } // Object-safe
```
Example of non-object safe traits:
```rust
trait Bar: Sized { } // Not object-safe
trait Baz: Bar { } // Not object-safe: `Sized` is an (indirect) super-trait of `Baz`

// Not object-safe: `Self` appears as a non-self type parameter in the where clauses
trait Foo where Self: PartialEq<Self> { }

// Not object-safe: `Self` appears as a non-self type parameter in the where clauses
// of a super-trait of `FooBar`.
trait FooBar: Foo { }
```

Then, whenever a trait `Foo<T1...Tk>` is object-safe, for each super-trait `SuperTrait<...>` of `Foo<T1...Tk>` (including `Foo<T1...Tk>` itself), the compiler *unconditionally* generates the following impl:
```rust
impl<T1...Tk> SuperTrait<...> for dyn Foo<T1...Tk> { }
```

## Why the previous approach is unsound
The unsoundness comes from the generated impls above. For example, `dyn Foo<&'a i32>` in this example is a well-formed trait object:
```rust
trait Foo<X>: 'static { } // Object-safe

// The compiler generates this impl:
// impl<X> Foo<X> for dyn Foo<X> { }

fn bar<'a>() {
    let x: Box<dyn Foo<&'a i32>>;
}
```
[...]

PR #226, opened 2019-05-08
Add basic support for dyn and impl Trait

### LINKED ISSUE #218: add basic support for `dyn Trait` and `impl Trait`
**Status:** @KiChjang is working on the first steps.

---

The goal here is to extend chalk with basic support for `dyn Trait` and (opaque) `impl Trait` types. This involves a few steps:

- [x] extend chalk-parser to support `dyn` and `impl` syntax, along with the ast
- [x] extending `chalk-ir` to be able to represent such types
- [x] extending `chalk-solve` so that when we have to prove something like `dyn Foo: Foo`, we judge that to be true (similarly `impl Foo: Foo`) -- blocked on #216 
- [ ] adding tests for cases like
    - [ ] unifying `dyn Foo` with a struct type (error)
    - [ ] unifying `impl Foo` with a struct type (error)
    - [x] checking that `dyn Foo: Foo` (without any impl)
    - [x] checking that `dyn Foo: Bar` does *not* hold
    - [x] checking that `impl Foo: Foo`  (without any impl)
    - [x] checking that `impl Foo: Bar` does *not* hold
        - hmm, we're going to have to figure out "auto trait leakage"

This is judged to be a "medium" complex project, suitable for a dedicated beginner. After this work is done, the next step is https://github.com/rust-lang/chalk/issues/203.

KiChjang:
cc #218.

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -827,6 +827,10 @@ impl<T> Binders<T>
 where
     T: Fold,
 {
+    /// Substitute `parameters` for the variables introduced by these
+    /// binders. So if the binders represent (e.g.) `<X, Y> { T }` and
+    /// parameters is the slice `[A, B]`, then returns `[X => A, Y =>
+    /// B] T`.
     pub fn substitute(&self, parameters: &[Parameter]) -> T::Result {
         assert_eq!(self.binders.len(), parameters.len());
         Subst::apply(parameters, &self.value)
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -239,6 +239,9 @@ pub enum Ty {
     /// In the chalk parser, the traits that the object represents is parsed as
     /// a QuantifiedInlineBound, and is then changed to a list of where clauses
     /// during lowering.
+    ///
+    /// See the `Opaque` variant for a discussion about the use of
+    /// binders here.
     Dyn(Binders<Vec<QuantifiedWhereClause>>),
 
     /// An "opaque" type is one that is created via the "impl Trait" syntax.
@@ -246,6 +249,25 @@ pub enum Ty {
     /// is unknown, and hence the type is opaque to us. The only information
     /// that we know of is that this type implements the traits listed by the
     /// user.
+    ///
+    /// The "binder" here represents the unknown self type. So, a type like
+    /// `impl for<'a> Fn(&'a u32)` would be represented with two-levels of
+    /// binder, as "depicted" here:
+    ///
+    /// ```
+    /// exists<type> {
+    ///    vec![
+    ///        // A QuantifiedWhereClause:
+    ///        forall<region> { ^1: Fn(&^2 u32) }
+    ///    ]
+    /// }
+    /// ```
+    ///
+    /// The outer `exists<type>` binder indicates that there exists
+    /// some type that meets the criteria within, but that type is not
+    /// known. It is referenced within the type using `^1`, indicating
+    /// a bound type with debruijn index 1 (i.e., skipping through one
+    /// level of binder).
     Opaque(Binders<Vec<QuantifiedWhereClause>>),
 
     /// A "projection" type corresponds to an (unnormalized)
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -25,7 +25,7 @@ macro_rules! impl_debugs {
 use crate::cast::Cast;
 use crate::fold::shift::Shift;
 use crate::fold::{
-    DefaultInferenceFolder, DefaultPlaceholderFolder, DefaultTypeFolder, Fold, FreeVarFolder,
+    DefaultInferenceFolder, DefaultPlaceholderFolder, DefaultTypeFolder, Fold, FreeVarFolder, Subst,
 };
 use chalk_engine::fallible::*;
 use lalrpop_intern::InternedString;
@@ -820,6 +820,13 @@ impl<T> Binders<T> {
 
     pub fn len(&self) -> usize {
         self.binders.len()
+    }
+}
+
+impl<T> Binders<T> where T: Fold {
+    pub fn substitute(&self, parameters: &[Parameter]) -> T::Result {
+        assert_eq!(self.binders.len(), parameters.len());
+        Subst::apply(parameters, &self.value)
     }
 }
 
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -258,7 +258,7 @@ pub enum Ty {
     /// exists<type> {
     ///    vec![
     ///        // A QuantifiedWhereClause:
-    ///        forall<region> { ^1: Fn(&^2 u32) }
+    ///        forall<region> { ^1: Fn(&^0 u32) }
     ///    ]
     /// }
     /// ```
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -254,7 +254,7 @@ pub enum Ty {
     /// `impl for<'a> Fn(&'a u32)` would be represented with two-levels of
     /// binder, as "depicted" here:
     ///
-    /// ```
+    /// ```notrust
     /// exists<type> {
     ///    vec![
     ///        // A QuantifiedWhereClause:
@@ -823,7 +823,10 @@ impl<T> Binders<T> {
     }
 }
 
-impl<T> Binders<T> where T: Fold {
+impl<T> Binders<T>
+where
+    T: Fold,
+{
     pub fn substitute(&self, parameters: &[Parameter]) -> T::Result {
         assert_eq!(self.binders.len(), parameters.len());
         Subst::apply(parameters, &self.value)
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -235,6 +235,19 @@ pub enum Ty {
     /// an empty list).
     Apply(ApplicationTy),
 
+    /// A "dyn" type is a trait object type created via the "dyn Trait" syntax.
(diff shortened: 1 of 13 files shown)


src/lowering.rs
```diff
@@ -876,6 +876,18 @@ impl LowerTy for Ty {
...
+            Ty::Dyn { bounds } => Ok(chalk_ir::Ty::Dyn(
+                bounds.lower()?
+                    .flat_map(|qil| qil.into_where_clauses(chalk_ir::Ty::BoundVar(qil.binders.len())))
+                    .collect()
+            )),
+
+            Ty::Opaque { bounds } => Ok(chalk_ir::Ty::Opaque(
+                bounds.lower()?
+                    .flat_map(|qil| qil.into_where_clauses(chalk_ir::Ty::BoundVar(qil.binders.len())))
+                    .collect()
```
KiChjang: I'm not 100% sure that I'm lowering these correctly. Can someone confirm?

src/lowering.rs
```diff
@@ -876,6 +876,18 @@ impl LowerTy for Ty {
                 NameLookup::Parameter(d) => Ok(chalk_ir::Ty::BoundVar(d)),
             },
 
+            Ty::Dyn { bounds } => Ok(chalk_ir::Ty::Dyn(
+                bounds.lower()?
+                    .flat_map(|qil| qil.into_where_clauses(chalk_ir::Ty::BoundVar(qil.binders.len())))
+                    .collect()
```
nikomatsakis: Given the definitions I was suggesting above, I think we want:

```rust
env.in_binders(
    Some(chalk_ir::ParameterKind::Ty(XXX)),
    |env| {
        bounds
            .lower(env)?
            .flat_map(|qil| qil.into_where_clauses(chalk_ir::Ty::BoundVar(0)))
            .collect()
    },
)
```

with the caveat that I don't know what the right thing is to put in the `XXX`. There isn't really a right thing -- that's supposed to be the name the user would use to refer to this variable, and there is no such name. We could as a hack put in the string `__FIXME_SELF__` or something.

Also, argh, this code setup is pretty messy.

nikomatsakis: Why I think this is right:

- Invoking `in_binders` should introduce a new level of binders, such that when you are lowering the `bounds` vector, it will get the debruijn index arithmetic correct
- Invoking `into_where_clauses` will then generate the right thing, and the "self type" should be that type parameter at index 0

nikomatsakis:
@KiChjang so I didn't actually see your WIP commit until just now. I'm taking a look now at the issue and trying to figure out what the next steps are. Will try to leave you some advice =)

KiChjang:
@nikomatsakis I still need some guidance on how to extend the chalk solver to account for impl Trait and dyn Trait.

chalk-ir/src/lib.rs
```diff
@@ -226,13 +226,35 @@ pub enum Ty {
...
+    ///
+    /// The "binder" here represents the unknown self type. So, a type like
+    /// `impl for<'a> Fn(&'a u32)` would be represented with two-levels of
+    /// binder, as "depicted" here:
+    ///
+    /// ```
+    /// exists<type> {
+    ///    vec![
+    ///        // A QuantifiedWhereClause:
+    ///        forall<region> { ^1: Fn(&^2 u32) }
```
bjorn3: Should the `^2` be `^0`?

nikomatsakis: Sigh. Yes.

chalk-solve/src/clauses.rs
```diff
@@ -167,6 +167,23 @@ fn program_clauses_that_could_match(
                 }
             }
 
+            // Check for impl and dyn Traits so that we generate
+            // `Implemented(impl Foo + Bar) :- Implemented(Foo), Implemented(Bar)`
+            match trait_ref.self_type_parameter() {
+                Some(Ty::Opaque(qwc)) | Some(Ty::Dyn(qwc)) => {
+                    clauses.push(ProgramClauseImplication {
+                        consequence: trait_ref.clone().cast(),
```
nikomatsakis: This doesn't *quite* look right. For example, if this function were invoked with a `trait_ref` like `Implemented(dyn Foo: Bar)`, this would generate something like

```
implemented(dyn Foo: Bar) :-
    dyn Foo: Foo
```

but what we really want is 

```
Implemented(dyn Foo: Foo).
```

Moreover, something is a bit funny here -- you're just skipping over the binders in `qwc`, which isn't really right. What we want is more like:

```
let self_ty = trait_ref.self_type_parameter().unwrap(); // this should not really return an Option
let wc = qwc.instantiate(self_ty);
clauses.extend(wc.iter().cloned().casted());
```

what this will do is to take a `self_ty` like `dyn Foo` and create clauses like `Implemented(dyn Foo: Foo)`.

However, there are some complications.

For one, the `instantiated` method I referred to doesn't actually *exist* yet -- there is `subst::apply(&[self_ty], &qwc.value)`, which I think would be equivalent. The idea of `instantiate` would be to take a `Binders<T>` and create a `T`, substituting a value for the variables bound in `Binders`.

nikomatsakis:
Ok I took a stab at building this branch locally and I see errors in the following files. Some notes:

---

```
error[E0004]: non-exhaustive patterns: `&Dyn(_)` and `&Opaque(_)` not covered
  --> chalk-solve/src/clauses/env_elaborator.rs:49:15
   |
49 |         match ty {
   |               ^^ patterns `&Dyn(_)` and `&Opaque(_)` not covered
   |
   = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

These have to do with implied bounds. I would leave these rules as empty but add "FIXME #203" for now.

---

```
error[E0004]: non-exhaustive patterns: `(&Dyn(_), &Apply(_))`, `(&Opaque(_), &Apply(_))`, `(&Dyn(_), &Dyn(_))` and 9 more not covered
   --> chalk-solve/src/infer/unify.rs:107:15
    |
107 |         match (a, b) {
    |               ^^^^^^ patterns `(&Dyn(_), &Apply(_))`, `(&Opaque(_), &Apply(_))`, `(&Dyn(_), &Dyn(_))` and 9 more not covered
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

You basically want rules very analogous to the `Apply` rules. Although, hmm, it occurs to me that we're going to have some issues around sorting the where clauses in the types -- but let's just assume they're in a canonical ordering for now, we can patch that up later. (i.e., for now we'd say that `dyn (Foo + Bar)` and `dyn (Bar + Foo)` are distinct types -- later we'd assume some sort of normalization, so this code wouldn't change in any case).

So you want to add an analogous match arm like this one:

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/infer/unify.rs#L133-L139

and then add `Dyn/Opaque` to these match arms:

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/infer/unify.rs#L119-L122

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/infer/unify.rs#L141-L149

---

```
error[E0004]: non-exhaustive patterns: `(&Dyn(_), _)` and `(&Opaque(_), _)` not covered
   --> chalk-solve/src/solve/slg/aggregate.rs:171:15
    |
171 |         match (ty0, ty1) {
    |               ^^^^^^^^^^ patterns `(&Dyn(_), _)` and `(&Opaque(_), _)` not covered
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

For now we could probably handle "de-aggregation" in the same "not that great" way we handle some other cases, like so:

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/solve/slg/aggregate.rs#L183-L185

but we should add a FIXME for this. (Also, this code seems a bit wrong to me, in that I think it should try to ensure that `aggregate_tys(foo, foo)` also returns `foo`, but oh well, that's a separate bug, or maybe I'm missing something.)

---

```
error[E0004]: non-exhaustive patterns: `(&Dyn(_), &Apply(_))`, `(&Opaque(_), &Apply(_))`, `(&Dyn(_), &Dyn(_))` and 9 more not covered
   --> chalk-solve/src/solve/slg/resolvent.rs:342:15
    |
342 |         match (answer, pending) {
    |               ^^^^^^^^^^^^^^^^^ patterns `(&Dyn(_), &Apply(_))`, `(&Opaque(_), &Apply(_))`, `(&Dyn(_), &Dyn(_))` and 9 more not covered
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

As the comment says:

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/solve/slg/resolvent.rs#L340-L341

we ought to be able to match up the structure and descend down, so we probably just want something like this

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/solve/slg/resolvent.rs#L347

but we have to be sure the `Zip` trait is implemented.

---

```
error[E0004]: non-exhaustive patterns: `&Dyn(_)` and `&Opaque(_)` not covered
  --> chalk-solve/src/wf.rs:68:15
   |
68 |         match self {
   |               ^^^^ patterns `&Dyn(_)` and `&Opaque(_)` not covered
   |
   = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

I think this should just behave analogously to apply: so push the type into the accumulate and recurse on the patterns, much like this

https://github.com/rust-lang/chalk/blob/9dac49b7bdce17b6294e0d811b457593b7aa0d2b/chalk-solve/src/wf.rs#L67-L70

Issue #217, opened 2019-04-29
non-enumerable goals


nikomatsakis:
One of the challenges we've found with our chalk solver strategy is selecting the search order. Imagine we have a setup like this (using a kind of short-hand):

```
Foo(?T) :- Sized(?T), Bar(?T).
Bar(Option<u32>).
```

Now, given a query like `Foo(?T)`, it's clear that it would be best to first try and solve `Bar(?T)`. That would yield up a single answer (`[?T := Option<u32>]`), and we can readily validate that this type is sized. If we were so unlucky s to pick `Sized(?T)`, however, what would happen?

The traditional answer -- the one chalk-engine is designed for -- is that we would start to enumerate sized types as "possible answers". So we might start with `?T := u32`, for example. We would then try to solve `Bar(u32)` and fail. Eventually, we would find `Option<u32>`, succeed, and propagate the final result back. If the user asked for *another* result, we would run off and exhaustively search *all possible sized types* until we got to an ambiguous result (since that set is infinite). Obviously not a win.

In reality, we might not even get that far -- enumerating the full set of sized types is not really trivial nor possible, since sized is a builtin trait within rustc. 

To some extent, we can solve this by manipulating search order. We have a hook now for doing that and we can probably create some heuristics that would encourage us to choose `Bar(u32)`.

This issue concerns one possible solution. It builds on the idea of "delay literals" but uses them for another purpose. The basic idea would that if we select a goal like `Sized(?T)`, we would classify it as 'non-enumeratable' -- i.e., it is not possible to enumerate all the solutions. We would then add it to our solution as a kind of delayed literal.

So we would go from an ex-clause like:

```
Foo(?T) :- | Sized(?T), Bar(?T)
```

to

```
Foo(?T) :- Sized(?T) | Bar(?T)
```

then we would solve `Bar(?T)`. This would give us the ex-clause:

```
Foo(Option<u32)> :- Sized(Option<u32>) |
```

This, under normal SLG, is a complete answer. And that might even be "ok" for us to propagate -- but we might also do better by trying to "undelay" `Sized(Option<u32>)` here, which, indeed, is now solveable.

I'm not *100%* sure how this should look -- we need to dig a bit more into how delayed literals are handled in the existing solver as well, since these are clearly tightly related.

PR #229, opened 2019-06-04
Introduce concept of floundering

nikomatsakis:
This introduces the idea of "floundering" into chalk-engine and uses it to mark auto-traits as non-enumerable, along with a new `#[non_enumerable]` traits (used for testing purposes). See the edits to the README.md in one of the final commits for a high-level summary.

Fixes https://github.com/rust-lang/chalk/issues/217

### DIFF
--- a/chalk-engine/src/README.md
+++ b/chalk-engine/src/README.md
@@ -1,7 +1,5 @@
 # The on-demand SLG solver
 
-## Description of how it works
-
 The basis of the solver is the `Forest` type. A *forest* stores a
 collection of *tables* as well as a *stack*. Each *table* represents
 the stored results of a particular query that is being performed, as
@@ -9,7 +7,7 @@ well as the various *strands*, which are basically suspended
 computations that may be used to find more answers. Tables are
 interdependent: solving one query may require solving others.
 
-### Walkthrough
+## Walkthrough
 
 Perhaps the easiest way to explain how the solver works is to walk
 through an example. Let's imagine that we have the following program:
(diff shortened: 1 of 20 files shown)


shepmaster:
I’d like to requisition the addition of the turboflounder.

nikomatsakis:
(The error on stable seems to have something to do with generating documentation, I think -- the tests run successfully.)

nikomatsakis:
I was thinking about this some more. We'll see, but I imagine we'll want to tinker the story a bit to try and give better "type hints" in some cases. I've not yet come up with a complete example but the idea is something like:

```rust
trait Foo { }
impl<T> Foo for Vec<T> { }
```

Here, if yo have `?T: Foo`, it's going to flounder, but it could in principle infer that `?T = Vec<?U>` for *some* `?U`.

matklad:
@nikomatsakis just out of curiosity, could you briefly explain (or point to an existing explanation) *why* the need to enumerate solutions arises at all? From my little understanding of how the trait solving works, it seems like one needs only to check things, and not to enumerate them. From your comment I gather that this can be useful at least for error reporting, but I wonder if, when doing completions, we can just disable error reporting? Or is there a more fundamental reason why we need to enumerate things?

chalk-engine/src/README.md
```diff
@@ -231,6 +240,122 @@ Here you can see how the forest captures both the answers we have
 created thus far *and* the strands that will let us try to produce
 more answers later on.
 
+### Floundering
+
+The first thing we do when we create a table is to initialize it with
+a set of strands. These strands represent all the ways that one can
+solve the table's associated goal. For an ordinary trait, we would
+effectively create one strand per "relevant impl". But sometimes the
+goals are vague for this to be possible; other times, it may be possible
```
tmandry: nit: "_too_ vague"

tmandry:
```suggestion
goals are too vague for this to be possible; other times, it may be possible
```

scalexm:
@matklad I believe you'd need to enumerate things when inferring types from some trait method calls, e.g. like in:
```rust
struct Bar<T>(T);

trait Foo {
    fn foo(&self);
}

impl Foo for Bar<u64> {
    fn foo(&self) { }
}

impl Bar<u64> {
    fn print_u64(&self) {
        println!("hello u64");
    }
}

impl Bar<u8> {
    fn print_u8(&self) {
        println!("hello u8");
    }
}

fn main() {
    let x = Bar(5);

    // To infer the type of `x: Bar<?X>`, we'd need to solve the following goal:
    // `exists<?X> { Implemented(Bar<?X>: Foo) }`
    // In this case, the goal won't flounder and `?X` will resolve to `u64`.
    x.foo();

    // auto-completion: `print_` -> `print_u64()`, because `x` has been inferred
    // to be of type `Bar<u64>`.
    y.print_
}
```

If the goal had returned more than one answer or had floundered, you'd report an error.

PR #230, opened 2019-06-08
Adding parsing and AST->IR lowering for RevealMode + Overrides + added reveal { G } syntax

sunjay:
This begins to implement the first bullet in https://github.com/rust-lang/chalk/issues/219#issuecomment-498370956.

I added a number of things:

1. A new `reveal { G }` syntax similar to the `compatible { G }` syntax we already have for introducing the `RevealMode` (this is useful for testing and whatnot)
2. Parsing for `RevealMode` and `Overrides[V0](P0: Trait<P1..Pn>)` syntax
3. Lowering from the AST to the chalk IR for all of this
4. All the standard `Debug`, `Fold`, `Zip` stuff for the new variants and types

Unresolved questions for @nikomatsakis before this is merged:

- [ ] Do I need to add/change anything in the rust_ir? I didn't encounter any compiler errors, so I wasn't sure
- [ ] Did I look up the associated type's TypeId correctly during lowering?
- [ ] What should we do when lowering `Overrides` in clauses.rs? Right now I've just left it as `unimplemented!()`
    * I could just do the same thing as when we added `LocalImplAllowed` and create program clauses for the trait ref, but do we also have to take the associated type ID into account?

I'm going to do the second part of this ("Modify the lowering for default type to include the `Overrides` clause") after this PR.

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -212,7 +212,11 @@ fn program_clauses_that_could_match(
         DomainGoal::LocalImplAllowed(trait_ref) => db
             .trait_datum(trait_ref.trait_id)
             .to_program_clauses(db, clauses),
-        DomainGoal::Compatible(()) => (),
+        DomainGoal::Compatible(()) | DomainGoal::RevealMode(()) => (),
+        DomainGoal::Overrides(ov) => {
+            //TODO(sunjay)
+            unimplemented!()
+        }
     };
 }
 
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -213,7 +213,10 @@ impl Debug for UnselectedNormalize {
 
 impl Debug for Overrides {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
-        let Overrides {assoc_ty_id, trait_ref} = self;
+        let Overrides {
+            assoc_ty_id,
+            trait_ref,
+        } = self;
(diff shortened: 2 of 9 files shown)


sunjay:
I recommend excluding the rustfmt commit when you review this because it adds a lot of unnecessary noise.

chalk-solve/src/clauses.rs
```diff
@@ -214,7 +214,11 @@ fn program_clauses_that_could_match(
         DomainGoal::LocalImplAllowed(trait_ref) => db
             .trait_datum(trait_ref.trait_id)
             .to_program_clauses(db, clauses),
-        DomainGoal::Compatible(()) => (),
+        DomainGoal::Compatible(()) | DomainGoal::RevealMode(()) => (),
+        DomainGoal::Overrides(ov) => {
```
nikomatsakis: So here we need to find all the "sources" of overrides clauses. I think these would derive from lowering the "assoc type values" found in impl of the given trait, so the simplest thing would be to call `push_program_clauses_for_associated_type_values_in_impls_of` like we do for `Normalize`.

XVilka:
There are conflicts now - so should be rebased.

jackh726:
We discussed this in the meeting today (https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/design.20meeting.202020.2E04.2E14/near/193955431) and decided to close this. I'm gonna link this in a specialization-related issue for potential followup though

PR #231, opened 2019-07-20
:arrow_up: ena

matklad:


matklad:
CI fails on nightly: formatting regression presumably due to update of rustfmt? Seems unrelated to PR

Issue #233, opened 2019-08-17
Consider replacing failure with std::error


matklad:
I've noticed that chalk used failure for error handing, but I wonder if this is really necessary?

It seems like today, failure actually doesn't buy much over `std::error` (only backtraces) and that it hasn't become the goto error handing solution for rust, as it once was hoped.

shepmaster:
I'll recommend the [SNAFU crate](https://github.com/shepmaster/snafu/) to reduce some of the boilerplate of implementing `std::error::Error`. I've had positive interactions with the crate author in the past 😉.

matklad:
Yeah, I've also heard good things about SNAFU, but I wonder if error-helpers make sense in the context of chalk at all? 

Like, chalk doesn't do IO, so all errors are "the input program is wrong", they are not "something randomly crashed 1000 frames deep in the guts of a framework".

nikomatsakis:
@shepmaster is the idea of SNAFU that `derive(Snafu)` generates the selector names that sit alongside the enum?

nikomatsakis:
Must be, I guess

shepmaster:
@matklad here are my thoughts:

> they are not "something randomly crashed 1000 frames deep in the guts of a framework".

"frames" and "stacktraces" should be the method of last resort for figuring out what went wrong. Instead, I like to think of a "semantic" stack trace. From other comments in the SNAFU repo:

> I roughly equate each error type to a *semantic* level in a backtrace. When something wrong happens in a program, I think it's a useful exercise to explain **as a human** what it was:
> 
> > 1. I couldn't upload the image **I** to the server **S** because 
> > 1. I couldn't load the username and password from the config file **C** because
> > 1. I couldn't find the file
> 
> ```rust
> #[derive(Debug, Snafu)]
> enum Error {
>     UploadFile { source: config::Error, image: path::PathBuf, server: String },
>     // ... 
> }
> 
> mod config {
>     #[derive(Debug, Snafu)]
>     enum Error {
>         LoadCredentials { source: io::Error, path: path::PathBuf }, 
>         // ...
>     }
> }
> ```

I don't know enough about chalk to say for certain, but I definitely have the belief that while it might not have 1000 stack frames, it does have numerous tricky cases that it wants to report, ideally reporting in a "nice" way.

Using the same kind of idea, poorly applied to chalk's domain:

- Unable to find a method `foo` for the type `Bar`. The following methods almost matched:
  - The method `foo` from the trait `Quux` does not apply
[...]

PR #236, opened 2019-08-31
Remove failure from some of the chalk's crates

matklad:
cc #233


matklad:
Note: this doesn't remove failure from `src/lowerning`, as I am not sure what's the intended design there: at the moment, it uses both dynamcally dispatched `failure::Error` type, as well as an error enum. I think it should use only one of these two pattern, but I am not sure which one!

Issue #234, opened 2019-08-23
Ambiguity problems with associated types


flodiebold:
While handling associated types in rust-analyzer, I ran into two problems that I want to ask about. These are probably not related, so I can split this up if necessary, but I wanted to discuss them a bit first :slightly_smiling_face: 

Number 1:
```rust
#[test]
fn normalize_into_iterator() {
    test! {
        program {
            trait IntoIterator { type Item; }
            trait Iterator { type Item; }
            struct Vec<T> { }
            struct u32 { }
            impl<T> IntoIterator for Vec<T> {
                type Item = T;
            }
            impl<T> IntoIterator for T where T: Iterator {
                type Item = <T as Iterator>::Item;
            }
        }

        goal {
            forall<T> {
                exists<U> {
                    Normalize(<Vec<T> as IntoIterator>::Item -> U)
                }
            }
        } yields {
            "Unique"
        }
    }
}
```
I'd expect this test to pass because the second `IntoIterator` impl should not have any effect at all here, but the result is instead ambiguous, because apparently the `Normalize` rule from the second impl applies too, because it just checks that the type implements `IntoIterator`, and not that it does so with *that impl*. That seems really wrong.

Number 2:
```rust
#[test]
fn projection_equality() {
    test! {
        program {
            trait Trait1 {
                type Type;
            }
            trait Trait2<T> { }
            impl<T, U> Trait2<T> for U where U: Trait1<Type = T> {}

            struct u32 {}
            struct S {}
            impl Trait1 for S {
                type Type = u32;
            }
        }

        goal {
            exists<U> {
                S: Trait2<U>
            }
        } yields {
            "Unique"
        }
    }
}
```
I'd expect this to pass, but it yields ambiguity, I think because the `ProjectionEq` is ambiguous between `u32` and `Trait1::Type<S>`. But those are really the same type. I found #74, but it's not clear to me how this kind of thing is supposed to work with `ProjectionEq`. #111 is maybe related?

nikomatsakis:
Hmm so digging into number 1, something does indeed seem a bit wrong. 

We wind up with these two program clauses:

```
for<type> Normalize(<Vec<^0> as IntoIterator>::Item -> ^0) :- 
    Implemented(Vec<^0>: IntoIterator)
for<type> Normalize(<^0 as IntoIterator>::Item -> <^0 as Iterator>::Item) :- 
    Implemented(^0: IntoIterator)
```

I think what I'd *expect* is something like

```
[...]

PR #237, opened 2019-09-03
Require where clauses from the impl on Normalize goal

### LINKED ISSUE #235: Normalization doesn't take into account which impl actually applied
Split off from #234:
```rust
#[test]
fn normalize_into_iterator() {
    test! {
        program {
            trait IntoIterator { type Item; }
            trait Iterator { type Item; }
            struct Vec<T> { }
            struct u32 { }
            impl<T> IntoIterator for Vec<T> {
                type Item = T;
            }
            impl<T> IntoIterator for T where T: Iterator {
                type Item = <T as Iterator>::Item;
            }
        }

        goal {
            forall<T> {
                exists<U> {
                    Normalize(<Vec<T> as IntoIterator>::Item -> U)
                }
            }
        } yields {
            "Unique"
        }
    }
}
```
I'd expect this test to pass because the second IntoIterator impl should not have any effect at all here, but the result is instead ambiguous, because apparently the Normalize rule from the second impl applies too, because it just checks that the type implements IntoIterator, and not that it does so with that impl. That seems really wrong.

Niko:

> We wind up with these two program clauses:
```
for<type> Normalize(<Vec<^0> as IntoIterator>::Item -> ^0) :- 
    Implemented(Vec<^0>: IntoIterator)

for<type> Normalize(<^0 as IntoIterator>::Item -> <^0 as Iterator>::Item) :- 
    Implemented(^0: IntoIterator)
```
> I think what I'd expect is something like
```
for<type> Normalize(<Vec<^0> as IntoIterator>::Item -> ^0)

for<type> Normalize(<^0 as IntoIterator>::Item -> <^0 as Iterator>::Item) :- 
    Implemented(^0: Iterator)
```
> In particular, inlining the where-clauses and conditions of the impl into the rule. @scalexm -- do you recall why we set things up the way we did?

scalexm:
> I don’t think there’s any good reason why the where clauses are not inlined.
> I probably thought it was equivalent because there would be a « perfect match » with the impl, but of course two impls can have a non-empty intersection with respect to their receiver type.

flodiebold:
This should fix #235.

(I also included the other test from #234 in ignored form, since I already had it there...)

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -52,7 +52,7 @@ impl ToProgramClauses for AssociatedTyValue {
     ///
     /// Then for the following impl:
     /// ```notrust
-    /// impl<T> Iterable for Vec<T> {
+    /// impl<T> Iterable for Vec<T> where T: Clone {
     ///     type IntoIter<'a> = Iter<'a, T>;
     /// }
     /// ```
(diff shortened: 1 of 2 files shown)


flodiebold:
Build failure is apparently just formatting on nightly

nikomatsakis:
we should figure out this formatting business

Issue #241, opened 2019-09-13
refine `RustIrDatabase` interface


nikomatsakis:
This issue is for discussing the design of [`RustIrDatabase` trait](https://github.com/rust-lang/chalk/blob/b4a6b655578ee35b1b3f6b8579636269cf3b0b1a/chalk-solve/src/lib.rs#L17).

nikomatsakis:
First objection:

The [`impls_for_trait`](https://github.com/rust-lang/chalk/blob/b4a6b655578ee35b1b3f6b8579636269cf3b0b1a/chalk-solve/src/lib.rs#L39) method should supply hints about the types of impls we're interested in -- e.g., the self type. This is a crucial performance optimization in rustc. 

This method is used for three purposes:

- finding "all local" impls for coherence purposes
- finding impls relevant to solving `Implemented(T: Trait)` goals
- finding impls relevant to projection goals -- here we are really looking for the associated type values found *inside* impls

nikomatsakis:
Second objection:

The `all_structs` method should be removed as well. We want to avoid any kind of "give me all the things" methods, I think, if possible.

This method is used to enumerate possible instantiations of auto traits where the self type is an unbounded inference variable; it's also used for some other kinds of internal goals, like `DownstreamType`, with similar reasons.

nikomatsakis:
I [started the `remove-unselected-projection` branch](https://github.com/nikomatsakis/chalk-ndm/tree/remove-unselected-projection) working towards these two goals. It does a few things:

- Introduced a method to use for coherence, which is kind of a special case
- Insisted that we flounder all goals where the "self type" is an unbounded inference variable
    - I think we could refine this rule in the future, but it matches what rustc does

I'm currently working through the impact of that rule. It makes some of the coherence related tests fail, but I'm not sure how important that is.

If we take this step, though, we still don't quite reach the goal I wanted because of interactions with lazy normalization. See [this comment for more details](https://github.com/rust-lang/chalk/issues/234#issuecomment-530823193).

nikomatsakis:
Another concern:

I think I would prefer to get away from removing "complex structs"...maybe? Not entirely sure about this. But it seems like it might be nice to have relatively "flat" methods, so that we can map more easily to whatever "HIR-like" data structures exist on the other side (and perhaps for better incremental tracking).

flodiebold:
@nikomatsakis The branch you linked doesn't contain anything related to this as far as I can see; is it the right one?

[...]

PR #243, opened 2019-09-20
Remove all structs callback

nikomatsakis:
This PR is work towards #241 -- in particular, it eliminates the `all_structs` callback and adds a separate callback for coherence. Next step (which can be in a separate PR) is to modify the `impls_for_trait` callback to take a self-type -- in the case of rust-analyzer, where we intentionally mark all traits as "non-exhaustive", that self-type should always be at least shallowly known (except that it could be a projection).

cc @flodiebold 
r? @scalexm

### DIFF
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -38,7 +38,7 @@ pub struct ImplDatumBound {
     pub impl_type: ImplType,
 }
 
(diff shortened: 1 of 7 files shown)


nikomatsakis:
(Whoops, PR title was totally wrong; I guess I forgot to rename my branch)

nikomatsakis:
@scalexm not yet

PR #244, opened 2019-09-23
extend `impls_for_trait` callback to take `[Parameter]` values

nikomatsakis:
This gives access to hints about *which* impls we are interested in.

cc #241 

r? @flodiebold


nikomatsakis:
This is mostly about the external interface, so I put r? @flodiebold -- but cc @scalexm =)

PR #251, opened 2019-10-01
simplify rust-ir

nikomatsakis:
Minor tweaks to simplify the chalk-rust-ir -- mostly removing things out from `Binders` that didn't belong there. 

(Context: I'm investigating how to simplify the `program_clauses` logic to aid in implementing [this idea](https://github.com/rust-lang/chalk/issues/234#issuecomment-530823193).)

### DIFF
--- a/src/lowering.rs
+++ b/src/lowering.rs
@@ -780,6 +780,35 @@ impl LowerPolarity for Polarity {
     }
 }
 
+trait LowerImplType {
+    fn lower(&self) -> rust_ir::ImplType;
+}
+
+impl LowerImplType for ImplType {
+    fn lower(&self) -> rust_ir::ImplType {
+        match self {
+            ImplType::Local => rust_ir::ImplType::Local,
+            ImplType::External => rust_ir::ImplType::External,
+        }
+    }
+}
+
+trait LowerTraitFlags {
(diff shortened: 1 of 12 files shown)


src/lowering.rs
```diff
@@ -963,14 +960,17 @@ impl LowerImpl for Impl {
...
-                    ImplType::External => rust_ir::ImplType::External,
-                },
             })
         })?;
 
-        Ok(rust_ir::ImplDatum { binders: binders })
+        Ok(rust_ir::ImplDatum {
+            polarity,
+            binders: binders,
+            impl_type: match self.impl_type {
```
tmandry: nit: should probably add a `lower()` method for this, too

src/lowering.rs
```diff
@@ -1080,18 +1080,20 @@ impl LowerTrait for TraitDefn {
...
-                    non_enumerable: self.flags.non_enumerable,
-                },
                 associated_ty_ids,
             })
         })?;
 
-        Ok(rust_ir::TraitDatum { binders: binders })
+        Ok(rust_ir::TraitDatum {
+            binders: binders,
+            flags: rust_ir::TraitFlags {
```
tmandry: ditto

PR #262, opened 2019-10-14
Modularize tests

nikomatsakis:
The tests in src/test.rs are kind of out of control. This PR breaks them into some rough categories.


nikomatsakis:
I checked that the number of tests matches:

```
athena. rg '^test ' before | wc -l
197
athena. rg '^test ' after | wc -l
197
```

so hopefully I didn't forget any `mod foo` files.

PR #263, opened 2019-10-19
Output multiple solutions

olegshtch:
I've tried to use chalk as generic logic solver and found I can not get output if there more than one solution. 

This PR adds `--multiple` flag to chalki. If it enabled chalki iterates over multiple solutions with confirmation from user instead of generating "ambiguous" solution. If goal was provided with arguments it iterates over all solutions.

It uses callback function to manage iteration over solution because I couldn't export `ForestSolver` iterator structure.

### DIFF
--- a/chalk-solve/src/solve.rs
+++ b/chalk-solve/src/solve.rs
@@ -143,6 +143,9 @@ impl Solver {
     /// - `goal` the goal to solve
     /// - `f` -- function to proceed solution. New solutions will be generated
     /// while function returns `true`.
+    ///   - first argument is solution found
+    ///   - second argument is ther next solution present
+    ///   - returns true if next solution should be handled
     ///
     /// # Returns
     ///
--- a/chalk-solve/src/solve.rs
+++ b/chalk-solve/src/solve.rs
@@ -129,6 +129,30 @@ impl Solver {
         self.forest.solve(&ops, goal)
     }
 
+    /// Attempts to solve the given goal, which must be in canonical
+    /// form. Provides multiple solutions to function `f`.  This will do
+    /// only as much work towards `goal` as it has to (and that work
+    /// is cached for future attempts).
+    ///
+    /// # Parameters
+    ///
+    /// - `program` -- defines the program clauses in scope.
+    ///   - **Important:** You must supply the same set of program clauses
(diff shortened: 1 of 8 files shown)


chalk-solve/src/solve.rs
```diff
@@ -128,6 +128,35 @@ impl Solver {
...
+    /// is cached for future attempts).
+    ///
+    /// # Parameters
+    ///
+    /// - `program` -- defines the program clauses in scope.
+    ///   - **Important:** You must supply the same set of program clauses
+    ///     each time you invoke `solve`, as otherwise the cached data may be
+    ///     invalid.
+    /// - `goal` the goal to solve
+    /// - `f` -- function to proceed solution. New solutions will be generated
```
nikomatsakis: It's not obvious what the parameters to `f` represent, can you comment them here?

chalk-solve/src/solve/slg/unique.rs
```diff
@@ -0,0 +1,17 @@
+use crate::solve::slg::SlgContext;
+use crate::solve::slg::SlgContextOps;
+use chalk_ir::*;
+
+use chalk_engine::context;
+use chalk_engine::SimplifiedAnswer;
+
+/// Takes one answer and returns it as unique solution
+impl<'me> context::UniqueOps<SlgContext> for SlgContextOps<'me> {
+    fn make_unique_solution(
```
nikomatsakis: I don't really understand this name. It doesn't seem like it's making a "unique" solution -- it seems like it's just extracting the canonical substitution from the answer? Maybe we should just call it `constrained_subst_from_answer`  or something?

I think I would also just add it to `ContextOps` instead of creating a new trait.

olegshtch: I've moved and renamed this function.

chalk-solve/src/solve.rs
olegshtch: I've added comment about parameters and return value.

olegshtch:
@nikomatsakis Maybe it better to have two different syntax for testing? The frst for all answers and the second for the first some answers like `yields*` and `yields+` or `yields_all` and `yields_first`.

Issue #255, opened 2019-10-04
simplify chalk context trait and logic-engine


nikomatsakis:
When integrating chalk into rustc, we had to make a number of changes to account for the `'gcx` / `'tcx` lifetime split, but that split no longer exists. 

Some of the things we can simplify:

We no longer need to canonicalize when "storing" strands into the table, so we should be able to get rid of the `CanonicalStrand` type:

https://github.com/rust-lang/chalk/blob/13303bb0067c6ed0572322080ae367ee38f9e7c9/chalk-engine/src/strand.rs#L6-L12

Similarly, I don't think we need the whole "with_instantiated_foo" workarounds here:

https://github.com/rust-lang/chalk/blob/13303bb0067c6ed0572322080ae367ee38f9e7c9/chalk-engine/src/context.rs#L179-L202

It's going to take a bit of playing around to reconstruct the simplified structure though.

jackh726:
@nikomatsakis I've been playing with this over the weekend and have made a lot of progress. Still have some more work to do, but should have a PR sometime soon.

jackh726:
I'm gonna go ahead and close this. There *might* be room for simplification here. But the code has diverged so much from what it was when these changes were made, it's not clear if and what it is.

Issue #250, opened 2019-09-30
remove support for "negative cycles"


nikomatsakis:
Chalk today correctly handles "negative cycles", where you have something like

```
A :- not { B }.
B :- not { A }.
```

in the underlying logic. This requires a fairly complex mechanism called *delayed literals*. It is also, I believe, unnecessary for Rust -- and that should probably be a continued design goal. 

The reason to remove this logic is two-fold:

- It is quite complicated. It requires us to "delay" goals and then find a good time to "undelay" them, and it is hard to know when the right time for that is.
- It complicates fixing rust-lang/chalk#248, which probably requires a similar mechanism.

The only reason that we might need this would be specialization, which uses `not { }` reasoning in a more general way that could in principle recurse. However, we are guaranteed that it doesn't because we require an acyclic relationship between traits (at least at present), such that if impl A (for trait T) overrides another impl, the where clauses in A must not require solving anything from the trait T. So you should not get any sort of cycle.

nikomatsakis:
I talked to @tmandry and they're going to take a swing at this

nikomatsakis:
Or not! Looks like @jackh726 is on it ;)

jackh726:
The timing!

PR #264, opened 2019-10-20
Simplify chalk-engine a bit

jackh726:
cc https://github.com/rust-lang/chalk/issues/255

This removes the two different `Context`s from `Strand` and the `&'table mut dyn InferenceTable<C, I>` field (and consequently the lifetime parameter `'table`) has been removed, in lieu of passing `infer` directly to function arguments as needed. `WithInstantiatedUCanonicalGoal` and `WithInstantiatedExClause` were also removed in favor of just `FnOnce`s. And then the `with_instantiated_strand` indirection has also been removed.

`CanonicalStrand`s are still stored in `Table`, since I haven't yet figured out the best way to simplify that. But this gets us one step closer.

I'll probably still be playing around to try to remove `CanonicalStrand`, but I figured it would be better to at least make this PR since it seems fairly straightforward.

### DIFF
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -280,11 +280,6 @@ pub trait UnificationOps<C: Context> {
         constraints: Vec<C::RegionConstraint>,
     ) -> C::CanonicalConstrainedSubst;
 
-    fn sink_answer_subset(
-        &self,
-        value: &C::CanonicalConstrainedSubst,
-    ) -> C::CanonicalConstrainedSubst;
-
     fn lift_delayed_literal(&self, value: DelayedLiteral<C>) -> DelayedLiteral<C>;
 
     // Used by: logic
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -118,10 +118,6 @@ pub trait Context: Clone + Debug {
     /// the equality relation.
     type Variance;
 
-    /// The successful result from unification: contains new subgoals
-    /// and things that can be attached to an ex-clause.
-    type UnificationResult;
-
     /// Given an environment and a goal, glue them together to create
     /// a `GoalInEnvironment`.
     fn goal_in_environment(
@@ -294,18 +290,20 @@ pub trait UnificationOps<C: Context> {
     // Used by: logic
     fn invert_goal(&mut self, value: &C::GoalInEnvironment) -> Option<C::GoalInEnvironment>;
 
+    /// First unify the parameters, then add the residual subgoals
+    /// as new subgoals of the ex-clause.
+    /// Also add region constraints.
+    /// 
+    /// If the parameters fail to unify, then `Error` is returned
     // Used by: simplify
-    fn unify_parameters(
+    fn unify_parameters_into_ex_clause(
         &mut self,
         environment: &C::Environment,
         variance: C::Variance,
         a: &C::Parameter,
         b: &C::Parameter,
-    ) -> Fallible<C::UnificationResult>;
-
-    /// Add the residual subgoals as new subgoals of the ex-clause.
-    /// Also add region constraints.
-    fn into_ex_clause(&mut self, result: C::UnificationResult, ex_clause: &mut ExClause<C>);
+        ex_clause: &mut ExClause<C>,
+    ) -> Fallible<()>;
 }
 
 /// "Truncation" (called "abstraction" in the papers referenced below)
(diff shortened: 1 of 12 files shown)


nikomatsakis:
Yep, it looks like this is the commit to look at

5d8bc3360a7bd4b2ae14f2b386ca744cd859ab0c

jackh726:
@nikomatsakis Yes, I went through the history and found that. The problem is now we _can't_ store the `InferenceTable` in the Strand, since it now has a lifetime associated with it (from `&'me dyn RustIrDatabase`).

nikomatsakis:
@jackh726 I see. Interesting! Well, I think we can land this PR for now and look into this question as follow-up. The only thing I'm not sure about is whether it makes sense to remove the inference table from `Strand`, as I think we might likely want to put it back again.

nikomatsakis:
I suspect that the right answer here is going to be:

* move `program_clauses` from the `InferenceTable` trait to `ContextOps`
* then we can remove the field `program` from the `TruncatingInferenceTable`
* then we can store it in `Strand`

jackh726:
@nikomatsakis I can definitely try that later tonight. I can also start topic on Zulip if you want to discuss more in sync.

nikomatsakis:
@jackh726 zulip topic sounds good =)

chalk-engine/src/context.rs
```diff
@@ -401,3 +374,9 @@ pub trait AnswerStream<C: Context> {
     fn any_future_answer(&mut self, test: impl FnMut(&C::InferenceNormalizedSubst) -> bool)
         -> bool;
 }
+
+pub trait Canonical<T, C: Context> {
```
jackh726: I was looking at some of the git history since writing this, and it seems like there was a push to reduce the traits in `Context`, so I'm not sure this trait makes since.

chalk-engine/src/context.rs
```diff
@@ -258,20 +239,20 @@ pub trait InferenceTable<C: Context, I: Context>:
...
-    fn into_hh_goal(&mut self, goal: I::Goal) -> HhGoal<I>;
+    fn into_hh_goal(&mut self, goal: C::Goal) -> HhGoal<C>;
 
     // Used by: simplify
-    fn add_clauses(&mut self, env: &I::Environment, clauses: I::ProgramClauses) -> I::Environment;
+    fn add_clauses(&mut self, env: &C::Environment, clauses: C::ProgramClauses) -> C::Environment;
 
     /// Upcast this domain goal into a more general goal.
-    fn into_goal(&self, domain_goal: I::DomainGoal) -> I::Goal;
+    fn into_goal(&self, domain_goal: C::DomainGoal) -> C::Goal;
```
nikomatsakis: An aside: I think it may be worth creating a first PR that **just** removes the `I: Context` type parameters and collapses to one `C: Context` everywhere, but leaves everything else the same. It'd be good to pull out and land intermediate steps.

nikomatsakis: Well, we could probably get rid of the `WithFoo` traits and convert them to `impl FnOnce`, too, but leave everything else about them the same.

chalk-engine/src/context.rs
nikomatsakis: Yeah, not sure, I did push away from these sorts of traits at some point. I think it was because, for real flexibility, you wind up wanting to thread `&C` arguments around anyway, since some crates may need access to fields.

nikomatsakis: Oh and I think it was because of coherence reasons, too -- this way, you only needed to own the "context" type, but the other types might be from other crates

nikomatsakis: I think I'd prefer to have the methods collected on the `Context` type, but we could introduce "indirection helpers" if desired to make the code read more nicely. Still, I'm not convinced these methods are needed at all.

chalk-engine/src/context.rs
nikomatsakis: I guess this was fairly close to your original PR :)

chalk-engine/src/logic.rs
```diff
@@ -440,22 +367,9 @@ impl<C: Context> Forest<C> {
...
-                context,
-                num_universes,
-                canonical_strand,
-                DelayStrandAfterCycle { table },
-            );
-
-            *canonical_strand = delayed_strand;
-
+        for strand in self.tables[table].strands_mut() {
+            // FIXME: do we need to uncanonicalize this
```
nikomatsakis: I don't think so -- but I have to review this logic -- in any case I actually want to remove this "delaying" stuff, see #250, so maybe that too would be a good step to take before doing the rest of this PR, just to simplify things?

jackh726:
@nikomatsakis Ok, so I removed the "remove `CanonicalStrand`" commit, and added a couple other smaller cleanup changes. I do think that any followup work for removing `CanonicalStrand` should be in another PR, possibly after removing the delaying stuff.

Let me know if you disagree with the other smaller changes I made and I can just remove them.

PR #265, opened 2019-10-21
introduce `TypeFamily` and make chalk-ir generic over it

nikomatsakis:
This is a largely mechanical refactoring that introduces the `TypeFamily` trait and makes chalk-ir generic over it. It doesn't (yet) do anything interesting with the fact that chalk-ir is generic.

r? @tmandry

### DIFF
--- a/chalk-solve/src/infer/var.rs
+++ b/chalk-solve/src/infer/var.rs
@@ -48,7 +48,7 @@ impl EnaVariable {
     /// method, naturally you should know from context that the kind
     /// of this inference variable is a type (we can't check it).
     pub(crate) fn to_ty(self) -> Ty<ChalkIr> {
-        self.0.to_ty()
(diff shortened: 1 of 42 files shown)


chalk-solve/src/infer/var.rs
```diff
@@ -46,15 +47,15 @@ impl EnaVariable {
     /// Convert this inference variable into a type. When using this
     /// method, naturally you should know from context that the kind
     /// of this inference variable is a type (we can't check it).
-    pub(crate) fn to_ty(self) -> Ty {
-        self.0.to_ty()
+    pub(crate) fn to_ty(self) -> Ty<ChalkIr> {
+        self.0.to_ty::<ChalkIr>()
```
tmandry: hopefully this is inferred?

tmandry: (feel free to leave these as-is, though)

PR #267, opened 2019-10-24
Fold derive macro

detrumi:


Cargo.toml
```diff
@@ -48,4 +48,8 @@ path = "chalk-engine"
 version = "0.1.0"
 path = "chalk-rust-ir"
 
+[dependencies.fold-derive]
```
bjorn3: Maybe name it `chalk-macros`?

fold-derive/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }
+            None
+        }),
+        _ => None,
+    }
+}
+
+#[proc_macro_attribute]
+pub fn fold_family(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
bjorn3: This is not necessary, as you already declared it while declaring the `Fold` proc macro derive.

Cargo.toml
detrumi: There's already a `chalk-macros` crate, and we can't add these to that crate:
```
`proc-macro` crate types cannot export any items other than functions tagged with `#[proc_macro_derive]` currently
```
Maybe `chalk-derive` instead?

chalk-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::{format_ident, quote, ToTokens};
+use syn::{parse_macro_input, Data, DeriveInput, GenericParam, Ident, TypeParamBound};
+
+#[proc_macro_derive(Fold, attributes(has_type_family))]
+pub fn derive_fold(item: TokenStream) -> TokenStream {
```
nikomatsakis: I would like to see some documentation on the conventions being observed here (i.e., the fact that one should have a `TF: TypeFamily` attribute or whatever).

chalk-ir/src/lib.rs
```diff
@@ -570,7 +573,7 @@ impl<TF: TypeFamily> Parameter<TF> {
     }
 }
 
-#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Fold)]
```
nikomatsakis: this is definitely nice =)

chalk-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        Data::Union(..) => panic!("Fold can not be derived for unions"),
+    }
+}
+
+fn is_type_family_param(param: &GenericParam) -> Option<&Ident> {
+    match param {
+        GenericParam::Type(ref t) => t.bounds.iter().find_map(|b| {
+            if let TypeParamBound::Trait(trait_bound) = b {
+                if trait_bound.path.get_ident().map(|i| i.to_string())
```
nikomatsakis: is this always the final identifier?

chalk-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+            .expect("Expected has_type_family argument");
+        quote! { < #arg > } // TODO extend instead of overriding?
+    } else {
+        ty_generics.to_token_stream()
+    };
+
+    let mut impl_generics = input.generics.clone();
+    let mut where_clause = where_clause_ref.cloned();
+
+    // if any impl_generics G : TypeFamily:
```
nikomatsakis:
```suggestion
    // if any impl_generics G : HasTypeFamily:
```

I think this comment is wrong?

nikomatsakis: I'd probably extend it to say something like:

---

Look for the case where the type has a type parameter `G: HasTypeFamily`. In that case, we want to get the type family from `<G as HasTypeFamily>::TypeFamily`, instead of looking for a type parameter named `TF`.

chalk-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+    if let Some(param) = input
+        .generics
+        .params
+        .iter()
+        .find_map(|p| is_type_family_param(p))
+    {
+        impl_generics.params.push(GenericParam::Type(
+            syn::parse(quote! { TF: TypeFamily }.into()).unwrap(),
+        ));
+        generics = quote! { <TF> };
```
nikomatsakis: I think this should be

```rust
quote! { <<#param as HasTypeFamily>::TypeFamily> }
```

nikomatsakis: in that case, we don't need to introduce a new `TF` type parameter or anything

chalk-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+        .attrs
+        .iter()
+        .find(|a| a.path.is_ident("has_type_family"))
+    {
+        let arg = attr
+            .parse_args::<proc_macro2::TokenStream>()
+            .expect("Expected has_type_family argument");
+        quote! { < #arg > } // TODO extend instead of overriding?
+    } else {
+        ty_generics.to_token_stream()
```
nikomatsakis: OK so we assume that one of the following is true:

* there is a `#[has_type_family]` attribute
* there is a parameter `G: HasTypeFamily`
* there is a single generic parameter and it is a type family

Right?

nikomatsakis:
I'd also like to consider some changes to the `Fold` trait and I'm debating if we should just land this now and keep iterating :)

chalk-derive/src/lib.rs
detrumi: Good call, that also removed half of the where-clause.

chalk-derive/src/lib.rs
detrumi: The third option should also work when there's more parameters, I think

chalk-derive/src/lib.rs
detrumi: Oh right, I guess someone could fully qualify the path

detrumi:
> Nice! I left some comments. I think I still vaguely prefer to see us
> 
>     * add some procedural macro to generate `HasTypeFamily` implementations
> 
>     * base the `Fold` derive on those
> 
> 
> but on the other hand I would be happy to land this version (with some comments) and tinker with it, since this also seems reasonable.

I'm not sure the split would be that clean, but it's worth looking into.

PR #268, opened 2019-10-27
small fixes

guanqun:


chalk-macros/src/lib.rs
```diff
@@ -31,7 +31,7 @@ thread_local! {
 
 // When CHALK_DEBUG is enabled, we only allow this many frames of
 // nested processing, at which point we assume something has gone
-// awry.
```
shepmaster: For what it's worth, "awry" is a valid English word, but it's probably worth changing it to something more understandable if it is causing confusion.

guanqun: need to have a dictionary at my hand.

guanqun: thanks for the info.

bjorn3: away is not the right word.

PR #269, opened 2019-10-27
Add TypeName::Error variant

flodiebold:
This can be used (e.g. by rust-analyzer) as a placeholder for types where there
was an error e.g. during name resolution.

(I also considered making it a variant of `Ty`, but this seemed simpler.)


nikomatsakis:
Let's try it and see how it goes.

PR #270, opened 2019-10-28
Remove delayed literals

### LINKED ISSUE #79: On-demand solver does not support simplification of delayed literals
One thing that the on-demand solver does not support yet is **simplification** of delayed literals. In the case of negative loops we sometimes have to delay processing a given literal, and simplification is what comes back in the end to clean up the mess.

It's not clear how imp't this is: it an only arise with negative cycles, which are not yet possible in Rust since negative clauses never appear in goals, and may never be possible (the fact that we would have to deal with negative cycles is a good argument against adding such goals).

My strategy for doing it was going to be sort of crude: when we encounter an answer with a delayed literal (other than CannotProve, which can never be simplified), we have to first "force complete" the table. This means that we run until all strands are done, and then we further go over all the answers, find all the delayed literals we may need to simplify, and "force complete" their tables (this process may recurse, of course). Once this is done, we can run the simplification procedure (here is the [simplification code from the eager SLG solver](https://github.com/rust-lang-nursery/chalk/blob/b0d1417104ccdc52a72b96d2fe89e56a1e284e13/src/solve/slg/mod.rs#L1566-L1615), for reference, which includes a writeup). 

This is roughly where we would want to make this change:

https://github.com/rust-lang-nursery/chalk/blob/b0d1417104ccdc52a72b96d2fe89e56a1e284e13/src/solve/slg/on_demand/forest.rs#L136-L143

### LINKED ISSUE #250: remove support for "negative cycles"

jackh726:
Currently based on #264, but I can change that to master instead if this should land first.

Closes #79
Closes #250

### DIFF
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -45,6 +45,10 @@ impl<C: Context> Forest<C> {
     ///
     /// Thanks to subgoal abstraction and so forth, this should always
     /// terminate.
+    ///
+    /// # Panics
+    ///
+    /// Panics if a negative cycle was detected.
     pub fn force_answers(
         &mut self,
         context: &impl ContextOps<C>,
@@ -177,6 +181,9 @@ struct ForestSolver<'me, C: Context, CO: ContextOps<C>> {
 }
 
 impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C, CO> {
+    /// # Panics
+    ///
+    /// Panics if a negative cycle was detected.
     fn peek_answer(&mut self) -> Option<Answer<C>> {
         loop {
             match self
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -61,6 +61,9 @@ impl<C: Context> Forest<C> {
                     Err(RootSearchFail::Floundered) => return None,
                     Err(RootSearchFail::QuantumExceeded) => continue,
                     Err(RootSearchFail::NoMoreSolutions) => return Some(answers),
+                    Err(RootSearchFail::NegativeCycle) => {
+                        panic!("negative cycle was detected");
+                    }
                 }
             }
 
@@ -198,6 +201,10 @@ impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C
                 }
 
                 Err(RootSearchFail::QuantumExceeded) => {}
+
+                Err(RootSearchFail::NegativeCycle) => {
+                    panic!("negative cycle was detected");
+                }
             }
         }
     }
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -2,9 +2,10 @@ use crate::context::prelude::*;
 use crate::context::AnswerStream;
 use crate::logic::RootSearchFail;
 use crate::stack::{Stack, StackIndex};
-use crate::table::{Answer, AnswerIndex};
+use crate::table::AnswerIndex;
 use crate::tables::Tables;
-use crate::{DepthFirstNumber, SimplifiedAnswer, TableIndex};
+use crate::Answer;
+use crate::{DepthFirstNumber, TableIndex};
 
 pub struct Forest<C: Context> {
     context: C,
@@ -173,7 +174,7 @@ struct ForestSolver<'me, C: Context, CO: ContextOps<C>> {
 }
 
 impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C, CO> {
-    fn peek_answer(&mut self) -> Option<SimplifiedAnswer<C>> {
+    fn peek_answer(&mut self) -> Option<Answer<C>> {
         loop {
             match self
                 .forest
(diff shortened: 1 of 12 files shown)


nikomatsakis:
@jackh726 #264 is merged, maybe rebase?

jackh726:
@nikomatsakis done!

jackh726:
@nikomatsakis rebased again to current master

chalk-engine/src/logic.rs
```diff
@@ -352,7 +352,7 @@ impl<C: Context> Forest<C> {
             let mut visited = FxHashSet::default();
```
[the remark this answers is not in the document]
jackh726: That was actually the first thing I did! Panicing here causes a lot of tests to fail.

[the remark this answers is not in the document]
jackh726: Of course, there may be some things to do upstream to this that might prevent this path from being taken, but as the code is currently, it is taken quite a bit.

chalk-engine/src/forest.rs
```diff
@@ -213,11 +201,15 @@ impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C
                 }
 
                 Err(RootSearchFail::QuantumExceeded) => {}
+
+                Err(RootSearchFail::NegativeCycle) => {
+                    panic!("negative cycle was detected");
```
nikomatsakis:
```suggestion
                    // Negative cycles *ought* to be avoided by construction. Hence panic
                    // if we find one, as that likely indicates a problem in the chalk-solve
                    // lowering rules. (In principle, we could propagate this error out,
                    // and let chalk-solve do the asserting, but that seemed like it would
                    // complicate the function signature more than it's worth.)
                    panic!("negative cycle was detected");
```

chalk-engine/src/lib.rs
```diff
@@ -112,9 +111,9 @@ pub struct ExClause<C: Context> {
     /// would yield A.
     pub subst: C::Substitution,
 
-    /// Delayed literals: things that we depend on negatively,
-    /// but which have not yet been fully evaluated.
-    pub delayed_literals: Vec<DelayedLiteral<C>>,
+    /// True if any subgoals were depended upon negatively and
+    /// were not fully evaluated. This replaces delayed literals.
```
nikomatsakis:
```suggestion
    /// were not fully evaluated, or if we encountered a `CannotProve` 
    /// goal. (In the full SLG algorithm, we would use delayed literals here, 
    /// but we don't bother, as we don't need that support.)
```

src/test/slg.rs
```diff
@@ -304,6 +304,7 @@ fn only_draw_so_many_blow_up() {
 
 /// Here, P and Q depend on one another through a negative loop.
 #[test]
+#[should_panic]
```
nikomatsakis:
```suggestion
#[should_panic(expected = "negative cycle")]
```

src/test/slg.rs
```diff
@@ -593,14 +580,15 @@ fn example_2_2_EWFS() {
...
-                    delayed_literals: {}
+                    ambiguous: false
                 }
             ]"
         }
     }
 }
 
 #[test]
+#[should_panic]
```
nikomatsakis:
```suggestion
#[should_panic(expected = "negative cycle")]
```

src/test/slg.rs
```diff
@@ -623,23 +611,14 @@ fn example_2_3_EWFS() {
...
-                }
-            ]"
+            // Negative cycle -> panic
+            r""
         }
     }
 }
 
 #[test]
+#[should_panic]
```
nikomatsakis:
```suggestion
#[should_panic(expected = "negative cycle")]
```

src/test/slg.rs
```diff
@@ -658,34 +637,16 @@ fn example_3_3_EWFS() {
...
+            // Negative cycle -> panic
+            r""
         }
     }
 }
 
 /// Here, P is neither true nor false. If it were true, then it would
 /// be false, and so forth.
 #[test]
+#[should_panic]
```
nikomatsakis:
```suggestion
#[should_panic(expected = "negative cycle")]
```

src/test/slg.rs
```diff
@@ -962,18 +897,17 @@ fn cached_answers_3() {
...
         }
     }
 }
 
 /// Here, P depends on Q negatively, but Q depends only on itself.
-/// What happens is that P adds a negative link on Q, so that when Q
-/// delays, P is also delayed.
 #[test]
-fn negative_answer_delayed_literal() {
+#[should_panic]
```
nikomatsakis:
```suggestion
#[should_panic(expected = "negative cycle")]
```

Issue #248, opened 2019-09-30
coinduction impl fails to consider that recursive calls may later fail


nikomatsakis:
Today's handling of coinduction in the logic engine is unsound. We consider any purely coinductive loop to be true, but actually a coinductive loop could fail later on.

Consider something like this, where each C*i* is a coinductive goal:

```
C1 :- C2, C3.
C2 :- C1.
```

Here, C1 is clearly false, because there is no rules for C3. Therefore, C2 must also be false. However, under certain conditions our current code will incorrect consider that `C2` holds.

nikomatsakis:
This is where the bug in our logic lies:

https://github.com/rust-lang/chalk/blob/457b473bf34a889fce709c44ac870125f018775c/chalk-engine/src/logic.rs#L1061-L1086

We assume that a recursive, coinductive goal is successful -- but in fact we should make the result *conditional* on that recursive, coinductive goal. This is effectively a form of delayed goal.

nikomatsakis:
My plan for the fix, in short:

* If you see a co-inductive cycle, we create an answer with a delayed literal
    * So in our example, proving C2 would result in an answer "Yes, if C1"
* This delayed literal would propagate forward, so C1 would wind up with answer "Yes, if C1"
    * At this point, we can drop the delayed literal 
* Meanwhile if we start another query (say of C2) and we get the answer "Yes, if C1", we would have to also do the query C1 to see if the result is yes or no

I've not worked out all the details here to figure out how easy/hard this will be -- but it's a simpler form of delayed literals than what we used to have, I think.

One concern I have is that the resulting answer will depend on the table you start with.

jackh726:
So, I've been looking through the code for this, and the reason the added tests are failing are because C1, C2, and C3 are auto traits. And there isn't a mechanism at all to say "yes, C1 is an auto trait, but in this case it isn't"

jackh726:
So, I think the rules for how auto traits behave with explicit impls here need to be better defined (or at least defined for me, since I don't know them). 

What is surprising to me is that the second test's output doesn't match the first one. So there's more to this for sure.

jackh726:
Ok @nikomatsakis 

The good news is I have a fix. However, it's quite a bit overengineered and I think most of what I changed is unnecessary. So I'll spend time tomorrow un-overengineering it.

PR #272, opened 2019-10-30
Fix coinductive unsoundness

jackh726:
Fixes https://github.com/rust-lang/chalk/issues/248

Now, when a coinductive cycle is found, it is marked in `PositiveCycle` as such. In `pursue_strand`, these are ignored until all other subgoals have been proven. Once that is the case, we can assume that the coinductive cycles are true.

With this, I also did a lot of refactoring in terms of how the recursion is handled. Importantly, I moved the calls `ensure_answer_recursively` and `pursue_strand_recursively` out of `pursue_positive_subgoal` and `pursue_negative_subgoal`, which ends up making it really obvious that `purse_strand` can easily be done using a loop instead of recursion. So now, the stack protection feature is no longer needed.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -16,10 +16,10 @@ type RootSearchResult<T> = Result<T, RootSearchFail>;
 /// empty stack.
 #[derive(Debug)]
 pub(super) enum RootSearchFail {
-    /// The subgoal we were trying to solve cannot succeed.
+    /// The table we were trying to solve cannot succeed.
     NoMoreSolutions,
 
-    /// The subgoal cannot be solved without more type information.
+    /// The table cannot be solved without more type information.
     Floundered,
 
     /// We did not find a solution, but we still have things to try.
@@ -32,6 +32,10 @@ pub(super) enum RootSearchFail {
     /// A negative cycle was found. This is fail-fast, so even if there was
     /// possibly a solution (ambiguous or not), it may not have been found.
     NegativeCycle,
+
+    /// The current answer index is not useful. Currently, this is returned
+    /// because the current answer needs refining.
+    InvalidAnswer,
 }
 
 /// This is returned when we try to select a subgoal for a strand.
@@ -78,23 +82,14 @@ impl<C: Context> Forest<C> {
             Ok(()) => {
                 let answer = self.answer(table, answer_index);
                 let has_delayed_subgoals = C::has_delayed_subgoals(&answer.subst);
+                if has_delayed_subgoals {
+                    return Err(RootSearchFail::InvalidAnswer);
+                }
                 Ok(CompleteAnswer {
                     subst: C::canonical_constrained_subst_from_canonical_constrained_answer(
                         &answer.subst,
                     ),
-
-                    // We consider "delayed subgoals" currently to be
-                    // ambiguous, since -- after all -- we don't know
-                    // if they're true or not. But this is not as
-                    // precise as it could be, we should probably
-                    // consider them to be a *failure* actually. The
-                    // reason is that we will have created another
-                    // strand that is attempting to prove them
-                    // true. If that strand *succeeds*, we'll have
-                    // another answer. If it fails, we won't.  (And if
-                    // *it* encounters ambiguity, we'll have an
-                    // ambiguous answer.)
-                    ambiguous: answer.ambiguous || has_delayed_subgoals,
+                    ambiguous: answer.ambiguous,
                 })
             }
             Err(err) => Err(err),
@@ -616,9 +611,12 @@ impl<C: Context> Forest<C> {
                         // we might have to build on it (see the
                         // Delayed Trivial Self Cycle, Variant 3
                         // example).
+                        let (_, _, _, table_goal) =
+                            context.instantiate_ucanonical_goal(&self.tables[table].table_goal);
 
                         let answer = self.answer(table, answer_index);
-                        if let Some(strand) = self.create_refinement_strand(context, table, answer)
+                        if let Some(strand) =
+                            self.create_refinement_strand(context, table, answer, table_goal)
                         {
                             self.tables[table].enqueue_strand(strand);
                         }
@@ -668,6 +666,7 @@ impl<C: Context> Forest<C> {
         context: &impl ContextOps<C>,
         table: TableIndex,
         answer: &Answer<C>,
+        table_goal: C::Goal,
     ) -> Option<CanonicalStrand<C>> {
         // If there are no delayed subgoals, then there is no need for
         // a refinement strand.
@@ -679,16 +678,29 @@ impl<C: Context> Forest<C> {
         let (table, subst, constraints, delayed_subgoals) =
             context.instantiate_answer_subst(num_universes, &answer.subst);
 
+        // FIXME: really, these shouldn't be added to the delayed_subgoals at all
+        // however, because we can't compare the table goal until `canonicalize_answer_subst`
+        // is called in `pursue_answer`, this will require a bit of refactoring work
+        let filtered_delayed_subgoals = delayed_subgoals
+            .into_iter()
+            .filter(|delayed_subgoal| {
+                dbg!(
+                    &C::goal_from_goal_in_environment(delayed_subgoal),
+                    &table_goal,
+                    *C::goal_from_goal_in_environment(delayed_subgoal) != table_goal
+                );
+                *C::goal_from_goal_in_environment(delayed_subgoal) != table_goal
+            })
+            .map(Literal::Positive)
+            .collect();
+
         let strand = Strand {
             infer: table,
             ex_clause: ExClause {
                 subst,
                 ambiguous: answer.ambiguous,
                 constraints,
-                subgoals: delayed_subgoals
-                    .into_iter()
-                    .map(Literal::Positive)
-                    .collect(),
+                subgoals: filtered_delayed_subgoals,
                 delayed_subgoals: Vec::new(),
                 answer_time: TimeStamp::default(),
                 floundered_subgoals: Vec::new(),
@@ -1205,9 +1217,8 @@ impl<C: Context> Forest<C> {
     fn push_initial_strands(&mut self, context: &impl ContextOps<C>, table: TableIndex) {
         // Instantiate the table goal with fresh inference variables.
         let table_goal = self.tables[table].table_goal.clone();
-        context.instantiate_ucanonical_goal(&table_goal, |infer, subst, environment, goal| {
-            self.push_initial_strands_instantiated(context, table, infer, subst, environment, goal);
-        });
+        let (infer, subst, environment, goal) = context.instantiate_ucanonical_goal(&table_goal);
+        self.push_initial_strands_instantiated(context, table, infer, subst, environment, goal);
     }
 
     fn push_initial_strands_instantiated(
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -210,7 +210,7 @@ impl<C: Context> Forest<C> {
                             continue;
                         }
                         SubGoalSelection::NoRemainingSubgoals => {
-                            match self.on_no_remaining_subgoals(strand) {
+                            match self.on_no_remaining_subgoals(context, strand) {
                                 NoRemainingSubgoalsResult::RootAnswerAvailable => return Ok(()),
                                 NoRemainingSubgoalsResult::RootSearchFail(e) => return Err(e),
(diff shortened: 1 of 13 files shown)


chalk-engine/src/logic.rs
```diff
@@ -913,22 +907,80 @@ impl<C: Context> Forest<C> {
     /// from the NFTD paper.
     fn pursue_positive_subgoal(
```
jackh726: So, these two functions (this and `pursue_negative_subgoal` are really "pursuing" the strand, so a different name might be better. Ideas?

chalk-engine/src/logic.rs
```diff
@@ -1245,53 +1146,102 @@ impl<C: Context> Forest<C> {
                     "pursue_negative_subgoal: found neg cycle at depth {:?}",
                     min
                 );
-                let canonical_strand = Self::canonicalize_strand(strand);
-                return Err(StrandFail::PositiveCycle(
+                let canonical_strand =
+                    Self::canonicalize_strand_from(&mut infer, &ex_clause, selected_subgoal);
+                // FIXME: this should evaluate if this is a negative cycle
```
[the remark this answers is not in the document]
jackh726: To be honest, I'm not sure. I saw at one point that a negative cycle was getting stored into this and not resolved until higher up, and I don't think I did anything to *fix* that. But I also didn't investigate this. I left a comment here to get back to it later but never did. I'll probably just remove this.

chalk-engine/src/logic.rs
```diff
@@ -412,67 +378,90 @@ impl<C: Context> Forest<C> {
...
+                            return Err(StrandFail::Floundered);
+                        }
+                        continue;
+                    }
+                    if !strand.ex_clause.delayed_coinductive_cycles.is_empty() {
+                        strand
+                            .ex_clause
+                            .subgoals
+                            .extend(strand.ex_clause.delayed_coinductive_cycles.drain(..));
+                        allow_coinduction = true;
```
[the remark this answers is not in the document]
jackh726: Hmm, I have to think about if that is possible. I don't think so, because the `PositiveCycle` is returned before any chance to make a new subgoal is reached.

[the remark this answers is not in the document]
nikomatsakis: I also suspect not, but I wanted to trace out the flow and convince myself. :)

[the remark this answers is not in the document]
jackh726: FWIW, I'll add a comment to make sure this assertion explicit. (And also go through and verify).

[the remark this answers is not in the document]
jackh726: I also don't know what it would mean to *add* subgoals to a strand? Sure, we move subgoals to floundering and to delayed_coinductive, but they aren't "created" really after the initial `simplify_hh_goal`. Otherwise, if new subgoals can be added at any time, then that makes keeping track of whether or not a strand is "complete" much more difficult.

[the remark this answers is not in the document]
nikomatsakis: They can in fact be created by every unification. This is partly what I've been attempting to refactor away, but I don't think we're going to be able to do that completely.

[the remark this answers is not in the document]
nikomatsakis: Although I realize now that we actually *could*

jackh726:
I did make a new test that should work but instead panics. So this PR isn't ready.

```
#[test]
fn coinductive_multicycle() {
    test! {
        program {
            trait Any { }

            #[auto]
            trait C1 { }

            #[auto]
            trait C2 { }

            #[auto]
            trait C3 { }

            trait C4 { }

            forall<T> {
                T: C1 if T: C2
            }

            forall<T> {
                T: C2 if T: C3
            }

            forall<T> {
                T: C3 if T: C1
            }

            forall<T> {
                T: Any if T: C3
            }

            forall<T> {
                T: Any if T: C2
            }

            forall<T> {
                T: Any if T: C1
            }
        }

        goal {
            forall<X> { X: Any }
        } first 10 with max 3 {
            r"[]"
        }
    }
}
```

jackh726:
@nikomatsakis Ok I've updated so that the test (and a couple more) pass.

To summarize the updated changes: A strand is marked as `EnsureSuccess::Coinductive` if all remaining subgoals are coinductive (either by propogation up the stack, or from a direct cycle). Only when a `Coinductive` success reaches the root can it be turned into an answer. This means that all strands on the stack only had coinductive subgoals.

nikomatsakis:
@jackh726 I'll take a look. I had these two "inklings" I was pursuing with the old branch to try and turn into failing test cases. I can't quite tell if your test matches them.

Suspicion 1. The old code at least just checked whether something was coinductive, but it didn't look at how far up the stack the "coinduction" went. This seemed problematic. I suspect your recent changes address this.

Suspicion 2. When things recurse, it means that they match the same goal, but those goals can have unknown variables. I worry that we might have a recursion that only works for some values of those variables and it might not get detected.

nikomatsakis:
OK, I read through the latest version. Let me see if I can summarize what happens:

* When a (coinductive) cycle is encountered, the subgoal is recorded in the strand.
* This strand does not record a regular answer, but instead returns a `CoinductiveSuccess` with itself (canonicalized).
* If this is not the top of the stack, that strand will be stored back into its table by the caller. The caller will then mark its own goal as coinductive and keep going.
* If it *is* the root strand, and once all other subgoals are completed, we will consider it to be an answer.

I think this is going in the right direction but I have a few concerns:

First, I don't think it's correct to special case the root strand. We should be leveraging the depth here. In other words, the root of the coinductive cycle may not be the root of the overall *stack*.

Second, I think it would be better if the reasoning regarding strands was more local -- this logic feels very spread out to me and kind of harder to follow.

I left some notes on how I had imagined it working but I think I still need to try to make that more concrete. I will try to do this later today if I get a chance.

One thing I was wondering was whether there are smaller refactorings that we can pull out from this PR?

chalk-engine/src/logic.rs
```diff
@@ -115,7 +115,14 @@ impl<C: Context> Forest<C> {
         assert!(self.stack.is_empty());
 
         match self.ensure_answer_recursively(context, table, answer) {
-            Ok(EnsureSuccess::AnswerAvailable) => Ok(()),
+            Ok(EnsureSuccess::AnswerAvailable) => {
+                let answer = self.answer(table, answer);
+                for delayed_subgoal in answer.delayed_subgoals.clone() {
```
jackh726: I haven't written tests for  variant 2 or 3 yet, but no tests hit this. So I'm wondering what conditions this could happen.

chalk-engine/src/logic.rs
```diff
@@ -1006,14 +1028,33 @@ impl<C: Context> Forest<C> {
...
+
+                        // For of the delayed subgoals for the answer, we try to check
+                        // if this is a trivial self-cycle. If so, we don't need to add
+                        // this as a delayed_subgoal. Otherwise, we do.
+                        for delayed_subgoal in answer.delayed_subgoals.clone() {
+                            let table = self.stack[depth].table;
+                            let table_goal = &self.tables[table].table_goal;
+                            dbg!(&delayed_subgoal);
+                            dbg!(&table_goal);
+                            if delayed_subgoal == *table_goal {
```
jackh726: Right now, I'm only checking for equality. This is obviously not sufficient (see `cyclic_wf_requirements` test). What's the best way to do this?

chalk-ir/src/fold.rs
```diff
@@ -442,6 +442,24 @@ where
     }
 }
 
+impl<T, TF> Fold<TF> for UCanonical<T>
```
jackh726: I'm honestly not 100% sure what `Fold` does (or why). Does this make sense?

chalk-engine/src/logic.rs
```diff
@@ -115,7 +115,14 @@ impl<C: Context> Forest<C> {
         assert!(self.stack.is_empty());
 
         match self.ensure_answer_recursively(context, table, answer) {
-            Ok(EnsureSuccess::AnswerAvailable) => Ok(()),
+            Ok(EnsureSuccess::AnswerAvailable) => {
+                let answer = self.answer(table, answer);
+                if !answer.delayed_subgoals.is_empty() {
```
jackh726: Not sure what cases might hit this...no tests do currently.

jackh726:
Ok so I spent a bit of time digging into why two of the tests I added (`coinductive_conflicting2` and `coinductive_conflicting3`) are failing.

So, basically the `C1` and `C2` tables are Floundering because of the first check in `program_clauses` in `chalk-solve/solve/slg`. I still need to think about the logic here. But it seems separate from the coinductive cycles.

chalk-engine/src/logic.rs
jackh726: This has been moved to `goals_trivially_equal` but the same comment applies.

jackh726:
Before this is merged, it will definitely need to rebased to master. Additionally, there are some design questions I'm unsure about that need review and a second pair of eyes. Some are marked by `FIXME`s in the code. Additionally:
- I'm not sure if the way I'm applying the binders to the delayed subgoals is correct. It *works* for all the current and new tests, but I'm not sure if it's correct for *all* cases. If there are cases where it might not work, tests should be added. I tried to add a couple tests that I thought would break it, but it works as expected.
- Need to think about when a delayed subgoal can/would Flounder.
- In what cases will a delayed subgoal have a substitution and how does this get applied.
- The refinement part of this actually will probably be easier after #281 is merged.

nikomatsakis:
@jackh726 will you be able to rebase this atop the new master, which includes the nonrecursive changes?

jackh726:
@nikomatsakis yes I'll do this weekend

jackh726:
@nikomatsakis This has been rebased to master

chalk-engine/src/logic.rs
```diff
@@ -679,16 +678,29 @@ impl<C: Context> Forest<C> {
         let (table, subst, constraints, delayed_subgoals) =
             context.instantiate_answer_subst(num_universes, &answer.subst);
 
+        // FIXME: really, these shouldn't be added to the delayed_subgoals at all
+        // however, because we can't compare the table goal until `canonicalize_answer_subst`
+        // is called in `pursue_answer`, this will require a bit of refactoring work
```
nikomatsakis: I don't really understand this comment -- oh, are you saying that "really there shouldn't be anything in `delayed_subgoals` that is equal to `table_goal`? If so, I'm not sure I agree, I think that can arise after unification, but you are correct that detecting that equality cannot reliably be done with `==`. (But you don't need unification, we could make a method that does it.) Anyway, if that's what you mean, I agree, but we should probably file an issue and/or write a better comment.

jackh726: Ok, so there's a couple things going on here, and I'll try to explain it succinctly since I'm traveling today (so on mobile at least for the minute).

So first, it's maybe not obvious from the PR (and maybe not even *wrong*), but something I found when looking through `CHALK_DEBUG` output. But in `apply_answer_subst`, we can't *properly* check if the delayed subgoals are equal to the table goal, because the delayed subgoals get unique inference variables. (Perhaps this would be easier/quicker to check using the variant3 test and adding dbg statements there.)

Second, I would REALLY prefer to have these be checked *before* they are added to the answer. But, because of 1,  we have to wait until they are both canonicalized, which happens in pursue_answer. At that point it is *too late* because that doesn't know about the subgoal. That was mostly the point of this FIXME: we can be smarter here I think, but it requires a bit of refactoring.

Third, there are essentially two places that do this check. First is in `apply_answer_subst` and second is here. They aren't the same: removing this causes the variant3 test to fail. Removing the other causes...many of the conductive tests to fail (and a never-ending test for one that I don't remember off the top of my head). I'm not convinced that there couldn't be just one check in the right place.

So, overall. I'm a bit skeptical. Between both checks, everything works. But as I said, I don't know if it's perfectly correct and I definitely think there's room for improvement. Whether that needs to be done in this PR or can be done separately idk. 

Nonetheless, I should have some time soon to, at the very least, cleanup this FIXME and a couple other things. (Like the extra apply_binders function that was added but isn't used anymore). If we decide to land the PR as is, I'll file a followup issue with more detailed info.

nikomatsakis: OK, thanks for elaborating. Personally I'm inclined to convert this comment into an issue and land the PR, since I think it's clearly 'progress', but I definitely agree we can refactor this part of it. I'd be happy to take some time to chat it out.

jackh726:
This PR is good to merge now! I'll file a followup issue for the changes we talked about, and we can discuss.

nikomatsakis:
FYI @jackh726 I rebased this

PR #271, opened 2019-10-30
Create tests directory

detrumi:
Second attempt at #224 

Changes:
- Moved tests to `tests` directory
- `chalki` is now in `src/main.rs`
- Code that's used by both `chalki` and the tests is moved to the new `chalk-integration` crate

PR #276, opened 2019-11-05
Implement lowering errors manually

### LINKED ISSUE #233: Consider replacing failure with std::error

lnicola:
Fixes #233, but there's one more `failure` in the tree via `lalrpop -> (ascii-canvas ->) term -> dirs -> redox_users`. Bumping `term` should fix it. See also https://rustsec.org/advisories/RUSTSEC-2018-0015.html.

Should be reviewed one commit at a time. Note that I might have messed up some of the messages. Sorry for the churn.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,7 +159,6 @@ dependencies = [
  "chalk-solve 0.1.0",
  "diff 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


matklad:
LGTM, though I don't have r+ in this repo. It might make sense to move errors out to a separate file, as there are a lot of them!

nikomatsakis:
@lnicola this is awesome, but will have to be rebased -- we shuffled around the test file organization in #271. I hope it's not too painful, hopefully git should track everything successfully...? 🤞

lnicola:
Ah, no worries, I'll try to rebase it today.

lnicola:
I hope it's okay now.

nikomatsakis:
@lnicola bump away, this crate is not semver stable at all

nikomatsakis:
Oh, you mean in ascii-canvas etc.

PR #277, opened 2019-11-07
Remove CanonicalStrand

### LINKED ISSUE #255: simplify chalk context trait and logic-engine

jackh726:
Fixes #255

### DIFF
--- a/tests/test/slg.rs
+++ b/tests/test/slg.rs
@@ -294,7 +294,7 @@ fn only_draw_so_many_blow_up() {
 
         goal {
             exists<T> { T: Foo }
-        } fixed 767 with max 10 {
+        } fixed 2 with max 10 {
             // FIXME: should be 2?
             "Some(Ambig(Definite(Canonical { value: [?0 := Vec<^0>], binders: [Ty(U0)] })))"
         }
--- a/tests/test/slg.rs
+++ b/tests/test/slg.rs
@@ -294,7 +294,8 @@ fn only_draw_so_many_blow_up() {
 
         goal {
             exists<T> { T: Foo }
-        } fixed 2 with max 10 {
+        } fixed 767 with max 10 {
+            // FIXME: should be 2?
             "Some(Ambig(Definite(Canonical { value: [?0 := Vec<^0>], binders: [Ty(U0)] })))"
         }
     }
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -21,7 +21,7 @@ pub(crate) mod prelude;
 /// active. First, there is always the *global* context, but when we
 /// are in the midst of pursuing some particular strand, we will
 /// instantiate a second context just for that work, via the
-/// `instantiate_ucanonical_goal` and `instantiate_ex_clause` methods.
+/// `instantiate_ucanonical_goal` method.
 ///
 /// In the chalk implementation, these two contexts are mapped to the
 /// same type. But in the rustc implementation, this second context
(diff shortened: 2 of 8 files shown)


tests/test/slg.rs
```diff
@@ -294,7 +294,8 @@ fn only_draw_so_many_blow_up() {
 
         goal {
             exists<T> { T: Foo }
-        } fixed 2 with max 10 {
+        } fixed 767 with max 10 {
```
nikomatsakis: what's up with this? =)

jackh726: I honestly am not sure. Was hoping maybe you had some insight?

nikomatsakis: OK :) I'll take a look

nikomatsakis: OK -- I THINk I figured out the problem here. It's sort of subtle. The problem has to do with this rather ill-documented mechanism of `any_future_answer`. The problem, I believe, is that the `any_future_answer` method is being invoked with substitutions that include inference variables but some of those inference variables are not fully normalized. This leads to `may_invalidate` giving the wrong results -- and what's worse, it doesn't have access to the inference tables it needs.

The easiest fix would be to canonicalize before invoking the callback in `any_future_answer`, but that seems ungreat, since this routine is invoked frequently, and canonicalization is potentially expensive. I have a potential fix that adds a `normalize_subst` routine that we can invoke, but that's actually a similar-ish cost and also a bit of a "footgun", in that we don't want this operation 99% of the time. We could also thread down the inference table, but *that* is a bit annoying since sometimes the substitutions are coming from things that have been canonicalized (which the current code, with its assertions, doesn't seem to be accounting for).

Anyway, my "not great" fix is also not actually seeming to work, so let me make sure I am right before discussing deeper.

nikomatsakis:
I am wondering -- this is basically a runtime optimization. There are pros/cons to it. It saves execution time (less canonicalization) in favor of more memory usage (lots of inference tables). I'm wondering if we should hold off on this change until we are able to get a better idea of performance. I have ideas for how to make canonicalization really fast at runtime which I've never had a good reason to try and implement, so it's not clear that this tradeoff is worthwhile.

tests/test/slg.rs
jackh726: Eek. I definitely didn't expect that. I'll have to take a look

jackh726:
I'm curious what your ideas are for fast canonicalization? I think that coupled with maybe a slight refactor to some of the `instantiate_ex_clause` (have to think about this), in order to lessen the need for 10ish LOC just to convert a CanonicalStand to Strand might be sufficient

jackh726:
Before this is merged (if it would get merged), we should probably think about the performance vs. memory trade-offs of this, possibly accompanied by benchmarks. Probably worth it's own discussion and/or Hackmd doc. As is, I don't think it's ready for merge because of these design questions.

nikomatsakis:
We discussed in our meeting today and decided to close this and revisit if/when profiling leads us here =)

PR #279, opened 2019-11-08
Refactor recursion in pursue_strand

jackh726:
This factors out the change to recursion in `pursue_strand` from #272. With that, no stack protection feature is needed.

I've also added changes that I did further downstream, not in the PR, to pass strand by mutable reference to `incorporate*subgoal`.

No semantic changes are made with this.

### DIFF
--- a/chalk-engine/src/lib.rs
+++ b/chalk-engine/src/lib.rs
@@ -290,3 +290,25 @@ impl DepthFirstNumber {
         DepthFirstNumber { value }
     }
 }
+
+/// Because we recurse so deeply, we rely on stacker to
+/// avoid overflowing the stack.
+#[cfg(feature = "stack_protection")]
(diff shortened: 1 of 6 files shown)


chalk-engine/src/lib.rs
```diff
@@ -290,25 +290,3 @@ impl DepthFirstNumber {
         DepthFirstNumber { value }
     }
 }
-
-/// Because we recurse so deeply, we rely on stacker to
```
nikomatsakis: We may still need stack protection, although I agree it should be far, far less. But it's possible for us to recurse quite deeply across traits.

jackh726: Where do you think it should go? `ensure_answer_recursively`?

jackh726: I've also been thinking about how to maybe even remove *that* recursion though. But that wouldn't be for this PR.

PR #281, opened 2019-11-10
Convert ensure_answer_recursively to be iterative instead of recursive

jackh726:
Basically all of `pursue_strand`, `pursue_next_strand`, `incorporate_result_from_positive_subgoal`, and `incorporate_result_from_negative_subgoal` got merged into the one function, with changes made to make it iterative.

Right now, a lot of code has not been moved into separate functions, but I'm definitely working on this. All in all, it's a bit more LOC, but I've also added in a lot of documentation/comments to help follow what's happening.

### DIFF
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -237,7 +237,10 @@ impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C
         })
     }
 
-    fn any_future_answer(&mut self, test: impl FnMut(&C::Substitution) -> bool) -> bool {
+    fn any_future_answer(
+        &mut self,
+        test: impl FnMut(&C::InferenceNormalizedSubst) -> bool,
+    ) -> bool {
         self.forest.any_future_answer(self.table, self.answer, test)
     }
 }
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -199,18 +199,15 @@ impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C
     /// Panics if a negative cycle was detected.
     fn peek_answer(&mut self) -> Option<Answer<C>> {
         loop {
-            dbg!(&self.table, &self.answer);
             match self
                 .forest
                 .root_answer(self.context, self.table, self.answer)
             {
                 Ok(answer) => {
-                    dbg!(&answer);
                     return Some(answer.clone());
                 }
 
                 Err(RootSearchFail::Floundered) => {
-                    dbg!("Floundered");
                     let table_goal = &self.forest.tables[self.table].table_goal;
                     return Some(Answer {
                         subst: self.context.identity_constrained_subst(table_goal),
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -67,8 +67,11 @@ impl<C: Context> Forest<C> {
         for i in 0..num_answers {
             let i = AnswerIndex::from(i);
             loop {
-                match self.ensure_root_answer(context, table, i) {
-                    Ok(()) => break,
+                match self.root_answer(context, table, i) {
+                    Ok(answer) => {
+                        answers.push(answer.clone());
+                        break;
+                    }
                     Err(RootSearchFail::Floundered) => return None,
                     Err(RootSearchFail::QuantumExceeded) => continue,
                     Err(RootSearchFail::NoMoreSolutions) => return Some(answers),
@@ -82,8 +85,6 @@ impl<C: Context> Forest<C> {
                     }
                 }
             }
-
-            answers.push(self.answer(table, i).clone());
         }
 
         Some(answers)
@@ -200,10 +201,9 @@ impl<'me, C: Context, CO: ContextOps<C>> AnswerStream<C> for ForestSolver<'me, C
         loop {
             match self
                 .forest
-                .ensure_root_answer(self.context, self.table, self.answer)
+                .root_answer(self.context, self.table, self.answer)
             {
-                Ok(()) => {
-                    let answer = self.forest.answer(self.table, self.answer);
+                Ok(answer) => {
                     return Some(answer.clone());
                 }
 
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -4,15 +4,18 @@ use crate::logic::RootSearchFail;
 use crate::stack::{Stack, StackIndex};
 use crate::table::AnswerIndex;
 use crate::tables::Tables;
-use crate::{Answer, DepthFirstNumber, TableIndex, TimeStamp};
+use crate::{Answer, TableIndex, TimeStamp};
 
 pub struct Forest<C: Context> {
     context: C,
     pub(crate) tables: Tables<C>,
     pub(crate) stack: Stack<C>,
 
-    dfn: DepthFirstNumber,
-    work: TimeStamp,
+    /// This is a clock which always increases. It is
+    /// incremented every time a new subgoal is followed.
+    /// This effectively gives us way to track what depth
+    /// and loop a table or strand was last followed.
+    clock: TimeStamp,
 }
 
 impl<C: Context> Forest<C> {
@@ -21,8 +24,7 @@ impl<C: Context> Forest<C> {
             context,
             tables: Tables::new(),
             stack: Stack::default(),
-            dfn: DepthFirstNumber::MIN,
-            work: TimeStamp::default(),
+            clock: TimeStamp::default(),
         }
     }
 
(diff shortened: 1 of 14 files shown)


chalk-engine/src/forest.rs
```diff
@@ -4,15 +4,15 @@ use crate::logic::RootSearchFail;
...
-use crate::{DepthFirstNumber, TableIndex};
+use crate::{Answer, DepthFirstNumber, TableIndex, TimeStamp};
 
 pub struct Forest<C: Context> {
     context: C,
     pub(crate) tables: Tables<C>,
     pub(crate) stack: Stack,
 
     dfn: DepthFirstNumber,
+    work: TimeStamp,
```
nikomatsakis: I'd appreciate a comment here -- what does `work` represent?

chalk-engine/src/table.rs
```diff
@@ -72,8 +68,18 @@ impl<C: Context> Table<C> {
...
+    pub(crate) fn pop_next_strand_if(
+        &mut self,
+        test: impl Fn(&Strand<C>) -> bool,
+    ) -> Option<Strand<C>> {
+        let strand = self.strands.pop_front();
+        if let Some(strand) = strand {
+            if test(&strand) {
+                return Some(strand);
+            }
+            self.strands.push_front(strand);
```
nikomatsakis: isn't this going to just test and re-test the same strand over and over?

chalk-engine/src/table.rs
```diff
@@ -116,14 +120,12 @@ impl<C: Context> Table<C> {
...
-            Entry::Occupied(mut entry) => {
+            Entry::Occupied(entry) => {
                 let was_ambiguous = entry.get();
-                if !was_ambiguous || answer.ambiguous {
-                    false
-                } else {
-                    *entry.get_mut() = false;
-                    true
+                if *was_ambiguous && !answer.ambiguous {
+                    panic!("New answer was not ambiguous whereas previous answer was.");
```
nikomatsakis: This doesn't seem like a valid reason to panic to me -- it can definitely happen that we find an ambiguous answer through one route, but a valid (unambiguous) answer through another.

nikomatsakis: I don't quite understand why it was changed.

nikomatsakis: I should probably create some tests to exercise this path, huh?

nikomatsakis: That said, I think we could just ignore the answer and return `false` if we already have a better answer

chalk-engine/src/logic.rs
```diff
@@ -248,36 +223,70 @@ impl<C: Context> Forest<C> {
             self.tables[table].pop_next_strand_if(|strand| strand.last_pursued_time < work)
         {
             strand.last_pursued_time = work;
-            let result: StrandResult<C, ()> = self.pursue_strand(context, depth, strand);
```
nikomatsakis: Maybe I'll see the reason why, but it feels like this makes the enclosing function rather harder to read -- I'd rather not have nested loops in one function if we can avoid it.

nikomatsakis: Well, I guess I can see why, we do often return straight from there

chalk-engine/src/logic.rs
```diff
@@ -126,6 +126,44 @@ impl<C: Context> Forest<C> {
             .any(|strand| test(&strand.ex_clause.subst))
     }
 
+    fn ensure_precheck(
```
nikomatsakis: Needs a comment and a better name, I think. "precheck" doesn't give me a lot of feeling for what it does.

chalk-engine/src/logic.rs
```diff
@@ -157,38 +195,15 @@ impl<C: Context> Forest<C> {
...
-
-            if self.top_of_stack_is_coinductive_from(depth) {
-                return Ok(EnsureSuccess::Coinductive);
-            }
-
-            return Err(RecursiveSearchFail::PositiveCycle(Minimums {
-                positive: self.stack[depth].dfn,
-                negative: DepthFirstNumber::MAX,
-            }));
+        match self.ensure_precheck(table, answer) {
```
nikomatsakis: Nit: if let?

chalk-engine/src/logic.rs
```diff
@@ -157,38 +195,15 @@ impl<C: Context> Forest<C> {
...
+            Some(res) => return res,
+            None => {}
         }
 
         let dfn = self.next_dfn();
-        let depth = self.stack.push(table, dfn);
+        let work = self.increment_work();
```
nikomatsakis: Seems surprising to me that we implement `work` here -- I feel like that will cause us to do more work in the case of cycles than is warranted. In fact, "work" feels like the wrong thing to be tracking -- aren't we basically just tracking overall answers produced?

chalk-engine/src/table.rs
jackh726: So, it would, but the assumption is that any `Strand`s after this have a `>= work`.

chalk-engine/src/table.rs
jackh726: So yes, this path is never taken in any tests. I changed it to an explicit panic here because returning `true` here means that upstream the answer index gets incremented (since it's assumed `true` here means a new answer.

jackh726: So, if we want the behavior here to remain the same (that new unambiguous answers can replace ambiguous ones), then code elsewhere needs to change too.

chalk-engine/src/logic.rs
jackh726: Yep, I agree :) Name was changed in later commits.

chalk-engine/src/logic.rs
jackh726: I think you're correct in that this is the same as overall answers produced. But I need to look more into this. In *most* cases (at least from what I was seeing). `work` tracks `dfn` 1:1. So I need to think about what counter-examples exists and if that matters.

jackh726:
Before this is merged I need to rebase to `master` and remove the commits from the `CanonicalStrand` pr. I'll get this done sometime today. For the sake of simplicity of review, I'll make these as new commits (so the branch won't actually be mergeable probably, but the complete diff will reflect the final changes). @nikomatsakis when you're done reviewing, I'll squash all the commits and actually make it mergeable. Otherwise, I think it'll be kind of difficult to reapply the commits for this PR onto master directly.

jackh726:
@nikomatsakis Ok, I've reverted *most* of the changes from the `CanonicalStrand` PR here. However, there are a couple things that aren't basically identical to master:
- In `push_initial_strands_instantiated`, I've kept the change that clones the `InferenceTable` instead of relying on `resolvent_clause` to return a `CanonicalStrand` and rolling back the changes. I don't really think there's a performance difference and it's a bit cleaner this way in my opinion (and makes it easier in the future if/when we want to remove `CanonicalStrand`.
- `instantiate_ex_clause` returns a `(C::InferenceTable, Strand<C>)` rather than taking a `impl Fn(C::InferenceTable, Strand<C>) -> R`. When logic is recursive, this is fine, because the instantiated `Strand` is passed *into* `pursue_strand`. But with the iterative approach, this gets into the `active_strand` on the `Stack`. If we really need to go back to having it take a `impl Fn`, I think it's doable, but I didn't want to add yet another thing to follow.

Again, it still hasn't been rebased to the current master, but I'm just going to squash the commits before I do that and want to leave them separated for review, because I think that'll be easier.

nikomatsakis:
Per our discussion, @jackh726, I'm going to try and review this first -- sorry for the delay. On vacation this week so that'll slow me up but I might do a bit since this sort of thing is actually pretty fun. :)

chalk-engine/src/logic.rs
```diff
@@ -199,175 +142,737 @@ impl<C: Context> Forest<C> {
...
+                        // If the answer had delayed literals, we have to
+                        // ensure that `ex_clause` is also delayed. This is
+                        // the SLG FACTOR operation, though NFTD just makes it
+                        // part of computing the SLG resolvent.
+                        if self.answer(subgoal_table, answer_index).ambiguous {
+                            ex_clause.ambiguous = true;
+                        }
+
+                        // Increment time counter because we received a new answer.
+                        ex_clause.current_time.increment();
```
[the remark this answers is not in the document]
jackh726: I agree that having both can be a bit confusing. I'll change the name to `answer_time`. However, I'm not sure if there's a *better* way to do this, since one is bound to answers for a `Strand`, and the other really tracks the "clock" of pursuing an answer on the `Forest`. I'll try to also clarify the docs around these too.

chalk-engine/src/logic.rs
```diff
@@ -199,175 +142,737 @@ impl<C: Context> Forest<C> {
...
+        let selected_subgoal = strand.selected_subgoal.take().unwrap();
+        // We've already taken care of the case where we might be able
+        // to pursue this subgoal later (in a different strand). For
+        // this strand, we have an answer to this subgoal
+        match strand
+            .ex_clause
+            .subgoals
+            .remove(selected_subgoal.subgoal_index)
+        {
+            Literal::Positive(subgoal) => {
```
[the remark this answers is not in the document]
jackh726: Eh, I think I disagree on this one. Most of the `Negative` arm is comments and the `Positive` arm is really not that long.

jackh726:
@nikomatsakis your comments should be addressed (other than the one nit I disagree with, see comment). I decided to just inline the `check_table` checks since half of them aren't needed for one of the two call sites anyways.

jackh726:
🎉

PR #283, opened 2019-11-12
Refactor fold

nikomatsakis:
Refactor the fold crate to take a Target Family, so we now have `Fold<TF, TTF>`, where `TF` is the *source* target family and `TTF` is the *target* type family.

As part of this change, we no longer use `TF::Type` to reference types. Instead, we use `Ty<TF>` (and `Lifetime<TF>`). This is both more uniform (you always do `Foo<TF>` now no matter what), it also allows `Ty<TF>` to implement `for<TTF> Fold<TF, TTF>`, something which couldn't be expressed otherwise owing to the fact that Rust doesn't use Chalk yet.

I'm putting this as r? @detrumi because they are probably the one most familiar with fold after writing the derive, but @detrumi if you prefer I can try to find another reviewer. =)

### DIFF
--- a/chalk-ir/src/family.rs
+++ b/chalk-ir/src/family.rs
@@ -3,7 +3,7 @@ use crate::debug::Angle;
 use crate::fold::{Fold, Folder, ReflexiveFold};
 use crate::tls;
 use crate::zip::Zip;
-use crate::Lifetime;
+use crate::LifetimeData;
 use crate::Parameter;
 use crate::ParameterKind;
 use crate::ProjectionTy;
@@ -42,7 +42,7 @@ pub trait TypeFamily: Debug + Copy + Eq + Ord + Hash {
         + Hash
         + ReflexiveFold<Self>
         + Zip<Self>
-        + Lookup<Lifetime<Self>>
+        + Lookup<LifetimeData<Self>>
         + CastTo<Parameter<Self>>;
 
     /// Prints the debug representation of a projection. To get good
@@ -63,7 +63,7 @@ pub trait TypeFamily: Debug + Copy + Eq + Ord + Hash {
 
     /// Create an "interned" type from `lifetime`. You can also use
     /// the `Lifetime::intern` method, which is preferred.
-    fn intern_lifetime(lifetime: Lifetime<Self>) -> Self::Lifetime;
+    fn intern_lifetime(lifetime: LifetimeData<Self>) -> Self::Lifetime;
 }
 
 /// Implemented by types that have an associated type family (which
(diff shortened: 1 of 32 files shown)


chalk-ir/src/fold.rs
```diff
@@ -527,17 +642,17 @@ where
     }
 }
 
-impl<C: Context, TF: TypeFamily> Fold<TF> for ExClause<C>
+impl<C: Context, TF: TypeFamily> Fold<TF, TF> for ExClause<C>
```
detrumi: Why isn't `Fold<TF, TTF>` implemented for `ExClause`, `FlounderedSubgoal` and `Literal`? Are they never useful?

chalk-ir/src/fold.rs
```diff
@@ -55,20 +64,22 @@ pub use self::subst::Subst;
 /// ```rust,ignore
 /// let x = x.fold_with(&mut folder, 0);
 /// ```
-pub trait Folder<TF: TypeFamily>:
-    FreeVarFolder<TF> + InferenceFolder<TF> + PlaceholderFolder<TF> + TypeFolder<TF>
+pub trait Folder<TF: TypeFamily, TTF: TypeFamily>:
```
detrumi: Would it make sense to add a default here?
```Fold<TF: TypeFamily, TTF: TypeFamily = TF>```

Things like `Fold<ChalkIr>` read a bit better than `Fold<ChalkIr, ChalkIr>`, though I can also see that it could be confusing.

File: chalk-ir/src/fold.rs (hunk 1 above)
Comment (nikomatsakis) [reply chalk-ir/src/fold.rs]: Yeah, just not needed. I could probably generalize it anyway, I didn't want to mess with that stuff.

File: chalk-ir/src/fold.rs (hunk 2 above)
Comment (nikomatsakis) [reply chalk-ir/src/fold.rs]: I debated about it. It probably makes sense.

File: chalk-ir/src/family.rs [hunk 3]
```diff
@@ -63,12 +41,17 @@ pub trait TypeFamily: Debug + Copy + Eq + Ord + Hash {
     ) -> fmt::Result;
 
     /// Create an "interned" type from `ty`. You can also use the
-    /// `Ty::intern` method, which is preferred.
-    fn intern_ty(ty: Ty<Self>) -> Self::Type;
+    /// `TyData::intern` method, which is preferred.
+    fn intern_ty(ty: TyData<Self>) -> Self::InternedType;
```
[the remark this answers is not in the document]
nikomatsakis: You can't remove it from the trait -- that is how `ty.intern()` is implemented. :) I can improve the comment, though.

chalk-ir/src/lib.rs
```diff
@@ -212,7 +211,69 @@ pub enum TypeSort {
...
+
+impl<TF: TypeFamily> HasTypeFamily for Ty<TF> {
+    type TypeFamily = TF;
+}
+
+impl<TF: TypeFamily> Ty<TF> {
+    pub fn new(data: impl CastTo<TyData<TF>>) -> Self {
+        Ty {
+            interned: TF::intern_ty(data.cast()),
+        }
```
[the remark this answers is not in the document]
nikomatsakis: Actually, I think `intern()` invokes this method

nikomatsakis:
OK, I applied the feedback.

nikomatsakis:
Er, I pushed to the wrong branch. Let me fix that.

chalk-ir/src/family.rs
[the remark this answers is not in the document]
detrumi: Do you mean that some implementations of `TypeFamily` will end up needing it? Because Chalk compiles just fine if you remove it from the trait. None of the current `intern()` impl's use it and `Ty::new()` just calls `intern()` directly now.

PR #284, opened 2019-11-12
make chalk-rust-ir generic over type-family

nikomatsakis:
Building on #283, make the rust-ir types generic over type family.


nikomatsakis:
@tmandry, do you want to review? 

I extended the PR now so that it extends the generics all the way through chalk-solve, which is probably as far as we want to go (I envision rustc eventually "hooking in" to chalk-solve).

PR #288, opened 2019-11-18
ci: switch to github actions

pietroalbini:
This should also fix the failure to deploy documentation.

r? @nikomatsakis


lnicola:
It might be worth disabling `CARGO_INCREMENTAL` for CI.

pietroalbini:
Probably, but it shouldn't affect stuff *too much* as I didn't use caching in this PR.

PR #293, opened 2019-11-22
Change QuantifiedTy to QuantifiedApply

detrumi:
Note: the `forall_projection` test is currently triggering a stack overflow, but I haven't been able to get a trace:
```
$ RUST_BACKTRACE=1 target/debug/chalk
?- program
Enter a program; press Ctrl-D when finished
| trait Eq<T> { }
| impl<T> Eq<T> for T { }
| 
| trait DropLt<'a> { type Item; }
| impl<'a, T> DropLt<'a> for T { type Item = T; }
| 
| struct Unit { }
| struct Ref<'a, T> { }
| 
?- for<'a> <Unit as DropLt<'a>>::Item: Eq<Unit>

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

### DIFF
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -165,7 +165,7 @@ AssocTyValue: AssocTyValue = {
 };
 
 pub Ty: Ty = {
-    "for" "<" <l:Comma<LifetimeId>> ">" <t:Ty> => Ty::ForAll {
+    "for" "<" <l:Comma<LifetimeId>> ">" "fn" "(" <t:Ty> ")" => Ty::ForAll {
         lifetime_names: l,
         ty: Box::new(t)
     },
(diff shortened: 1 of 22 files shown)


bjorn3:
> but I haven't been able to get a trace:

For stack overflows, you need to use a debugger to get a backtrace. libbacktrace doesn't support unwinding the original stack from a signal handler.

detrumi:
Looks like unification gets stuck in a loop, so maybe substitution is wrong?

nikomatsakis:
@detrumi I guess the next step is to update the parser

chalk-parse/src/parser.lalrpop
```diff
@@ -165,7 +165,7 @@ AssocTyValue: AssocTyValue = {
 };
 
 pub Ty: Ty = {
-    "for" "<" <l:Comma<LifetimeId>> ">" <t:Ty> => Ty::ForAll {
+    "for" "<" <l:Comma<LifetimeId>> ">" "fn" "(" <t:Ty> ")" => Ty::ForAll {
```
nikomatsakis: we should probably also permit the syntax ``"fn" "(" ... ")"` with an empty list of lifetime ids

PR #295, opened 2019-11-22
Use string_cache instead of lalrpop-intern

### LINKED ISSUE #287: Remove dependency on lalrpop-intern
Salsa uses `lalrpop-intern` internally for string interning. However, `lalrpop-intern` is not thread safe: its IDs are `Send`, but the interner is stored in a thread-local. 

cc @flodiebold : I remember discussing this issue before, have we just forgotten about this?

jackh726:
Fixes #287 

Unfortunately, `DefaultAtom` from `string_cache` is not `Copy`, so have to add `.clone()` in many places.


nikomatsakis:
Hmm. I think what I'd rather do is not (at least outside the AST) encode any specific string representation, instead moving that into the the context.

jackh726:
Ok, in that case I won't be able to change this PR for at least a week.

nikomatsakis:
I'm going to close this PR, but I think we could best make this change atop the proposal in #309

PR #296, opened 2019-11-23
Add regression test for issue 111

### LINKED ISSUE #111: figure out why we get ambiguity for this futures example
Given the traits/types found in this gist:

https://gist.github.com/nikomatsakis/58717b0588393ca14f2d6168e7c86188

Why do we get ambiguity here?

```bash
> target/debug/chalki --program=$HOME/tmp/futures-model.chalk --goal='forall<T> { if (T: FutureResult) { exists<I, E> { T: Future<Output = Result<I, E>> } } }'
Ambiguous; no inference guidance
```

Note that if we (hackily) don't force the solver to give us back a result for `I` and `E`, we get `Unique` as expected:

```bash
> target/debug/chalki --program=$HOME/tmp/futures-model.chalk --goal='forall<T> { if (T: FutureResult) { T: Future, exists<I, E> { T: Future<Output = Result<I, E>> } } }'
Unique; substitution [], lifetime constraints []
```

cc @scalexm, who wanted to investigate this

jackh726:
Closes #111 

As noted in https://github.com/rust-lang/chalk/issues/111#issuecomment-548571024, this has been fixed. Added a test so we can close the issue.

### DIFF
--- a/tests/test/misc.rs
+++ b/tests/test/misc.rs
@@ -1,0 +1,42 @@
+//! Tests that don't fit a single category
+
+use super::*;
+
+// Regression test for rust-lang/chalk#111
+#[test]
+fn futures_ambiguity() {
+    test! {
+        program {
+            struct Result<T, E> { }
+
+            trait Future {
+                type Output;
+            }
+
+            trait FutureResult
+                where
+                Self: Future<Output = Result<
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Huh, ok. Well, good enough.

PR #304, opened 2019-12-06
experimentally add an (empty) mdbook

nikomatsakis:
It... should get published to http://rust-lang.github.io/chalk/

Maybe @pietroalbini can advise me on the github actions setup ;)


nikomatsakis:
(One thing I was wondering was whether to build/test the book in a separate "action", similar to how formatting is done)

nikomatsakis:
It seems like it'd be important to cache the mdbook result -- or maybe [download binaries](https://github.com/rust-lang/mdBook/releases)

pietroalbini:
So, while for other repositories splitting the book build and the crate build in separate jobs this would've been great (as it runs them in parallel), it doesn't really work for Chalk.

The issue is, deploying to GitHub Pages overrides the previously uploaded content, and in the current CI config you try to upload both the crate docs and the book in separate steps. That means one will override the other. There are two solutions here:

* Build the crate docs and the book in the same job, uploading them together at the end.
* Have a builder for building the crate (docs) and one builder to build the book: each of them upload [a GitHub Actions artifact](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/persisting-workflow-data-using-artifacts) containing the generated content, and then a separate builder downloads both artifacts, merges them together and uploads the result to GitHub Pages.

The second approach is best if building both the crate docs and the book takes time, but (modulo downloading mdbook) building the book takes no time at all. So, I think it makes more sense to just go with the first approach.

You'll need to move all the book building steps at the end of the `test` job (but before the upload), and change the deploy step to copy the book somewhere in the deployed directory.

> One thing I was wondering was whether to build/test the book in a separate "action", similar to how formatting is done

This makes sense if building and testing take a lot of time (like, more than a minute each), but with such fast building and testing time it's not worth it.

> It seems like it'd be important to cache the mdbook result -- or maybe download binaries

Yeah. Downloading the binary is probably easier and faster, as you don't need to worry about what to cache and you don't need to rebuild mdbook when the cache expires.

nikomatsakis:
@pietroalbini ok I reverted my commit where I tried creating a "parallel" github action

nikomatsakis:
I also hacked up some logic to download binary and merge the output. It's probably wrong. I'm not sure if there's even a way to test this...?

pietroalbini:
The changes look good to me!

> I also hacked up some logic to download binary and merge the output. It's probably wrong. I'm not sure if there's even a way to test this...?

The build and test code surely works, as those steps are run on PRs as well and the build is green! The deployment changes you made also *look* correct.

What I usually do for testing is to push a WIP commit that removes the condition, to see if it actually works before merging to master:

```diff
-if: matrix.rust == 'stable' && github.ref == 'refs/heads/master'
+if: matrix.rust == 'stable'
```

Then if everything works I remove that commit and merge the PR. A bit of an issue here is that you can't do that, as this PR comes from a fork and (rightfully) forks can't access secret variables. It would've worked if you pushed your branch to the main repository though.

pietroalbini:
Pushed a commit doing that in a temporary `test-niko-ci` branch: https://github.com/rust-lang/chalk/commit/12a371fcef1b1835d38bd3c35dc942a9a5d5bddb

And it worked! https://github.com/rust-lang/chalk/runs/341591546



You can delete that branch and merge this PR!

PR #294, opened 2019-11-22
When truncating a goal, don't truncate the environment

jackh726:
Fixes #280 
Fixes #289

Not sure if this is the *correct* fix. It does solve the issue and, conceptually, I don't see a particular reason why we need to truncate the `environment` of a goal. However, there may be other cases we could theoretically still enter an infinite loop and this doesn't necessarily protect against that.

Issue #280, opened 2019-11-09
Infinite stack growth in chalk


matklad:
This issue is to track a problem, discussed in this [Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/infinite.20stack.20growth.20in.20chalk).

The original problem was observed in rust-analyzer (rust-analyzer/rust-analyzer#2052): chalk exhausts all available memory by expanding the stack: https://gist.github.com/matklad/c6320c19b08afc8e6d8ca8612354d5b8

Here's a stand-alone rust-analyzer test that shows the problem:

https://github.com/matklad/rust-analyzer/commit/d907b00f440ca0a127e6975994a54391ce5a976d

There's no isolated chalk test at the moment for this.

nikomatsakis:
Relevant factors:

* rust-analyzer size must be 4 (the default used in chalk, 10, makes the problem go away)

nikomatsakis:
OK, so I've reproduced the problem, and I know what it is, but I'm not sure yet what I think is the best solution.

You can find the test case on my branch [issue-280-rust-analyzer-infinite-growth][b]. 

[b]: https://github.com/nikomatsakis/chalk-ndm/tree/issue-280-rust-analyzer-infinite-growth

The problem is this:

* To impose the maximum size requirement of 4, we introduce inference variables
* The answer to these inference variables winds up being unnormalized projection types (I'm not quite sure how that comes about yet, actually, have to look a bit more into that)
* When we go to incorporate the answer, we wind up with new subgoals that never ends

Ah wait -- I think the problem is actually the opposite. The variable comes from the *answer*, and the unnormalized projection type appears in the environment (it's the type that was replaced with a variable).

flodiebold:
After #279, am I right in assuming this manifests as OOM instead of stack overflow now?

matklad:
@flodiebold to be precise, this never was a stackoverflow, because chalk used to grow the stack automatically.

jackh726:
Ok, so I've looked into this a bit tonight and figured I would write down more or less what I've found, in case other people want to take a look (or have ideas on the best solution).

@nikomatsakis's comment above (https://github.com/rust-lang/chalk/issues/280#issuecomment-552621555) is correct, but I'll go into more detail.

### Reproducing the issue ###
[...]

PR #305, opened 2019-12-07
permit inference variables to be unified with unnormalized projections

nikomatsakis:
This fixes the rust-analyzer problems found in #280 and serves as an alternative to https://github.com/rust-lang/chalk/pull/294. I'm not really sure which approach I prefer -- there might even be a "third way" that is better still.

If nothing else, though, I think that the regression tests found on this branch is the correct form. I had to make a few small fixes to the test macro so that we could select the solver and specify the "max size" correctly.


jackh726:
So, what's interesting about this approach is how it might overlap with the solution to the associated types double answer (#234). Like, the change to the other test output is interesting.

nikomatsakis:
Decided to close for now. Opened #306 to track my concerns with the current workaround.

PR #309, opened 2019-12-13
Move ids to type family

nikomatsakis:
This PR moves the `RawId` type into the type family. It is presented as `TF::DefId`, mirroring the name that rustc uses. The goal here is to allow rustc to map this to `DefId`.

For now, I kept the various newtypes (e.g., `StructId` and so forth) exactly as they are. At some point we are going to move to move things like `TypeName` so that we have an "interned" `TypeName`, but I'm not 100% sure about the various newtypes, they may want to stay like this, or else move into the `TypeFamily`, or maybe we just get rid of them (but I think it's nice to have the distinctions between various kinds of ids).

I think a good rule though would be that `TypeName` should be opaque to most of the chalk logic (apart perhaps from the clause generation logic, which would interact with it through methods on `RustIrDatabase`), which implies that we may want to extract some of its variants out into `TypeName`.

### DIFF
--- a/chalk-solve/src/coherence.rs
+++ b/chalk-solve/src/coherence.rs
@@ -14,7 +14,7 @@ mod solve;
 pub struct CoherenceSolver<'db, TF: TypeFamily> {
     db: &'db dyn RustIrDatabase<TF>,
     solver_choice: SolverChoice,
-    trait_id: TraitId,
+    trait_id: TraitId<TF>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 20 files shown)


nikomatsakis:
@detrumi I think you were poking at making a change like this, maybe you'd care to review?

chalk-solve/src/coherence.rs
```diff
@@ -41,19 +41,25 @@ impl std::error::Error for CoherenceError {}
...
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
-pub struct SpecializationPriorities {
-    map: BTreeMap<ImplId, SpecializationPriority>,
+pub struct SpecializationPriorities<TF: TypeFamily> {
+    map: BTreeMap<ImplId<TF>, SpecializationPriority>,
 }
 
-impl SpecializationPriorities {
+impl<TF: TypeFamily> SpecializationPriorities<TF> {
+    pub fn new() -> Self {
```
jackh726: If `SpecializationPriorities` derives `Default`, why is this needed?

nikomatsakis: True, we don't need this, we can write `::default`...

nikomatsakis: Ah, I remember. The problem is that `TF` doesn't necessarily implement `Default`. The fix would be either:

* to manually implement `Default` for `SpecializationPriorities<TF>` (for any `TF`, not just `TF: Default`)
* to require that all `TypeFamily: Default`, as we did with `Clone` and various other traits

The former is more typing that `new` which I think is why I didn't do it. The latter is actually a fine strategy so long as type families remain "dummy" types that just pull together a bunch of other logic. But I think that eventually we will want a `TypeFamily` to actually *be* an arena, in which case it may not want to implement `Default`. (i.e., in rustc, I expect a type family to be the `tcx`). That said, we could also make it so that you have `TF::Context` to represent the arena value -- but I think that may be more annoying in the end. I thought it'd be useful if we actually make a `TF` be the context itself, since that will help with type inference.

(I contemplated renaming `TF` to `TCX`... but I'm not sure we plan to keep that name in the compiler.)

nikomatsakis:
I'm going to go ahead and merge this as is for now. Seems like the question of `Default` is not a major problem.

PR #311, opened 2019-12-19
Cleanup chalk ir

nikomatsakis:
This is a kind of random series of edits cleaning up the chalk-ir and moving it closer to what I think we actually want. The main changes:

* Introduce a `TyData::Placeholder` variant and remove the `TypeName` entry
* Introduce `AssocTypeId` to replace `TypeId`
* Remove `TypeKindId` and instead just specify `StructId` or `TraitId` always
* Remove various uses of `Identifier` (but not all of them) from the `RustIrDatabase`
* Separate out the `fold` code a bit into more modules
* Replace `Goal::And` with `Goal::All` (which will be better when we are lowering projections and things)


nikomatsakis:
I'm still a bit unsure how to handle placeholders -- in particular, placeholders for associated types. I was initially thinking that we would want to unify them under `TyData::Placeholder`, then I changed my mind, now I'm thinking I was right to begin with. This is partly why I left various commits that show the stages, so it'd be easy to revert. I need to start writing up a doc that justifies and explains the various decisions.

chalk-integration/src/lowering.rs
```diff
@@ -47,12 +50,12 @@ struct Env<'k> {
...
 struct AssociatedTyLookup {
-    id: chalk_ir::TypeId<ChalkIr>,
+    id: chalk_ir::AssocTypeId<ChalkIr>,
     addl_parameter_kinds: Vec<chalk_ir::ParameterKind<()>>,
 }
 
-enum NameLookup {
-    Type(chalk_ir::TypeKindId<ChalkIr>),
+enum TypeLookup {
+    Struct(chalk_ir::StructId<ChalkIr>),
```
jackh726: So, something feels weird to me about having the name of this enum be `TypeLookup`, but only be able to be a struct or parameter. Maybe I need to look more at what this code does and is used for, but it seems...weird.

chalk-integration/src/lowering.rs
```diff
@@ -64,16 +67,39 @@ const SELF: &str = "Self";
 const FIXME_SELF: &str = "__FIXME_SELF__";
 
 impl<'k> Env<'k> {
-    fn lookup(&self, name: Identifier) -> LowerResult<NameLookup> {
+    fn lookup_type(&self, name: Identifier) -> LowerResult<TypeLookup> {
```
jackh726: Similarly here. Name of `lookup_type` can only return struct, parameter, or not struct.

chalk-integration/src/lowering.rs
```diff
@@ -915,63 +954,67 @@ trait LowerTy {
...
-                        .lower(env)?
-                        .iter()
-                        .flat_map(|qil| {
-                            qil.into_where_clauses(chalk_ir::TyData::BoundVar(0).intern())
-                        })
-                        .collect())
-                },
-            )?)
+            Ty::Opaque { ref bounds } => Ok(chalk_ir::TyData::Opaque(chalk_ir::BoundedTy {
+                bounds: env.in_binders(
```
jackh726: This code is straight duplicated. Maybe this is worth a helper function?

tests/lowering/mod.rs
```diff
@@ -343,7 +343,7 @@ fn gat_higher_ranked_bound() {
     lowering_success! {
         program {
             trait Fn<T> {}
-            trait Ref<'a, T> {}
+            struct Ref<'a, T> {}
```
jackh726: Why change this?

nikomatsakis: It was wrong to begin with. Traits are not types (and we now draw a sharper distinction). You can make a `dyn` type to reference `Ref`, but that is not what this was supposed to be testing.

chalk-integration/src/lowering.rs
nikomatsakis: The name choice was intentional: structs and parameters are the only types in the system. Perhaps the thing that is *actually* odd is using the term `StructId` when "structs" in Chalk are really meant to represent any user-defined type (i.e., structs, enums, and unions). I was actually considered renaming `StructId` to `UserTypeId` or something else like that.

In particular, traits are not types -- i.e., you can't say that value's type is `Iterator`. You *can* say that it is `dyn Iterator`, but in that case the role of the trait is not to *be* a type but to *bound* a type. In other words, that is short for `exists<T> { T: Iterator }` (this is particularly explicit in chalk's representation), and `T` is the type.

chalk-integration/src/lowering.rs
nikomatsakis: Right -- this looks up something that is intended to be a type.

chalk-integration/src/lowering.rs
nikomatsakis: Actually the `Opaque` code should be removed, I recently realized this is the wrong way to model it (that gist I've been preparing talks a bit about it). I'll do that now.

nikomatsakis:
I'm debating about how to go forward here. I think I will land this PR and create follow-up ones for further refactorings (e.g., removing `Opaque` or changing how placeholders for associated types work).

chalk-integration/src/lowering.rs
jackh726: Ah, I see. In that case, I'm fine with these changes. Changing `Struct` to `UserTypeId` sounds less nice. At that point, just go to `TypeId`.

jackh726:
Sounds good to me!

PR #312, opened 2019-12-21
draft proposal: chalk-ir

nikomatsakis:
This PR contains book chapters describing the chalk-ir crate. These chapters are also meant to serve as a proposal that we can discuss at a compiler team meeting, although I would expect such a meeting would focus more on the high-level bits than the *nitty gritty* of the design.

This text does not align precisely with the state of the master branch.

PR #315, opened 2019-12-31
more refactorings to chalk-ir

nikomatsakis:
Moving closer to the proposal in #312 

* Remove `TyData::Opaque` -- this was meant to be impl Trait, but I realized later that this would be modeled quite differently, as a form of "alias type" (not yet added)
* Rename `BoundTy` to `DynTy`, since that is more obvious what it means
* Make the representation of `Goal` part of `TypeFamily`
* Remove explicit uses of `Box` from `GoalData`
* Move to `Arc` for the chalk-ir
* Simplify `GoalData` by removing `LeafGoal`

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -221,6 +221,7 @@ impl<TF: TypeFamily> ToProgramClauses<TF> for StructDatum<TF> {
                     builder.push_clause(
                         DomainGoal::$goal(self_ty.clone()),
                         Some(DomainGoal::$goal(
+                            //
                             // This unwrap is safe because we asserted
                             // above for the presence of a type
                             // parameter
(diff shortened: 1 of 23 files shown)


chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -221,6 +221,7 @@ impl<TF: TypeFamily> ToProgramClauses<TF> for StructDatum<TF> {
                     builder.push_clause(
                         DomainGoal::$goal(self_ty.clone()),
                         Some(DomainGoal::$goal(
+                            //
```
jackh726: nit: extra line

tests/test/existential_types.rs
```diff
@@ -14,12 +14,6 @@ fn dyn_Clone_is_Clone() {
         } yields {
             "Unique; substitution []"
         }
-
-        goal {
-            impl Clone: Clone
```
jackh726: Should these tests be elsewhere? (Or just commented out?) Or are `impl Trait`s just going to be in lowering now? Presumably, we *do* want these tests in some form?

nikomatsakis: I think they should just be removed for now. We'll have to figure out how to test impl Trait when we grow support for it

PR #316, opened 2020-01-04
intern "substitutions"

nikomatsakis:
We used to directly embed `Vec<Parameter<TF>>`. This branch changes those references to `Substitution<TF>` which (internally) relies on an `TF::InternedSubstitution`. This should enable rustc to use interned slices.

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -409,7 +409,7 @@ impl<TF: TypeFamily> Substitution<TF> {
     /// Displays the substitution in the form `< P0, .. Pn >`, or (if
     /// the substitution is empty) as an empty string.
     pub fn with_angle(&self) -> Angle<'_, Parameter<TF>> {
-        Angle(&self.parameters)
+        Angle(self.parameters())
     }
 }
 
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -425,7 +425,7 @@ impl<TF: TypeFamily> Display for Substitution<TF> {
 
         write!(f, "[")?;
 
-        for (index, value) in self.parameters.iter().enumerate() {
+        for (index, value) in self.iter().enumerate() {
             if first {
                 first = false;
             } else {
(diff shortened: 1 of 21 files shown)


chalk-ir/src/debug.rs
```diff
@@ -150,27 +151,22 @@ struct SeparatorTraitRef<'me, TF: TypeFamily> {
...
-        TF::debug_projection(self, fmt).unwrap_or_else(|| {
-            write!(
-                fmt,
-                "({:?}){:?}",
-                self.associated_ty_id,
-                Angle(&self.parameters)
-            )
-        })
+        TF::debug_projection(self, fmt)
+            .unwrap_or_else(|| write!(fmt, "({:?}){:?}", self.associated_ty_id, &self.substitution))
```
jackh726: Should this be `&self.substitution.with_angle()`

chalk-ir/src/family.rs
```diff
@@ -146,8 +155,19 @@ pub trait TypeFamily: Debug + Copy + Eq + Ord + Hash {
     /// method).
     fn intern_goal(data: GoalData<Self>) -> Self::InternedGoal;
 
-    /// Lookup the `GoalData` that was interned to create a `InternedGoal`.
+    /// Lookup the `LifetimeData` that was interned to create a `InternedLifetime`.
```
jackh726: These docs don't reflect the function...

chalk-ir/src/lib.rs
```diff
@@ -1175,17 +1176,60 @@ pub enum Constraint<TF: TypeFamily> {
...
 
 /// A mapping of inference variables to instantiations thereof.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Fold, Hash, HasTypeFamily)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, HasTypeFamily)]
 pub struct Substitution<TF: TypeFamily> {
     /// Map free variable with given index to the value with the same
     /// index. Naturally, the kind of the variable must agree with
     /// the kind of the value.
-    pub parameters: Vec<Parameter<TF>>,
+    parameters: TF::InternedSubstitution,
```
jackh726: At this point, why not just replace `Substitution` with an associated type?

nikomatsakis: This pattern is what we use for all the "interned things". Having the newtype'd wrapper allows us to implement traits like `Fold` and add methods. You can't *quite* make `Fold` work without the newtype'd wrapper. The other advantage is consistency: you always write `Foo<TF>`, never `TF::Foo`, in regular code.

PR #317, opened 2020-01-05
Refactor tests in `slg` into the general test infra 

jackh726:
Before this, all the tests in `tests/test/slg` used their own form of `test!` which was different than all the other tests. This was not only confusing, but many tests that were in `tests/test/slg` could be put in a more logical place based on what they test.

This PR essentially just converts the `tests/test/slg` tests to the new format. In the process, a few changes occurred:
1) `AnswerStream` no longer returns `Option<CompleteAnswer<C>>`, but instead an `AnswerResult<C>` that makes the case of `Floundered` explicit.
2) `Context::make_solution` now takes a `UCanonicalGoalInEnvironment`. This might not be necessary (`identity_constrained_subst` takes it but might just be able to take `CanonicalGoalInEnvironment`), but I didn't dig into it. It also means that it gets `canonical`ized on every new answer.
3) `solve_multiple` now passes a `SubstitutionResult` to the passed function. This allows distinguishing ambiguity, as well as floundering.
4) `SlgContext` has a `expected_answers` field which is an `Option<usize>` and is checked when aggregating a solution and replaces `num_cached_answers_for_goal`. This is only used in two tests, so there might be a more elegant way to do this.
5) `yields[...]` now uses `TestGoal::Aggregated` whereas `yields_first[...]` and `yields_all[...]` use `TestGoal::All`. Only one `Solver` can be selected now, which is mostly because I was trying to figure out macros, but also is really unneeded complexity (at least for now).
6) `force_answers`, `num_cached_answers_for_goal`, and `TestSolver` have been removed since they are unneeded. This removes duplicated logic between `iter_answers` and `force_answers`.

One point to ask about: the `subgoal_cycle_uninhabited` test is now giving me a well-formed error, either though the program is basically identical to `subgoal_cycle_inhabited` minus an impl.


tests/test/slgext.rs
```diff
@@ -103,13 +103,49 @@ fn flounder() {
...
             // FiXME(jackh726): need to refactor framework to detect flouder/ambiguous
             "for<?U0> { substitution [?0 := ^0], lifetime constraints [] }"
         }
     }
 }
 
+// Test that, when solving `?T: Sized`, we only wind up pulling a few
+// answers before we stop.
+// This is similar to the `breadth_first` test, except the order of the 
+// FIXME: This is basically the same as `breadth_first`. Is it testing something different?
```
nikomatsakis: I think the only difference is the order of the impls, which can however be significant (e.g., which program clause do we try first?)

tests/test/slg.rs
```diff
@@ -1,147 +0,0 @@
...
-            // Strip trailing commas to handle both nightly and stable debug formatting
-            let result = result.replace(",\n", "\n");
-            let expected = expected.replace(",\n", "\n");
-            assert_test_result_eq(&expected, &result);
-        }
-    });
-}
-
-#[test]
-fn subgoal_cycle_uninhabited() {
```
nikomatsakis: Why remove this test?

nikomatsakis: Ah yeah, you wrote this:

> One point to ask about: the subgoal_cycle_uninhabited test is now giving me a well-formed error, either though the program is basically identical to subgoal_cycle_inhabited minus an impl.

Did you in fact move the test?

nikomatsakis: I believe that the "SLG Solver" tests skipped the well-formedness

tests/test/slgext.rs
jackh726: Right. I guess I just don't know if this is trying to test something different or not.

tests/test/slg.rs
jackh726: It's in `misc` now :) However, it's ignored because it's giving me a well-formed error and wanted to see if you could figure out why.

jackh726:
Ok @nikomatsakis what should I do here?

For the `only_draw_so_many` test, should I: remove the the `FIXME`, but keep the test; remove the test (since it doesn't seem to be testing that `breadth_first` isn't testing, even if order is different); or keep both the test and FIXME for someone in the future to check

For the `subgoal_cycle_uninhabited` test, can you tell what's going on? Not sure what to do.

jackh726:
@nikomatsakis This should be good to go now.

I removed the `breadth_first` test, since the `only_draw_so_many` test covers the same idea. The FIXME on the `only_draw_so_many` test has been removed and I added a small additional comment to the test to clarify that it's all testing breadth-first behavior.

I also changed it so that `checked_program` always uses the default `max_size`. Also made sure we don't nest `set_current_program`s. This honestly can be cleaned up at some point too, but for now it's a pretty minimal diff. Because the `max_size` was set to 2, the well-formed checks were failing. So, `subgoal_cycle_uninhabited` test is now working properly.

nikomatsakis:
@jackh726 ok -- did we ever figure out why the one test was yielding different results? I'm a bit behind on Zulip, sorry..

jackh726:
> @jackh726 ok -- did we ever figure out why the one test was yielding different results? I'm a bit behind on Zulip, sorry..

If you mean `subgoal_cycle_uninhabited` failing the WF check, yes. It was because it was using max size of 2 for that. Now, all WF checks use the default max size (10).

PR #319, opened 2020-01-07
Update GLOSSARY.md

declanvk:


GLOSSARY.md
```diff
@@ -72,8 +72,8 @@ literal is given the number of its binder. The indices start at zero from the
 innermost binder increasing from the inside out.
 
 Given the example `forall<T> { exists<U> { T: Foo<Item=U> } }` the
-literal names `U` and `T` are replaced with `0` and `1` respectively: `forall<0>
-{ exists<1> { 0: Foo<Item=1> } }`.
+literal names `U` and `T` are replaced with `0` and `1` respectively: `forall<1>
+{ exists<0> { 1: Foo<Item=0> } }`.
```
nikomatsakis: Hmm, the "forall" and "exists" binders should probably be "anonymous", as they don't correspond to any fixed numbering:

```
forall<_> { exists<_> { 1: Foo<Item = 0> } }
```

The idea is that the binder doesn't have a name, it is referenced by the depth at the point of use.

PR #320, opened 2020-01-08
Rename projection to alias

detrumi:
Most types with `projection` in the name were renamed to `alias`, but I've kept the projection-specific code (mostly `program_clauses.rs` and `split.rs` mostly the same. It's not entirely clear to me which parts will need to be generalized to all kinds of aliases, and which parts stay specific to associated type projections.

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -1021,9 +1021,7 @@ impl LowerTy for Ty {
                 .intern())
             }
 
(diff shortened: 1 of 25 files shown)


nikomatsakis:
Let's go ahead and take this step -- we may wind up changing details of it as we go, but I think it's the right direction to be heading in, and it'll also mean far fewer merge conflicts for follow-up work.

PR #321, opened 2020-01-11
Some cleanups and add a `solve_limited`

jackh726:
This *could* be two separate PRs, but the cleanups are small:
- Moved methods from `InferenceTable` onto `Context` directly, since they don't need an `&self`
- Made `any_future_answer` take a `&self` instead of a `&mut self`

The bulk of the PR involves adding a `should_continue` argument to `peek_answer` and `next_answer`, which allows stopping of solving before an answer is ready, when a `QuantumExceeded` has been returned. Right now, it's a argument-less function, so there's no real "feedback", so it's pretty much limited to either a time-based or call number-based implementation.
However, some ideas for how to extend this:
- Pass the current (or expected) substitution to the function. This could allow the caller to accept that an answer is "good enough".
- Be a bit smarter with the returned answer after `false` is returned. Right now, the returned answer is the current substitution as `Guidance(Suggested(...))`. Instead, we could take into account the *possible* answers (similar to how `any_future_answer` works). Or, maybe we just return `None` instead.

### DIFF
--- a/chalk-solve/src/solve.rs
+++ b/chalk-solve/src/solve.rs
@@ -135,7 +135,37 @@ impl<TF: TypeFamily> Solver<TF> {
         goal: &UCanonical<InEnvironment<Goal<TF>>>,
     ) -> Option<Solution<TF>> {
         let ops = self.forest.context().ops(program);
-        self.forest.solve(&ops, goal)
+        self.forest.solve(&ops, goal, || true)
+    }
+
+    /// Attempts to solve the given goal, which must be in canonical
+    /// form. Returns a unique solution (if one exists).  This will do
+    /// only as much work towards `goal` as it has to (and that work
+    /// is cached for future attempts). In addition, the solving of the
+    /// goal can be limited by returning `false` from `should_continue`.
+    ///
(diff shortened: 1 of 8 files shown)


jackh726:
So I'm not sure why CI is complaining. It builds/tests fine locally on stable 1.40.0. And the CI is complaining that `Fn` is a struct not a trait?

chalk-solve/src/solve.rs
```diff
@@ -135,7 +135,37 @@ impl<TF: TypeFamily> Solver<TF> {
...
+    /// goal can be limited by returning `false` from `should_continue`.
+    ///
+    /// # Parameters
+    ///
+    /// - `program` -- defines the program clauses in scope.
+    ///   - **Important:** You must supply the same set of program clauses
+    ///     each time you invoke `solve`, as otherwise the cached data may be
+    ///     invalid.
+    /// - `goal` the goal to solve
+    /// - `should_continue` if `false` is returned, the no further solving will be done
```
nikomatsakis: can you add a note that `Guidance` will be returned, in that case?

PR #324, opened 2020-01-20
Implement Impl Trait

detrumi:
Part of #335

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -390,6 +390,7 @@ impl LowerProgram for Program {
                                     .iter()
                                     .map(|b| b.lower(&empty_env))
                                     .collect::<Result<Vec<_>, _>>()?,
+                                ty: opaque_ty.ty.lower(&empty_env)?,
                             }),
                         );
                     }
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -239,7 +239,7 @@ impl LowerProgram for Program {
                     trait_ids.insert(type_kind.name, id);
                     trait_kinds.insert(id, type_kind);
                 }
-                Item::ImplTrait(_impl_trait) => todo!(),
+                Item::ImplTrait(_impl_trait) => continue,
                 Item::Impl(_) => continue,
                 Item::Clause(_) => continue,
             };
@@ -362,7 +362,7 @@ impl LowerProgram for Program {
                 Item::Clause(ref clause) => {
                     custom_clauses.extend(clause.lower_clause(&empty_env)?);
                 }
-                Item::ImplTrait(_) => todo!(),
+                Item::ImplTrait(ref _impl_trait) => todo!(),
             }
         }
 
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -1,14 +1,12 @@
 use chalk_ir::cast::{Cast, Caster};
 use chalk_ir::interner::ChalkIr;
 use chalk_ir::{
-    self, AssocTypeId, BoundVar, ClausePriority, DebruijnIndex, ImplId, ImplTraitId,
+    self, AssocTypeId, BoundVar, ClausePriority, DebruijnIndex, ImplId, OpaqueTyId,
     QuantifiedWhereClauses, StructId, Substitution, TraitId,
 };
 use chalk_parse::ast::*;
 use chalk_rust_ir as rust_ir;
-use chalk_rust_ir::{
-    Anonymize, AssociatedTyValueId, ImplTraitDatum, IntoWhereClauses, ToParameter,
-};
+use chalk_rust_ir::{Anonymize, AssociatedTyValueId, IntoWhereClauses, OpaqueTyDatum, ToParameter};
 use lalrpop_intern::intern;
 use std::collections::BTreeMap;
 use std::sync::Arc;
@@ -19,7 +17,7 @@ use crate::{Identifier as Ident, RawId, TypeKind, TypeSort};
 
 type StructIds = BTreeMap<Ident, chalk_ir::StructId<ChalkIr>>;
 type TraitIds = BTreeMap<Ident, chalk_ir::TraitId<ChalkIr>>;
-type ImplTraitIds = BTreeMap<Ident, chalk_ir::ImplTraitId<ChalkIr>>;
+type OpaqueTyIds = BTreeMap<Ident, chalk_ir::OpaqueTyId<ChalkIr>>;
 type StructKinds = BTreeMap<chalk_ir::StructId<ChalkIr>, TypeKind>;
 type TraitKinds = BTreeMap<chalk_ir::TraitId<ChalkIr>, TypeKind>;
 type AssociatedTyLookups = BTreeMap<(chalk_ir::TraitId<ChalkIr>, Ident), AssociatedTyLookup>;
@@ -35,7 +33,7 @@ struct Env<'k> {
     struct_kinds: &'k StructKinds,
     trait_ids: &'k TraitIds,
     trait_kinds: &'k TraitKinds,
-    impl_trait_ids: &'k ImplTraitIds,
+    opaque_ty_ids: &'k OpaqueTyIds,
     associated_ty_lookups: &'k AssociatedTyLookups,
     /// Parameter identifiers are used as keys, therefore
     /// all identifiers in an environment must be unique (no shadowing).
@@ -70,7 +68,7 @@ struct AssociatedTyLookup {
 enum TypeLookup {
     Struct(chalk_ir::StructId<ChalkIr>),
     Parameter(BoundVar),
-    ImplTrait(chalk_ir::ImplTraitId<ChalkIr>),
+    Opaque(chalk_ir::OpaqueTyId<ChalkIr>),
 }
 
 enum LifetimeLookup {
@@ -93,8 +91,8 @@ impl<'k> Env<'k> {
             return Ok(TypeLookup::Struct(*id));
         }
 
-        if let Some(id) = self.impl_trait_ids.get(&name.str) {
-            return Ok(TypeLookup::ImplTrait(*id));
+        if let Some(id) = self.opaque_ty_ids.get(&name.str) {
+            return Ok(TypeLookup::Opaque(*id));
         }
         if let Some(_) = self.trait_ids.get(&name.str) {
             return Err(RustIrError::NotStruct(name));
@@ -231,10 +229,10 @@ impl LowerProgram for Program {
 
         let mut struct_ids = BTreeMap::new();
         let mut trait_ids = BTreeMap::new();
-        let mut impl_trait_ids = BTreeMap::new();
+        let mut opaque_ty_ids = BTreeMap::new();
         let mut struct_kinds = BTreeMap::new();
         let mut trait_kinds = BTreeMap::new();
-        let mut impl_trait_kinds = BTreeMap::new();
+        let mut opaque_ty_kinds = BTreeMap::new();
         for (item, &raw_id) in self.items.iter().zip(&raw_ids) {
             match item {
                 Item::StructDefn(defn) => {
@@ -251,9 +249,9 @@ impl LowerProgram for Program {
                 }
                 Item::OpaqueTyDefn(defn) => {
                     let type_kind = defn.lower_type_kind()?;
-                    let id = ImplTraitId(raw_id);
-                    impl_trait_ids.insert(defn.identifier.str, id);
-                    impl_trait_kinds.insert(id, type_kind);
+                    let id = OpaqueTyId(raw_id);
+                    opaque_ty_ids.insert(defn.identifier.str, id);
+                    opaque_ty_kinds.insert(id, type_kind);
                 }
                 Item::Impl(_) => continue,
                 Item::Clause(_) => continue,
@@ -266,7 +264,7 @@ impl LowerProgram for Program {
         let mut impl_data = BTreeMap::new();
         let mut associated_ty_data = BTreeMap::new();
         let mut associated_ty_values = BTreeMap::new();
-        let mut impl_trait_data = BTreeMap::new();
+        let mut opaque_ty_data = BTreeMap::new();
         let mut custom_clauses = Vec::new();
         for (item, &raw_id) in self.items.iter().zip(&raw_ids) {
             let empty_env = Env {
@@ -274,7 +272,7 @@ impl LowerProgram for Program {
                 struct_kinds: &struct_kinds,
                 trait_ids: &trait_ids,
                 trait_kinds: &trait_kinds,
-                impl_trait_ids: &impl_trait_ids,
+                opaque_ty_ids: &opaque_ty_ids,
                 associated_ty_lookups: &associated_ty_lookups,
                 parameter_map: BTreeMap::new(),
             };
@@ -380,7 +378,7 @@ impl LowerProgram for Program {
                     custom_clauses.extend(clause.lower_clause(&empty_env)?);
                 }
                 Item::OpaqueTyDefn(ref opaque_ty) => {
-                    if let Some(&value) = impl_trait_ids.get(&opaque_ty.identifier.str) {
+                    if let Some(&value) = opaque_ty_ids.get(&opaque_ty.identifier.str) {
                         let parameter_kinds = opaque_ty
                             .parameter_kinds
                             .iter()
@@ -390,10 +388,10 @@ impl LowerProgram for Program {
                         let binders = empty_env
                             .in_binders(parameter_kinds, |env| opaque_ty.ty.lower(&env))?;
 
-                        impl_trait_data.insert(
+                        opaque_ty_data.insert(
                             value,
-                            Arc::new(ImplTraitDatum {
-                                impl_trait_id: value,
+                            Arc::new(OpaqueTyDatum {
+                                opaque_ty_id: value,
                                 bounds: opaque_ty
                                     .bounds
                                     .lower(&empty_env)?
@@ -401,8 +399,11 @@ impl LowerProgram for Program {
                                     .flat_map(|qil| {
                                         qil.into_where_clauses(
                                             empty_env.interner(),
-                                            chalk_ir::TyData::BoundVar(0)
-                                                .intern(empty_env.interner()),
+                                            chalk_ir::TyData::BoundVar(BoundVar::new(
+                                                DebruijnIndex::INNERMOST,
+                                                todo!(),
+                                            ))
+                                            .intern(empty_env.interner()),
                                         )
                                     })
                                     .collect(),
@@ -425,8 +426,8 @@ impl LowerProgram for Program {
             impl_data,
             associated_ty_values,
             associated_ty_data,
-            impl_trait_ids,
-            impl_trait_data,
+            opaque_ty_ids,
+            opaque_ty_data,
             custom_clauses,
         };
 
@@ -606,12 +607,12 @@ impl LowerTypeKind for TraitDefn {
 impl LowerTypeKind for OpaqueTyDefn {
     fn lower_type_kind(&self) -> LowerResult<TypeKind> {
         Ok(TypeKind {
-            sort: TypeSort::ImplTrait,
+            sort: TypeSort::Opaque,
             name: self.identifier.str,
-            binders: chalk_ir::Binders {
-                binders: vec![], //TODO do we need binders here?
-                value: (),
-            },
+            binders: chalk_ir::Binders::new(
+                vec![], //TODO do we need binders here?
+                (),
+            ),
         })
     }
 }
@@ -1059,12 +1060,12 @@ impl LowerTy for Ty {
                     }
                 }
                 TypeLookup::Parameter(d) => Ok(chalk_ir::TyData::BoundVar(d).intern(interner)),
-                TypeLookup::ImplTrait(id) => Ok(chalk_ir::TyData::Alias(
-                    chalk_ir::AliasTy::ImplTrait(chalk_ir::ImplTraitTy {
-                        impl_trait_id: id,
+                TypeLookup::Opaque(id) => Ok(chalk_ir::TyData::Alias(chalk_ir::AliasTy::Opaque(
+                    chalk_ir::OpaqueTy {
+                        opaque_ty_id: id,
                         substitution: chalk_ir::Substitution::empty(interner),
-                    }),
-                )
+                    },
+                ))
                 .intern(interner)),
             },
 
@@ -1094,7 +1095,7 @@ impl LowerTy for Ty {
             Ty::Apply { name, ref args } => {
                 let id = match env.lookup_type(name)? {
                     TypeLookup::Struct(id) => id,
-                    TypeLookup::Parameter(_) | TypeLookup::ImplTrait(_) => {
+                    TypeLookup::Parameter(_) | TypeLookup::Opaque(_) => {
                         Err(RustIrError::CannotApplyTypeParameter(name))?
                     }
                 };
@@ -1363,7 +1364,7 @@ impl LowerGoal<LoweredProgram> for Goal {
         let env = Env {
             struct_ids: &program.struct_ids,
             trait_ids: &program.trait_ids,
-            impl_trait_ids: &program.impl_trait_ids,
+            opaque_ty_ids: &program.opaque_ty_ids,
             struct_kinds: &program.struct_kinds,
             trait_kinds: &program.trait_kinds,
             associated_ty_lookups: &associated_ty_lookups,
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -1,12 +1,14 @@
 use chalk_ir::cast::{Cast, Caster};
 use chalk_ir::interner::ChalkIr;
 use chalk_ir::{
-    self, AssocTypeId, BoundVar, ClausePriority, DebruijnIndex, ImplId, QuantifiedWhereClauses,
-    StructId, Substitution, TraitId,
+    self, AssocTypeId, BoundVar, ClausePriority, DebruijnIndex, ImplId, ImplTraitId,
+    QuantifiedWhereClauses, StructId, Substitution, TraitId,
 };
 use chalk_parse::ast::*;
 use chalk_rust_ir as rust_ir;
-use chalk_rust_ir::{Anonymize, AssociatedTyValueId, IntoWhereClauses, ToParameter};
+use chalk_rust_ir::{
+    Anonymize, AssociatedTyValueId, ImplTraitValue, IntoWhereClauses, ToParameter,
+};
 use lalrpop_intern::intern;
 use std::collections::BTreeMap;
 use std::sync::Arc;
@@ -225,6 +227,7 @@ impl LowerProgram for Program {
         let mut trait_ids = BTreeMap::new();
         let mut struct_kinds = BTreeMap::new();
         let mut trait_kinds = BTreeMap::new();
+        let mut impl_trait_ids = BTreeMap::new();
         for (item, &raw_id) in self.items.iter().zip(&raw_ids) {
             match item {
                 Item::StructDefn(defn) => {
@@ -239,7 +242,9 @@ impl LowerProgram for Program {
                     trait_ids.insert(type_kind.name, id);
                     trait_kinds.insert(id, type_kind);
                 }
-                Item::ImplTrait(_impl_trait) => continue,
+                Item::ImplTrait(impl_trait) => {
+                    impl_trait_ids.insert(impl_trait.identifier.str, ImplTraitId(raw_id));
+                }
                 Item::Impl(_) => continue,
                 Item::Clause(_) => continue,
             };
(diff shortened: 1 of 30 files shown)


jackh726:
I'll definitely look over this later, but can you add the tests removed in https://github.com/rust-lang/chalk/commit/6252238d7ddd1e07602e63ccaff4802a08ea0dc5?

detrumi:
> I'll definitely look over this later, but can you add the tests removed in [6252238](https://github.com/rust-lang/chalk/commit/6252238d7ddd1e07602e63ccaff4802a08ea0dc5)?

Thanks, that's really helpful! (wasn't aware that we had those tests)
Note that it's a bit early for a review, there's still a lot of todo's and some parts that need to change (I misunderstood how the goals would look like for example, which is why those tests are so useful)

chalk-ir/src/lib.rs
```diff
@@ -569,17 +576,23 @@ impl<TF: TypeFamily> ParameterData<TF> {
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Fold, HasTypeFamily)]
-pub struct ProjectionTy<TF: TypeFamily> {
-    pub associated_ty_id: AssocTypeId<TF>,
-    pub substitution: Substitution<TF>,
+pub enum AliasTy<TF: TypeFamily> {
+    Projection(ProjectionTy<TF>),
+    ImplTrait(ImplTraitId<TF>),
```
flodiebold: wouldn't the impl Trait need to be parameterized similar to the projection? E.g. in a situation like `foo<T>() -> impl Trait`, the `impl Trait` is really parameterized by the `T`, right? See for example [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5dc9ff4597b49fc3eb5d6ec40a9b5e6c).

(Now I'm wondering how `foo(impl Trait) -> impl Trait` works... does it rely on rustc desugaring the argument impl Trait to a type parameter?)

flodiebold: (Thinking a bit more about that, probably yes -- the APIT from outside the function, and the RPIT from inside the function, can't really be represented using `ImplTrait` anyway...)

detrumi: Good call, I've added substitutions.

nikomatsakis:
Needs rebase now, @detrumi =)

chalk-solve/src/lib.rs
```diff
@@ -36,6 +36,9 @@ pub trait RustIrDatabase<I: Interner>: Debug {
     /// Returns the `AssociatedTyValue` with the given id.
     fn associated_ty_value(&self, id: AssociatedTyValueId<I>) -> Arc<AssociatedTyValue<I>>;
 
+    /// Returns the `ImplTraitDatum` with the given id.
+    fn impl_trait_data(&self, id: ImplTraitId<I>) -> Arc<ImplTraitDatum<I>>;
```
nikomatsakis: We don't have to do it in this PR, but I think eventually we will want to add a callback here like

```rust
fn hidden_type(&self, id: ImplTraitId<I>) -> Ty<I>;
```

chalk-rust-ir/src/lib.rs
```diff
@@ -422,6 +422,23 @@ pub struct AssociatedTyValueBound<I: Interner> {
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold)]
+pub struct ImplTraitDatum<I: Interner> {
+    /// The placeholder `!T` that corresponds to the opaque type `T`.
+    pub impl_trait_id: ImplTraitId<I>,
+
+    /// Trait bounds for the opaque type.
+    pub bounds: Vec<QuantifiedWhereClause<I>>,
+
+    /// The "hidden type" that the opaque type is equal to when revealed.
+    pub ty: Binders<Ty<I>>,
```
nikomatsakis: let's call this `hidden_ty` (but, as I wrote, I think we'll want to remove it eventually)

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,70 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
     }
 }
 
+impl<I: Interner> ToProgramClauses<I> for ImplTraitDatum<I> {
+    /// Given `opaque type T<..>: A + B = HiddenTy;`, we generate:
+    ///
+    /// ```notrust
+    /// AliasEq(T<..> = HiddenTy) :- Reveal.
+    /// AliasEq(T<..> = !T).
```
nikomatsakis: Nit: `!T` has type parameters, too, so maybe write `= !T<...>`

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,70 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+    /// ```
+    /// where `!T` is the placeholder for the unnormalized type `T<..>`.
+    fn to_program_clauses(&self, builder: &mut ClauseBuilder<'_, I>) {
+        let interner = builder.db.interner();
+        let alias = AliasTy::ImplTrait(ImplTraitTy {
+            impl_trait_id: self.impl_trait_id,
+            substitution: builder.substitution_in_scope(),
+        });
+
+        let alias_ty = Ty::new(interner, alias.clone());
```
nikomatsakis: Hmm, I expected (I think) that we would treat this analogously to other placeholders -- though I've gone a bit back and forth on how that should be. Currently, though, we have a [special variant of `TypeName`](http://rust-lang.github.io/chalk/chalk_ir/enum.TypeName.html#variant.AssociatedType) for the placeholder case. I think maybe we want a variant of `TypeName` also for impl trait placeholders.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,70 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+                .cast(interner),
+            ));
+        });
+
+        for bound in &self.bounds {
+            // Implemented(!T: Bound).
+            builder.push_fact(bound.value.clone().into_well_formed_goal(interner));
+        }
+
+        for auto_trait_id in builder.db.auto_traits() {
```
nikomatsakis: If we made the "hidden type" lazy, then we'd want to hold off on generating these rules unless we are asked to solve `Implemented(!T: AutoTrait)` where `AutoTrait` is, well, an auto-trait, and `!T` is some kind of impl trait placeholder type. But that would be partly done by modiying the `to_program_clauses` callback.

chalk-rust-ir/src/lib.rs
```diff
@@ -422,6 +422,23 @@ pub struct AssociatedTyValueBound<I: Interner> {
...
+/// ```ignore
+/// opaque type T: A + B = HiddenTy;
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold)]
+pub struct ImplTraitDatum<I: Interner> {
+    /// The placeholder `!T` that corresponds to the opaque type `T`.
+    pub impl_trait_id: ImplTraitId<I>,
+
+    /// Trait bounds for the opaque type.
+    pub bounds: Vec<QuantifiedWhereClause<I>>,
```
nikomatsakis: So, in the example of `T: A + B` above, what would the value of these where-clauses be? In particular, how is the `T` represented? My suggestion would be that we want to use a "binder" here for the 'self-type', kind of like we do for `dyn Write` or whatever. This is because sometimes we'll want to instantiate the where-clauses with the placeholder `!T` and sometimes (perhaps in follow-up PRs...) we'll want to instantiate those same clauses with the hidden type that is being inferred, etc.

chalk-solve/src/solve/slg/aggregate.rs
```diff
@@ -229,9 +229,15 @@ impl<I: Interner> AntiUnifier<'_, '_, I> {
                 self.aggregate_application_tys(apply1, apply2)
             }
 
-            (TyData::Alias(alias1), TyData::Alias(alias2)) => {
-                self.aggregate_alias_tys(alias1, alias2)
-            }
+            (
+                TyData::Alias(AliasTy::Projection(proj1)),
+                TyData::Alias(AliasTy::Projection(proj2)),
+            ) => self.aggregate_projection_tys(proj1, proj2),
```
nikomatsakis: I wonder if -- before this PR, at least -- this code was even live. I think we never *generate* "alias types" as part of an answer (only their associated placeholder), though we've considered dropping that invariant from time to time.

tests/test/existential_types.rs
```diff
@@ -2,6 +2,33 @@
...
+                T: Trait
+            }
+        } yields {
+            "Unique; substitution []"
+        }
+
+        goal {
+            T: Trait
+        } yields {
+            "No possible solution"
```
nikomatsakis: ok this doesn't seem right =) shouldn't this be solveable, since the bounds say `T: Trait`?

This could well be related to the 'placeholder' thing I was talking about above

chalk-solve/src/clauses/program_clauses.rs
nikomatsakis: We want to do this in part because we want unification to treat the placeholder differently (it shouldn't try to prove `AliasEq` if we're using the placeholder, it should only be equal to other placeholders)

chalk-integration/src/lowering.rs
```diff
@@ -236,6 +246,12 @@ impl LowerProgram for Program {
                     trait_ids.insert(type_kind.name, id);
                     trait_kinds.insert(id, type_kind);
                 }
+                Item::OpaqueTyDefn(defn) => {
```
nikomatsakis: We seem a bit inconsistent in whether we are calling this "opaque ty" or "impl trait" here -- is there a reason to use different terms in different places? I'm somewhat indifferent about which we use. I have a vague preference for `impl Trait` (more tightly tied to Rust syntax) but then rustc is using (I think) opaque type.

chalk-integration/src/lowering.rs
```diff
@@ -354,6 +372,38 @@ impl LowerProgram for Program {
...
+                            Arc::new(ImplTraitDatum {
+                                impl_trait_id: value,
+                                bounds: opaque_ty
+                                    .bounds
+                                    .lower(&empty_env)?
+                                    .iter()
+                                    .flat_map(|qil| {
+                                        qil.into_where_clauses(
+                                            empty_env.interner(),
+                                            chalk_ir::TyData::BoundVar(0)
```
nikomatsakis: Something seems suspicious here:

(a) I expect these where clauses to be lowering inside of the `parameter_kinds` binder, first off. This is because you may have something like

```rust
type Foo<T> = impl Iterator<Item = T>
```

(b) Using `BoundVar` here seems a bit surprising because we're kind of missing a *binder*. If we were taking the approach I suggested above, though, of indeed adding a `BoundVar` to represent "the hidden type" (or "the opaque type placeholder" as case may be), that would be an additional binder we push on. Hmm, we should probably talk this out a bit more on Zulip.

chalk-rust-ir/src/lib.rs
nikomatsakis: Oh, also, I think we really want something like:

```rust
 pub struct ImplTraitDatum<I: Interner> { 
     pub impl_trait_id: ImplTraitId<I>, 
     pub bound: Binders<ImplTraitDatumBound<I>>,
}
```

where `ImplTraitDatumBound` contains the where clauses + the hidden type. The reason is that both of them may reference the parameter kinds:

```rust
type Foo<T> = impl Iterator<Item = T> // hidden type = Vec<T>
```

chalk-integration/src/lowering.rs
detrumi: Yeah, we chose `opaque type` syntax in the AST, so I'm using that around the ast types. Would you prefer it to be called ImplTrait in the code everywhere, while still using `opaque type` syntax?

detrumi: After discussing this, we decided to go with "opaque type" in most places, and also rename ImplTraitDatum to OpaqueTyDatum, for example

tests/test/existential_types.rs
detrumi: Right, I've updated the test. But what about a situation where we have a hidden type that doesn't actually implement the bound trait? Do we just assume it?

detrumi: nvm, that's what the WF rules are for of course

detrumi:
@nikomatsakis Could you make some time to do the review you were planning to do?

nikomatsakis:
Gah @detrumi sorry I forgot! Will try to do so now.

chalk-integration/src/lowering.rs
```diff
@@ -357,6 +376,52 @@ impl LowerProgram for Program {
...
                 }
+                Item::OpaqueTyDefn(ref opaque_ty) => {
+                    if let Some(&opaque_ty_id) = opaque_ty_ids.get(&opaque_ty.identifier.str) {
+                        let parameter_kinds = opaque_ty
+                            .parameter_kinds
+                            .iter()
+                            .map(|k| k.lower())
+                            .collect::<Vec<_>>();
+
+                        let binders = empty_env.in_binders(parameter_kinds, |env| {
```
nikomatsakis, line 387: We should probably add some comments. Maybe:

---

// Introduce the parameters declared on the opaque type definition. So if we have `type Foo<P1..Pn> = impl Trait<T1..Tn>`, this would introduce `P1..Pn`

chalk-integration/src/lowering.rs
```diff
@@ -357,6 +376,52 @@ impl LowerProgram for Program {
...
+                            let hidden_ty = opaque_ty.ty.lower(&env)?;
+
+                            let bounds: chalk_ir::Binders<Vec<chalk_ir::Binders<_>>> = env
```
nikomatsakis, line 390: // Introduce a variable to represent the hidden "self type". This will be used in the bounds. So the `impl Trait<T1..Tn>` will be lowered to `exists<Self> { Self: Trait<T1..Tn> }`.

chalk-integration/src/lowering.rs
```diff
@@ -357,6 +376,52 @@ impl LowerProgram for Program {
...
+                                .in_binders(
+                                    Some(chalk_ir::ParameterKind::Ty(intern(FIXME_SELF))),
+                                    |env1| {
+                                        let interner = env1.interner();
+                                        Ok(opaque_ty
+                                            .bounds
+                                            .lower(&env1)?
+                                            .iter()
+                                            .flat_map(|qil| {
+                                                qil.into_where_clauses(
```
nikomatsakis, line 400: // Instantiate the bounds with the innermost bound variable, which represents `Self`, as the self type.

chalk-ir/src/zip.rs
```diff
@@ -233,17 +234,21 @@ struct_zip!(impl[
...
     conditions
 });
+struct_zip!(impl[I: Interner] Zip<I> for ProjectionTy<I> {
+    associated_ty_id,
+    substitution
+});
+struct_zip!(impl[I: Interner] Zip<I> for OpaqueTy<I> {
+    opaque_ty_id,
+    substitution
+});
```
nikomatsakis, line 251: heh we should write a derive for this too!

chalk-solve/src/clauses.rs
```diff
@@ -262,31 +266,35 @@ fn program_clauses_that_could_match<I: Interner>(
...
+                let associated_ty_datum = db.associated_ty_data(proj.associated_ty_id);
+                let trait_id = associated_ty_datum.trait_id;
+                let trait_parameters = db.trait_parameters_from_projection(proj);
+                push_program_clauses_for_associated_type_values_in_impls_of(
+                    builder,
+                    trait_id,
+                    trait_parameters,
+                );
+            }
+            AliasTy::Opaque(_) => (),
```
nikomatsakis, line 291: It seems like we're missing something here. I would expect if we have the goal

```rust
Normalize(OpaqueTy -> ...)
```

that we would generate clauses from the opaque type definition to the hidden type (but only if `Reveal` is in the environment).

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+        let alias_ty = Ty::new(
+            interner,
+            ApplicationTy {
+                name: TypeName::OpaqueType(self.opaque_ty_id),
+                substitution,
+            },
+        );
+
+        builder.push_binders(&self.bound, |builder, opaque_ty_bound| {
+            // AliasEq(T<..> = HiddenTy) :- Reveal.
```
nikomatsakis, line 149: Ah, I see why you didn't need the normalize rule. I think I expected this to be `Normalize(...)` and to have a general rule `AliasEq(A = B) :- Normalize(A -> B)`, but I'm not really sure that it adds anything to do the separation, and we could always do it later.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+    /// AliasEq(T<..> = HiddenTy) :- Reveal.
+    /// AliasEq(T<..> = !T<..>).
+    /// Implemented(!T<..>: A).
+    /// Implemented(!T<..>: B).
+    /// Implemented(!T<..>: Send) :- Implemented(HiddenTy: Send). // For all auto traits
+    /// ```
+    /// where `!T<..>` is the placeholder for the unnormalized type `T<..>`.
+    fn to_program_clauses(&self, builder: &mut ClauseBuilder<'_, I>) {
+        let interner = builder.interner();
+        let substitution = builder.substitution_in_scope();
```
nikomatsakis, line 134: I suspect this is incorrect. When this is invoked, there won't yet *be* any substitution in scope, you need to (I think) push the binders first.

So move this logic *inside* the call to `push_binders`  below.

In other words, you have `type Foo<P1...Pn> = impl Trait<T1..Tn> = HiddenTy` or whatever, and you want to make some rules like

```
forall<P1..Pn> { // <-- created by the push-binders call
    AliasEq(Foo<P1..Pn> = HiddenTy), // references those P1..Pn
}
```

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+            builder.push_fact(DomainGoal::Holds(
+                AliasEq {
+                    alias: alias.clone(),
+                    ty: alias_ty.clone(),
+                }
+                .cast(interner),
+            ));
+
+            for bound in &opaque_ty_bound.bounds {
+                // Implemented(!T<..>: Bound).
```
nikomatsakis, line 171: PS, thanks for the comments! Even minimal stuff like this really helps to make it easier to track what's going on.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+    /// where `!T<..>` is the placeholder for the unnormalized type `T<..>`.
+    fn to_program_clauses(&self, builder: &mut ClauseBuilder<'_, I>) {
+        let interner = builder.interner();
+        let substitution = builder.substitution_in_scope();
+        let alias = AliasTy::Opaque(OpaqueTy {
+            opaque_ty_id: self.opaque_ty_id,
+            substitution: substitution.clone(),
+        });
+
+        let alias_ty = Ty::new(
```
nikomatsakis, line 140: can we rename this to `alias_placeholder_ty`?

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+                AliasEq {
+                    alias: alias.clone(),
+                    ty: alias_ty.clone(),
+                }
+                .cast(interner),
+            ));
+
+            for bound in &opaque_ty_bound.bounds {
+                // Implemented(!T<..>: Bound).
+                builder.push_binders(&bound, |builder, bound| {
```
nikomatsakis, line 172: This I think is not quite right either. These binders represent the hidden self-type, right? I think they want to be *instantiated* with the placeholder. That is, right now given (say) `type Foo<P1..Pn> = impl Trait<T1..Tn>` you're generating something like

```
forall<P1..PN> { // generated by the first push_binders
    forall<X> { // generated by the second push_binders
        forall<> { // generated by the 3rd push_binders; we don't have `for<'a>` or anything, so empty
            X: Trait
        }
    }
}
```

instead I think what you want is to do 

```rust
for bound in &opaque_ty_bound.bounds {
  // Implemented(!T<..>: Bound).
  let bound_with_placeholder_ty = bound.substitute(interner, &[alias_placeholder_ty]);
  builder.push_binders(&bound_with_placeholder_ty , |builder, bound| {
    builder.push_fact(bound.into_well_formed_goal(interner));
  });
}
```

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,84 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+                    builder.push_binders(&bound, |builder, bound| {
+                        builder.push_fact(bound.into_well_formed_goal(interner));
+                    });
+                });
+            }
+
+            for auto_trait_id in builder.db.auto_traits() {
+                // Implemented(!T<..>: AutoTrait) :- Implemented(HiddenTy: AutoTrait).
```
nikomatsakis, line 180: This looks right

chalk-solve/src/clauses.rs
nikomatsakis: (Never mind, I figured this out later)

chalk-solve/src/clauses/program_clauses.rs
detrumi: Oh right, all these lines can be moved inside the `push_binders`. :+1:

nikomatsakis:
I think the remaining test failures are because you have to modify the code that generates 'which program clauses might implement a given goal'.

In particular, when you run the test, you are asked to prove;

```
goal { T : Clone }
using solver: SLG { max_size: 10, expected_answers: None }
| creating new table TableIndex(0) and goal UCanonical {
:     canonical: Canonical {
:         value: InEnvironment {
:             environment: Env([]),
:             goal: Implemented(impl for<> OpaqueTyDatumBound { hidden_ty: Ty, bounds: for<type> [for<> Implemented(^1.0: Clone)] }: Clone),
:         },
:         binders: [],
:     },
:     universes: 1,
: } {
```

that winds up in this match arm:

https://github.com/rust-lang/chalk/blob/48664238ade731830dc263ba3ce08ecb1d30c8a2/chalk-solve/src/clauses.rs#L147

this has logic for ["things derived from traits and impls"](https://github.com/rust-lang/chalk/blob/48664238ade731830dc263ba3ce08ecb1d30c8a2/chalk-solve/src/clauses.rs#L150-L159) and ["things where dyn Trait is the self type"](https://github.com/rust-lang/chalk/blob/48664238ade731830dc263ba3ce08ecb1d30c8a2/chalk-solve/src/clauses.rs#L238) but we need some logic for "cases where an alias placeholder is the self type".

Something like:

```rust
if let /* AliasPlaceholder */ = self_ty.data(interner) {
    db.opaque_type(...).to_program_clauses(...);
}
```

(BTW, I find the debug output for impl trait types (`OpaqueTyDatumBound { hidden_ty: Ty, bounds: for<type> [for<> Implemented(^1.0: Clone)] }`) rather confusing -- I guess it's looking up the definition? I think I'd prefer to see just the name of the impl trait, like `Foo`.)

chalk-solve/src/clauses/program_clauses.rs
detrumi: Created #397 for that

nikomatsakis:
@detrumi I guess this is waiting on a rebase and then perhaps some more tests? 

It'd be good to create an updated list of what's left to do.

I forget if we decided to defer handling of things generic opaque types like `type Foo<X> = impl Iterator<Item = X>`? It seems like a lot of the code is there for that already.

detrumi:
> @detrumi I guess this is waiting on a rebase and then perhaps some more tests?
> 
> It'd be good to create an updated list of what's left to do.
> 
> I forget if we decided to defer handling of things generic opaque types like `type Foo<X> = impl Iterator<Item = X>`? It seems like a lot of the code is there for that already.

Yes, I started on the rebase but quite a lot of things changed last week, so that takes some work :sweat_smile: 

For what's left, the list in #335 is still fairly up-to-date, the only additional points are adding documentation and more tests, which I'd like to do in separate PRs.

detrumi:
@nikomatsakis Should be ready to merge now!

nikomatsakis:
@detrumi heh, sorry, one more rebase? ❤️

detrumi:
> @detrumi heh, sorry, one more rebase? heart

@nikomatsakis Done!

chalk-integration/src/program.rs
```diff
@@ -320,4 +363,12 @@ impl RustIrDatabase<ChalkIr> for Program {
     fn interner(&self) -> &ChalkIr {
         &ChalkIr
     }
+
+    fn auto_traits(&self) -> Vec<TraitId<ChalkIr>> {
```
nikomatsakis, line 367: Hmm, I didn't notice this before. I don't think we want this method...

chalk-solve/src/lib.rs
```diff
@@ -85,6 +88,9 @@ pub trait RustIrDatabase<I: Interner>: Debug {
     fn well_known_trait_id(&self, well_known_trait: WellKnownTrait) -> Option<TraitId<I>>;
 
     fn interner(&self) -> &I;
+
+    /// Returns the ids of all auto traits.
+    fn auto_traits(&self) -> Vec<TraitId<I>>;
```
nikomatsakis, line 93: Hmm, I didn't notice this before. I don't think we can land with this method... we don't want to require clients to enumerate all traits that exist.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -118,6 +118,85 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+            let substitution = Substitution::from1(interner, alias_placeholder_ty.clone());
+            for bound in &opaque_ty_bound.bounds {
+                // Implemented(!T<..>: Bound).
+                let bound_with_placeholder_ty = bound.substitute(interner, &substitution);
+                builder.push_binders(&bound_with_placeholder_ty, |builder, bound| {
+                    builder.push_fact(bound);
+                });
+            }
+
+            for auto_trait_id in builder.db.auto_traits() {
```
nikomatsakis, line 180: Since we need to get rid of this method, what I think we should be doing instead is to look and check:

If we are generate clauses for a goal like 

```rust
Implemented(!T: Foo)
```

where `!T` is an alias placeholder *and* `Foo` is an auto-trait, then we generate these clauses, and otherwise we don't.

The `to_program_clauses` trait doesn't work for this, but we can just a free function or another extension trait.

chalk-integration/src/program.rs
```diff
@@ -42,6 +42,12 @@ pub struct Program {
     pub associated_ty_values:
         BTreeMap<AssociatedTyValueId<ChalkIr>, Arc<AssociatedTyValue<ChalkIr>>>,
 
+    // From opaque type name to item-id. Used during lowering only.
+    pub opaque_ty_ids: BTreeMap<Identifier, OpaqueTyId<ChalkIr>>,
+
+    /// For each opaque type:
+    pub opaque_ty_data: BTreeMap<OpaqueTyId<ChalkIr>, Arc<OpaqueTyDatum<ChalkIr>>>,
```
nikomatsakis, line 49: This can be done in a follow-up PR, but I think we should add a `opaque_names: BTreeMap<TraitId<ChalkIr>, Identifier>` or something like that, created during lowering...

chalk-integration/src/program.rs
```diff
@@ -126,6 +157,14 @@ impl tls::DebugContext for Program {
         )
     }
 
+    fn debug_opaque_ty(
+        &self,
+        opaque_ty: &OpaqueTy<ChalkIr>,
+        fmt: &mut fmt::Formatter<'_>,
+    ) -> Result<(), fmt::Error> {
+        write!(fmt, "impl {:?}", opaque_ty.opaque_ty_id)
```
nikomatsakis, line 165: ...and then reference the `opaque_type_names` here to print the *name* of the opaque type, and not "impl"

chalk-integration/src/program.rs
```diff
@@ -100,21 +106,46 @@ impl tls::DebugContext for Program {
...
         }
     }
 
+    fn debug_opaque_ty_id(
+        &self,
+        opaque_ty_id: OpaqueTyId<ChalkIr>,
+        fmt: &mut fmt::Formatter<'_>,
+    ) -> Result<(), fmt::Error> {
+        if let Some(d) = self.opaque_ty_data.get(&opaque_ty_id) {
+            write!(fmt, "{:?}", d.bound.skip_binders().hidden_ty)
```
nikomatsakis, line 124: ...or maybe here we want to print the name? anyway, somewhere. It'd be nice for opaque types to print out with their name and for the placeholders to print with `!Foo`.

chalk-solve/src/clauses/program_clauses.rs
detrumi: Right, so we generate such clauses only when there's such a goal. Somewhere around `push_auto_trait_impls` probably?

Anyways, I'll just remove this part for now, I'll add it to the list for next steps.

PR #325, opened 2020-02-06
Goals and folding goals

nikomatsakis:
Refactors how the Folder is setup:

* Simpler trait setup: Instead of having a ton of small traits, we have one larger trait with default methods
* Documented, at least somewhat, in the chalk book
* Now permits "overriding" the handling of goals and clauses

The motivation for this work is to work towards the "syntactic equality" goal, where we desugar aliases into a combination of an inference variable and an alias goal ahead of time, rather than building "rich type equality" into the solver itself.


chalk-ir/src/fold.rs
```diff
@@ -376,3 +314,70 @@ impl<TF: TypeFamily, TTF: TargetTypeFamily<TF>> Fold<TF, TTF> for Lifetime<TF> {
...
+                folder.fold_free_placeholder_lifetime(*universe, binders)
+            }
+            LifetimeData::Phantom(..) => unreachable!(),
+        }
+    }
+}
+
+/// Folding a goal invokes the `fold_goal` callback (which will, by
+/// default, invoke super-fold).
+impl<TF: TypeFamily, TTF: TargetTypeFamily<TF>> Fold<TF, TTF> for Goal<TF> {
```
jackh726: Why remove this from `boring_impls`?

chalk-ir/src/lib.rs
```diff
@@ -1013,6 +1010,52 @@ impl<T> UCanonical<T> {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasTypeFamily)]
+/// A general goal; this is the full range of questions you can pose to Chalk.
```
jackh726: Not the correct doc

chalk-ir/src/lib.rs
```diff
@@ -1013,6 +1010,52 @@ impl<T> UCanonical<T> {
...
+
+impl<TF: TypeFamily> Goals<TF> {
+    pub fn from(goals: impl IntoIterator<Item = impl CastTo<Goal<TF>>>) -> Self {
+        use crate::cast::Caster;
+        Goals {
+            goals: goals.into_iter().casted().collect(),
+        }
+    }
+
+    pub fn new() -> Self {
```
jackh726: Uh, super nit, but I would put this above `from`

chalk-ir/src/lib.rs
```diff
@@ -1013,6 +1010,52 @@ impl<T> UCanonical<T> {
...
+        Self::from(None::<Goal<TF>>)
```
jackh726: This works, but I think I'd prefer the straightforward `Goals { goals: vec![] }`

chalk-ir/src/fold.rs
```diff
@@ -62,205 +62,150 @@ pub use self::subst::Subst;
...
-{
+pub trait Folder<TF: TypeFamily, TTF: TargetTypeFamily<TF> = TF> {
+    /// Creates a `dyn` value from this folder. Unfortunately, this
+    /// must be added manually to each impl of Folder; it permits the
+    /// default implements below to create a `&mut dyn Folder` from
+    /// `Self` without knowing what `Self` is (by invoking this
+    /// method). Effectively, this limits impls of `Folder` to types
+    /// for which we are able to create a dyn value (i.e., not `[T]`
+    /// types).
+    fn as_dyn(&mut self) -> &mut dyn Folder<TF, TTF>;
```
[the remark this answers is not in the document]
nikomatsakis: Ah, yeah, forgot about that. I tried for a while to work around it but all my workarounds were worse. We could of course just use `impl Trait` instead of `dyn Trait`, I suppose, but I"m trying to avoid that in an effort to improve compilation times.

chalk-ir/src/fold.rs
nikomatsakis: It's not boring. It invokes a callback on folder rather than just recursively reproducing its structure.

chalk-ir/src/lib.rs
nikomatsakis: I didn't do that because I intend to refactor this more later and make it interned. I should maybe do that in this same PR though, forgot I took the first steps.

chalk-derive/src/lib.rs
```diff
@@ -147,6 +147,140 @@ pub fn derive_fold(item: TokenStream) -> TokenStream {
     panic!("derive(Fold) requires a parameter that implements HasTypeFamily or TypeFamily");
 }
 
+/// Derives Fold for structs and enums for which one of the following is true:
+/// - It has a `#[has_type_family(TheTypeFamily)]` attribute
+/// - There is a single parameter `T: HasTypeFamily` (does not have to be named `T`)
+/// - There is a single parameter `TF: TypeFamily` (does not have to be named `TF`)
+#[proc_macro_derive(SuperFold, attributes(has_type_family))]
+pub fn derive_super_fold(item: TokenStream) -> TokenStream {
```
jackh726: So I was looking at this again. The only place this is used is on `ProgramClause`. Do you think this will get used in the future elsewhere? Tbh, it's a lot of complexity for a single derive.

nikomatsakis: Hmm, good point. I think I originally expected to use it on `Goal` and `DomainGoal` but for various reasons didn't wind up doing so. One reason in particular that is pretty broadly applicable is that the `Goal<'tcx>` / `GoalData<'tcx>` split means that "super fold" on `Goal` is effectively just *fold* on `GoalData`. I think we could remove it.

it's also true that this code is largely copy and paste from the other fold definition, so it could probably be collapsed.

chalk-ir/src/lib.rs
nikomatsakis: Did the rest of the refactoring now.

chalk-derive/src/lib.rs
nikomatsakis: Removed the `SuperFold` derive.

chalk-ir/src/fold.rs
nikomatsakis: Maybe this naming scheme isn't obvious to everyone else though. :)

nikomatsakis:
OK, @jackh726, I think I addressed all your comments.

chalk-ir/src/fold.rs
jackh726: Yeah, my thought for "boring" is either boilerplate-like code or code that is extremely simple.

jackh726:
Yeah, LGTM

Issue #328, opened 2020-02-14
Extend `TypeFamily` with some context


nikomatsakis:
The various `intern` methods on `TypeFamily` do not currently permit any context. See e.g. [`intern_ty`]. We want that to take an `&self` parameter. This is going to be a somewhat extensive refactoring, however, because it's going to require threading a value around so that we can invoke the methods on it.

[`intern_ty`]: http://rust-lang.github.io/chalk/chalk_ir/family/trait.TypeFamily.html#tymethod.intern_ty

nikomatsakis:
Proposed design to start is this. We will move from

```rust
trait TypeFamily {
    type InternedType;
    ...

    fn intern_ty(data: TyData<Self>) -> Self::InternedType;
    fn ty_data(ty: &Self::InternedType) -> &TyData<Self>;
}
```

to

```rust
trait TypeFamily {
    type InternedType;
    ...

    fn intern_ty(&self, data: TyData<Self>) -> Self::InternedType;
    fn ty_data(ty: &Self::InternedType) -> &TyData<Self>;
}
```

Note that I left `ty_data` alone -- I do expect to add `&self` there too, but not as part of *this* issue. We'll do that next. Also, I only talked about types, but eventually we'll want to add this to all the other sorts of interned things. But the first step will be the hardest.

yaahc:
Indirectly related but we renamed TypeFamily to Interner here

https://github.com/rust-lang/chalk/pull/329

nikomatsakis:
This was done in #330

PR #330, opened 2020-02-14
Prepare for adding context to the Interner

yaahc:
part of #328

### DIFF
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -119,6 +119,9 @@ pub trait Context: Clone + Debug {
     /// the equality relation.
     type Variance;
 
+    /// The type used to store concrete representations of "core types" from chalk-ir.
+    type Interner;
+
     /// Given an environment and a goal, glue them together to create
     /// a `GoalInEnvironment`.
     fn goal_in_environment(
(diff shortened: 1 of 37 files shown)


yaahc:
The current puzzle is how best to access an `&I` from within the various functions on `SlgContext<I>`

```
error[E0061]: this function takes 3 parameters but 2 parameters were supplied
   --> chalk-solve/src/solve/slg.rs:164:37
    |
164 |           let (mut infer, subst, _) = InferenceTable::from_canonical(goal.universes, &goal.canonical);
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 3 parameters
    |
   ::: chalk-solve/src/infer.rs:49:5
    |
49  | /     pub(crate) fn from_canonical<T>(
50  | |         interner: &I,
51  | |         num_universes: usize,
52  | |         canonical: &Canonical<T>,
...   |
67  | |         (table, subst, value)
68  | |     }
    | |_____- defined here

```

chalk-integration/src/db.rs
```diff
@@ -54,7 +54,7 @@ impl ChalkDatabase {
         tls::set_current_program(&program, || op(&program))
     }
 
-    pub fn parse_and_lower_goal(&self, text: &str) -> Result<Goal<ChalkIr>, ChalkError> {
+    pub fn parse_and_lower_goal(&self, interner: &ChalkIr, text: &str) -> Result<Goal<ChalkIr>, ChalkError> {
```
jackh726: Is this used?

chalk-engine/src/context.rs
```diff
@@ -119,6 +119,8 @@ pub trait Context: Clone + Debug {
     /// the equality relation.
     type Variance;
 
+    type Interner;
```
jackh726: If you could add a short doc comment, that would be good

chalk-integration/src/query.rs
```diff
@@ -28,7 +28,7 @@ pub trait LoweringDatabase: RustIrDatabase<ChalkIr> {
     #[salsa::input]
     fn solver_choice(&self) -> SolverChoice;
 
-    fn program_ir(&self) -> Result<Arc<Program>, ChalkError>;
+    fn program_ir<'a>(&self, interner: ChalkIr) -> Result<Arc<Program>, ChalkError>;
```
jackh726: I think this lifetime is uneeded.

chalk-ir/src/lib.rs
```diff
@@ -811,6 +811,12 @@ pub struct AliasEq<I: Interner> {
     pub ty: Ty<I>,
 }
 
+impl<I: Interner> AliasEq<I> {
+    pub fn interner(&self) -> &I {
```
jackh726: Yeah, I don't think this function makes sense here.

yaahc: It has to either come from the AliasEq or it has to be part of the trait and no other implementor of this trait uses the interner yet.

jackh726: @detrumi is working the `impl Trait`/`Alias`, so this portion of the API might change. But regardless, even if the other types don't use it and we still need it, I don't think having `AliasEq` store the interner is correct.

PR #331, opened 2020-02-15
Don't return definite guidance if we find one solution and then flounder

flodiebold:
If we find one solution and then flounder, this previously resulted in a `Guidance::Definite` result. I think the reason is that `any_future_answer` has no answers to check and hence returns true.

### DIFF
--- a/chalk-solve/src/solve/slg/aggregate.rs
+++ b/chalk-solve/src/solve/slg/aggregate.rs
@@ -70,6 +70,10 @@ impl<I: Interner> context::AggregateOps<SlgContext<I>> for SlgContextOps<'_, I> 
                 break Guidance::Unknown;
             }
 
+            if let AnswerResult::Floundered = answers.peek_answer(|| should_continue()) {
+                break Guidance::Unknown;
+            }
+
             if !answers
                 .any_future_answer(|ref mut new_subst| new_subst.may_invalidate(interner, &subst))
             {
--- a/tests/test/misc.rs
+++ b/tests/test/misc.rs
@@ -96,6 +96,63 @@ fn flounder() {
     }
 }
 
+/// Don't return definite guidance if we flounder after finding one solution.
+#[test]
+fn flounder_ambiguous() {
+    test! {
+        program {
+            trait IntoIterator { }
+            #[non_enumerable]
+            trait OtherTrait { }
+
+            struct Ref<T> { }
+            struct A { }
+
+            impl IntoIterator for Ref<A> { }
+            impl<T> IntoIterator for Ref<T> where T: OtherTrait { }
+        }
+
+        goal {
+            exists<T> { Ref<T>: IntoIterator }
+        } yields {
+            "Ambiguous; no inference guidance"
+        }
+    }
+}
+
+/// Don't return definite guidance if we are able to merge two solutions and the
+/// third one matches that as well (the fourth may not).
+#[test]
+fn normalize_ambiguous() {
+    test! {
+        program {
+            trait IntoIterator { type Item; }
+
+            struct Ref<T> { }
+            struct A { }
+            struct B { }
+            struct C { }
+
+            struct D { }
+
+            impl IntoIterator for Ref<A> { type Item = Ref<A>; }
+            impl IntoIterator for Ref<B> { type Item = Ref<B>; }
+            impl IntoIterator for Ref<C> { type Item = Ref<C>; }
+            impl IntoIterator for Ref<D> { type Item = D; }
+        }
+
+        goal {
+            exists<T, U> {
+                Normalize(<Ref<T> as IntoIterator>::Item -> U)
+            }
+        } yields {
+            // FIXME: this is wrong!
+            "Ambiguous; definite substitution"
+            // "Ambiguous; no inference guidance"
+        }
+    }
+}
+
 // Test that, when solving `?T: Sized`, we only wind up pulling a few
 // answers before we stop.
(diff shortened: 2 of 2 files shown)


flodiebold:
I think there may still be a problem with the `any_future_answer` check, because I found another case where we sometimes erroneously get definite guidance; in that case, there's more than one solution, but the first few solutions are all of the form `SomeStruct<?0>` (well, `&?0`). I'm not seeing any floundering in that case.

jackh726:
I have to think about this a bit more, but I don't think this fix is the most *correct* one to make.

> I think the reason is that any_future_answer has no answers to check and hence returns true.
Well, that can't be the case because it can still `Flounder`.

This *might* be a problem with `MayInvalidate`, but I don't know yet.

Nevertheless, thanks for the test. If you remove `non_enumerable` from `OtherTrait`, it returns `Ambiguous` right? Also, can you add a short comment to the test to explain what it is for? Thanks.

flodiebold:
> I have to think about this a bit more, but I don't think this fix is the most _correct_ one to make.

That's very possible, especially since there still seems to be some similar problem.

> Nevertheless, thanks for the test. If you remove `non_enumerable` from `OtherTrait`, it returns `Ambiguous` right?

Yes.

> Also, can you add a short comment to the test to explain what it is for? Thanks.

Done.

flodiebold:
Ok, I was able to reproduce the second problem:
```rust

/// Don't return definite guidance if we are able to merge two solutions and the
/// third one matches that as well (the fourth may not).
#[test]
fn normalize_ambiguous() {
    test! {
        program {
            trait IntoIterator { type Item; }

            struct Ref<T> { }
            struct A { }
            struct B { }
            struct C { }

            struct D { }

            impl IntoIterator for Ref<A> { type Item = Ref<A>; }
            impl IntoIterator for Ref<B> { type Item = Ref<B>; }
            impl IntoIterator for Ref<C> { type Item = Ref<C>; }
            impl IntoIterator for Ref<D> { type Item = D; }
        }

        goal {
            exists<T, U> {
                Normalize(<Ref<T> as IntoIterator>::Item -> U)
            }
        } yields {
            "Ambiguous; no inference guidance"
        }
    }
}
```

This currently results in
```
Ambiguous; definite substitution for<?U0,?U0> { [?0 := ^0, ?1 := Ref<^1>] }
```
because `any_future_answer` actually just looks at the *next* answer, which happens to match in this case.

jackh726:
`any_future_answer` will test every strand with `MayInvalidate`. So this again points to `MayInvalidate` not being correct.

jackh726:
Also, this test passes on current `master`, but not with the changes in this PR.

flodiebold:
> Also, this test (in the comment) passes on current master, but not with the changes in this PR.

Ah, I think with my change it hits the 'answer cached' case:
https://github.com/rust-lang/chalk/blob/f85bb2ed57b78728bf14fc1f0101ac5415752e4d/chalk-engine/src/logic.rs#L213-L216

chalk-solve/src/solve/slg/aggregate.rs
```diff
@@ -69,6 +69,10 @@ impl<I: Interner> context::AggregateOps<SlgContext<I>> for SlgContextOps<'_, I>
                 break Guidance::Unknown;
             }
 
+            if let AnswerResult::Floundered = answers.peek_answer(|| should_continue()) {
+                break Guidance::Suggested(subst);
```
nikomatsakis: Hmm, I think if we *floundered* we probably want `Guidance::Unknown`, as that suggests that we never really explored the options.

nikomatsakis:
But I see there were a lot more comments :) I have to read those and fully grok what they're saying

flodiebold:
@nikomatsakis Done! I added the second test as well, with the currently wrong result.

tests/test/misc.rs
```diff
@@ -96,6 +96,63 @@ fn flounder() {
...
+        } yields {
+            "Ambiguous; no inference guidance"
+        }
+    }
+}
+
+/// Don't return definite guidance if we are able to merge two solutions and the
+/// third one matches that as well (the fourth may not).
+#[test]
+fn normalize_ambiguous() {
```
nikomatsakis, line 126: Does this test have anything to do with floundering?

tests/test/misc.rs
```diff
@@ -96,6 +96,63 @@ fn flounder() {
...
+            impl IntoIterator for Ref<C> { type Item = Ref<C>; }
+            impl IntoIterator for Ref<D> { type Item = D; }
+        }
+
+        goal {
+            exists<T, U> {
+                Normalize(<Ref<T> as IntoIterator>::Item -> U)
+            }
+        } yields {
+            // FIXME: this is wrong!
```
nikomatsakis, line 149: We should probably create a FIXME policy of creating open issues to track these...

tests/test/misc.rs
flodiebold: I don't think so (except that this test only reproduces the problem with the fix for the other problem, I think because it calls `peek_answer`)

nikomatsakis: I see, curious.

nikomatsakis:
I'm confused I guess if we want to land this now and investigate later or do more investigation. It seems like it's improving some problems you encountered in rust-analyzer?

flodiebold:
I think we could land this now. It's not that important for rust-analyzer anymore though since I switched to the recursive solver ;)

PR #336, opened 2020-02-27
Remove default dependency on lalrpop-intern

jackh726:
Closes #295 

So, the default interner (`ChalkIr`) is now available only with `cfg(test)` or `cfg(feature = "default-interner")`. The former for obvious reasons; the latter because of `chalk-integration`. Interner now has an `Identifier` associated type for the whole *one* place it's used in `chalk-ir` (`AssociatedTyDatum`). `RawId` has also been moved to the same configuration, along with the dependency on `lalrpop-intern`. I figured this is enough to close the linked issue, since these two cases (tests and chalk-integration) *we* control and know they are single-threaded. We could, in the future, change out lalrpop-intern for a library (like string_cache) that *is* thread-safe.

Also moved `TypeSort` and `TypeKind` to `chalk-integration` since they weren't used in `chalk_ir`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,7 +164,6 @@ dependencies = [
  "salsa 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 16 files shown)


nikomatsakis:
This seems like a good step. I've been wanting to remove the `ChalkIr` "default" out from `chalk-ir` and into `chalk-integration` in any case, do you think that would be possible?

Issue #260, opened 2019-10-08
Make chalk_solve::Solver UnwindSafe


matklad:
In rust-analyzer, we use panics to implement cancellation. That means that, while we use chalk to solve some goal, the callbacks from chalk into rust-analyzer might panic, unwinding past the Solver.

This unfortunately leaves the solver in an inconsistent state at the moment. See https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/Logic.20panics for some more details. 

This issue has been assigned to @marmeladema via [this comment](https://github.com/rust-lang/chalk/issues/260#issuecomment-616114664).

matklad:
Note that we now try to catch panics within the chalk itself, to make rust-analyzer more robust:

https://github.com/rust-analyzer/rust-analyzer/pull/2818

At the moment, we just reset chalk's state on panic, but it would be more convenient if chalk always was consistent after a panic.

jackh726:
With #338 merged, closing this only requires some regression tests for panic safety. This is pretty easy, so I'll mark it as such.

PR #338, opened 2020-02-28
Make chalk-engine more UnwindSafe

jackh726:
cc. https://github.com/rust-lang/chalk/issues/260

I can't call this 100% UnwindSafe, so I'm not going to say this *closes* the linked issue. But I think with proper tests added to confirm, it might be.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -4,6 +4,7 @@ use crate::context::{
 use crate::fallible::NoSolution;
 use crate::forest::Forest;
 use crate::hh::HhGoal;
+use crate::stack::{Stack, StackIndex};
 use crate::strand::{CanonicalStrand, SelectedSubgoal, Strand};
 use crate::table::AnswerIndex;
 use crate::{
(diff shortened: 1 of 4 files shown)


chalk-engine/src/logic.rs
```diff
@@ -77,11 +78,18 @@ impl<C: Context> Forest<C> {
         table: TableIndex,
         answer_index: AnswerIndex,
     ) -> RootSearchResult<CompleteAnswer<C>> {
-        assert!(self.stack.is_empty());
+        let stack = Stack::default();
```
nikomatsakis: interesting

jackh726:
@matklad and @flodiebold 

This might make it so that rust-analyzer doesn't have to reset chalk state on panic

matklad:
>I can't call this 100% UnwindSafe, so I'm not going to say this closes the linked issue. But I think with proper tests added to confirm, it might be.

So this doesn't literally implement `UnwindSafe` trait? I think it might be good to add such an impl even if we are not 100% sure it is correct, and let rust-analyzer find the bugs in practice? We unwind a lot, so, if something somewhere asserts, we'll notice it :D

jackh726:
If I get some time to add at least a couple tests (at least for regression in the future), then I'll also add `UnwindSafe` and close the related issue.

I would definitely like to see if everything is good for rust-analyzer without resetting state. I'm not sure if there are some corner cases that might be hit anywhere.

PR #342, opened 2020-03-02
Add self param to intern_data

yaahc:


chalk-ir/src/debug.rs
```diff
@@ -21,6 +21,13 @@ impl<I: Interner> Debug for AssocTypeId<I> {
     }
 }
 
+impl<I: Interner> Debug for Ty<I> {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        I::debug_ty(self, fmt).unwrap_or_else(|| unimplemented!())
```
yaahc: dont know what the default here should be

nikomatsakis: yeah that's a tricky one

Issue #340, opened 2020-03-02
extend interning of lifetimes, parameters, goals with `&self` arguments


nikomatsakis:
Building on https://github.com/rust-lang/chalk/issues/328, we should extend the other `intern_*` methods with `&self` arguments:

* [x] [`intern_lifetime`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.intern_lifetime)
* [x] [`intern_parameter`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.intern_parameter)
* [x] [`intern_goal`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.intern_goal)
* [x] [`intern_goals`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.intern_goals)
* [x] [`intern_substitution`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.intern_substitution)

nikomatsakis:
This is done! @yaahc rocks!

PR #345, opened 2020-03-06
Add &self arg to intern_param

yaahc:
related to #340

### DIFF
--- a/chalk-derive/src/lib.rs
+++ b/chalk-derive/src/lib.rs
@@ -34,10 +34,7 @@ pub fn derive_fold(item: TokenStream) -> TokenStream {
                     &self,
                     folder: &mut dyn Folder < 'i, #arg, #arg >,
                     binders: usize,
-                ) -> ::chalk_engine::fallible::Fallible<Self::Result>
-                where
-                    #arg: 'i,
-                {
+                ) -> ::chalk_engine::fallible::Fallible<Self::Result> {
                     #body
                 }
             }
--- a/chalk-derive/src/lib.rs
+++ b/chalk-derive/src/lib.rs
@@ -30,11 +30,14 @@ pub fn derive_fold(item: TokenStream) -> TokenStream {
             impl #impl_generics Fold < #arg, #arg > for #type_name #ty_generics #where_clause_ref {
                 type Result = Self;
 
-                fn fold_with(
+                fn fold_with<'i>(
                     &self,
-                    folder: &mut dyn Folder < #arg, #arg >,
+                    folder: &mut dyn Folder < 'i, #arg, #arg >,
                     binders: usize,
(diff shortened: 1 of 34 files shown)


chalk-ir/src/fold/boring_impls.rs
```diff
@@ -71,25 +71,26 @@ impl<I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Parameter<I> {
...
         let data = self.data().fold_with(folder, binders)?;
-        Ok(Parameter::new(data))
+        Ok(Parameter::new(folder.target_interner(), data))
     }
 }
 
 impl<I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Substitution<I> {
     type Result = Substitution<TI>;
     fn fold_with(&self, folder: &mut dyn Folder<I, TI>, binders: usize) -> Fallible<Self::Result> {
+        let folded: Vec<_> = self.iter().map(|p| p.fold_with(folder, binders)).collect();
```
yaahc: Had to collect here and blow to deal with lifetimes. `p.fold_with` takes a &mut ref to the folder and `target_interner()` takes a & ref to the folder.

chalk-derive/src/lib.rs
```diff
@@ -30,11 +30,14 @@ pub fn derive_fold(item: TokenStream) -> TokenStream {
...
 
-                fn fold_with(
+                fn fold_with<'i>(
                     &self,
-                    folder: &mut dyn Folder < #arg, #arg >,
+                    folder: &mut dyn Folder < 'i, #arg, #arg >,
                     binders: usize,
-                ) -> ::chalk_engine::fallible::Fallible<Self::Result> {
+                ) -> ::chalk_engine::fallible::Fallible<Self::Result>
+                where
```
nikomatsakis: Nit: this diff is probably not needed, but I guess it doesn't hurt -- in particular `#arg` will normally be a type name like `ChalkIr`, so this is "clearly true"

chalk-ir/src/lib.rs
```diff
@@ -954,14 +958,18 @@ pub enum ProgramClause<I: Interner> {
     ForAll(Binders<ProgramClauseImplication<I>>),
 }
 
+impl<I: Interner> HasInterner for ProgramClause<I> {
```
nikomatsakis: Pretty sure we can `#[derive]` this

chalk-ir/src/lib.rs
```diff
@@ -983,6 +991,13 @@ pub struct Canonical<T> {
     pub binders: Vec<ParameterKind<UniverseIndex>>,
 }
 
+impl<T> HasInterner for Canonical<T>
```
nikomatsakis: Similarly, I suspect we can `#[derive]` this

Issue #341, opened 2020-03-02
extend lookup methods for things beyond types with `&self`


nikomatsakis:
Building on https://github.com/rust-lang/chalk/issues/339 and https://github.com/rust-lang/chalk/issues/340, we should extend the `*_data` methods with `&self`:

* [x] [`lifetime_data`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.lifetime_data) -- #349 
* [x] [`parameter_data`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.parameter_data) -- #350
* [x] [`goal_data`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.goal_data) -- #352 
* [x] [`goals_data`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.goals_data) -- #353 
* [x] [`substitution_data`](http://rust-lang.github.io/chalk/chalk_ir/interner/trait.Interner.html#tymethod.substitution_data) -- #354 
* [ ] [refactor how we handle Debug impls to share code](https://github.com/rust-lang/chalk/pull/353#discussion_r394930949) -- #355

Mcat12:
@nikomatsakis As of #355, this issue should be complete?

nikomatsakis:
I think that's correct.

PR #349, opened 2020-03-09
Add self param to lifetime_data

yaahc:
part of #341

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -27,7 +27,6 @@ impl<I: Interner> Debug for Ty<I> {
     }
 }
 
-
 impl Display for UniverseIndex {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         write!(fmt, "U{}", self.counter)
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -21,6 +21,13 @@ impl<I: Interner> Debug for AssocTypeId<I> {
     }
 }
 
+impl<I: Interner> Debug for Ty<I> {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        I::debug_ty(self, fmt).unwrap_or_else(|| unimplemented!())
+    }
+}
+
+
 impl Display for UniverseIndex {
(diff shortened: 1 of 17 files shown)


chalk-ir/src/debug.rs
```diff
@@ -21,6 +21,18 @@ impl<I: Interner> Debug for AssocTypeId<I> {
...
 
+impl<I: Interner> Debug for Ty<I> {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        I::debug_ty(self, fmt).unwrap_or_else(|| unimplemented!())
+    }
+}
+
+impl<I: Interner> Debug for Lifetime<I> {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        I::debug_lifetime(self, fmt).unwrap_or_else(|| unimplemented!())
```
yaahc: This is triggering a failure in one of the tests, presumably because the test does not setup the TLS, but the test also doesn't instantiate a `Program` so I'm not sure how best to go about fixing it.

chalk-ir/src/interner.rs
```diff
@@ -272,19 +295,33 @@ mod default {
...
 
+        fn debug_ty(ty: &Ty<ChalkIr>, fmt: &mut fmt::Formatter<'_>) -> Option<fmt::Result> {
+            tls::with_current_program(|prog| Some(prog?.debug_ty(ty, fmt)))
+        }
+
+        fn debug_lifetime(
+            lifetime: &Lifetime<ChalkIr>,
+            fmt: &mut fmt::Formatter<'_>,
+        ) -> Option<fmt::Result> {
+            tls::with_current_program(|prog| Some(prog?.debug_lifetime(lifetime, fmt)))
```
nikomatsakis: so to fix the test failures, we could adapt this like so

```rust
tls::with_current_program(|prog| Some(prog?.debug_lifetime(lifetime, fmt)))
    .or_else(|| XXX)
```

where `XXX` gets the `lifetime.data` and (since we know it is an `Arc`...) just invokes `Debug` on it manually.

nikomatsakis: the other (maybe better) option would be to get the `Program` in scope, as that will result in cleaner debug output anyhow

PR #353, opened 2020-03-17
Add self paramter to Interner::goals_data

yaahc:
part of https://github.com/rust-lang/chalk/issues/341

### DIFF
--- a/chalk-integration/src/program.rs
+++ b/chalk-integration/src/program.rs
@@ -216,7 +216,7 @@ impl RustIrDatabase<ChalkIr> for Program {
         self.impl_data.values().any(|impl_datum| {
             let impl_trait_ref = &impl_datum.binders.value.trait_ref;
             impl_trait_ref.trait_id == auto_trait_id
-                && match impl_trait_ref.self_type_parameter().data(interner) {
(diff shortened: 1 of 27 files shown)


chalk-integration/src/program.rs
```diff
@@ -147,6 +147,35 @@ impl tls::DebugContext for Program {
...
         }
     }
+
+    fn debug_goal(
+        &self,
+        goal: &Goal<ChalkIr>,
+        fmt: &mut fmt::Formatter<'_>,
+    ) -> Result<(), fmt::Error> {
+        let interner = self.interner();
+        match goal.data(interner) {
```
nikomatsakis: Hmm, I realize as I read this -- I think *this* match should live in a `Debug` impl on `GoalData`, probably, which could be in the chalk-ir crate. In other words, this function (and probably the others nearby) should look like

```rust
let interner = self.interner();
write!(fmt, "{:?}", goal.data(interner));
Ok(())
```

nikomatsakis: but I think we should do this in a follow-up PR, I'm going to move this to a separate issue

PR #355, opened 2020-03-19
Extract Debug for GoalData implementation.

crlf0710:
Follow up from #353
cc #341

### DIFF
--- a/chalk-integration/src/program.rs
+++ b/chalk-integration/src/program.rs
@@ -4,9 +4,9 @@ use chalk_ir::debug::Angle;
 use chalk_ir::interner::ChalkIr;
 use chalk_ir::tls;
 use chalk_ir::{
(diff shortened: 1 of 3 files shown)


crlf0710:
@nikomatsakis 
I've added a new commit to create some helper structs and moved the rest debug formatting logic that needs an interner to their Debug implementations. 
* Is this the correct direction?
* `ParameterData` is actually `ParameterKind`, which already have a different `Debug` implementation. So I also created a helper struct for this. Is this ok?
* `SeparatorTraitRef` definition already lives in `debug.rs`, should i do something more?
* What should i do for those methods (debug_struct_id, debug_trait_id, debug_assoc_type_id, debug_alias) that doesn't use interner at all? Should i leave them alone?

nikomatsakis:
Thanks @crlf0710

Issue #356, opened 2020-03-23
extend `TraitDefn` with concept of "well-known trait" identifier


nikomatsakis:
In order to accomodate built-in traits, we first have to be able to identify them. The idea here is to extend `TraitDatum` with a kind of `well_known: Option<WellKnownTrait>` field. [I sketched an outline of how to do this on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/builtin.20traits/near/191539918).

nikomatsakis:
@yaahc is I think going to look into this

yaahc:
Sounds good, I'm assuming this will be a short issue to do so I'll hold off on starting with the `Sized` trait until you've got this landed.

nikomatsakis:
@yaahc I was hoping you could do this work :)

nikomatsakis:
That is, I think we can add it with `Sized` and maybe a few other traits but not really support them yet

nikomatsakis:
Let me know if you don't have time though and I'll clear the assignment

yaahc:
Oooo, I misread, i thought you said you were going to look into it. Lol okay yea I got this :+1:

PR #359, opened 2020-03-24
Start working on builtin traits in chalk

yaahc:
closes https://github.com/rust-lang/chalk/issues/356


chalk-rust-ir/src/lib.rs
```diff
@@ -87,6 +87,7 @@ pub struct StructFlags {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+/// JRL What does a TraitDatum represent?
 pub struct TraitDatum<I: Interner> {
```
jackh726: This essentially just captures the trait definition.

chalk-rust-ir/src/lib.rs
```diff
@@ -130,8 +138,11 @@ pub struct TraitDatumBound<I: Interner> {
 pub struct TraitFlags {
     pub auto: bool,
     pub marker: bool,
+    /// JRL What does upstream trait mean?
```
jackh726: The `upstream` flag is used for coherence checking.

chalk-rust-ir/src/lib.rs
```diff
@@ -130,8 +138,11 @@ pub struct TraitDatumBound<I: Interner> {
...
     pub upstream: bool,
+    /// JRL What fundamental traits are there? Does this just mean #[fundamental] like `Box`?
```
jackh726: Currently there are none :) It would be the equivalent of the struct fundamental though

chalk-rust-ir/src/lib.rs
```diff
@@ -130,8 +138,11 @@ pub struct TraitDatumBound<I: Interner> {
...
     pub fundamental: bool,
+    /// JRL Also what are these?
     pub non_enumerable: bool,
```
jackh726: `non_enumerable` just means we can't list all impls for a given trait

yaahc: Does this just mean in practice that

if defined in library and exported then non_enumerable = true
else non_enumerable = false

yaahc:
@jackh726 I updated the docs to include your explanations and added 2 more questions.

chalk-rust-ir/src/lib.rs
flodiebold: In practice at least rust-analyzer makes every trait `non_enumerable`, because we don't want Chalk to go off trying to find e.g. every type implementing `Iterator` :grimacing:

flodiebold: I.e. this isn't about orphan rules or something like that, it's a hint that Chalk shouldn't try enumerating types that implement the trait because there are too many, basically.

nikomatsakis: rustc makes every trait non-enumerable as well. eventually it may just be the only option I think.

chalk-integration/src/lowering.rs
```diff
@@ -1241,6 +1241,7 @@ impl LowerTrait for TraitDefn {
             binders: binders,
             flags: self.flags.lower(),
             associated_ty_ids,
+            well_known: self.well_known.as_ref().map(|t| t.lower()),
```
nikomatsakis: Nit: if well-known is `Copy` (as it probably should be?), then you don't need `as_ref`

chalk-parse/src/ast.rs
```diff
@@ -48,6 +48,12 @@ pub struct TraitDefn {
     pub where_clauses: Vec<QuantifiedWhereClause>,
     pub assoc_ty_defns: Vec<AssocTyDefn>,
     pub flags: TraitFlags,
+    pub well_known: Option<WellKnownTrait>,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
```
nikomatsakis: Copy :)

chalk-rust-ir/src/lib.rs
```diff
@@ -87,6 +87,10 @@ pub struct StructFlags {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+/// A rust intermediate representation (rust_ir) of a Trait Definition.
+///
+/// Not to be confused with a rust_ir for a Trait Implementation, which is
+/// represented with ??? JRL
```
nikomatsakis: I think you're asking about `ImplDatum`?

chalk-rust-ir/src/lib.rs
```diff
@@ -87,6 +87,10 @@ pub struct StructFlags {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+/// A rust intermediate representation (rust_ir) of a Trait Definition.
```
nikomatsakis: I always find it useful to include some Rust source. Maybe something like:

---

The intermediate representation that represents a trait definition. For example, given the following Rust code:

```rust
trait Foo<T>
where
    T: Debug,
{
    type Bar<U>;
}
```

this would represent the `trait Foo` declaration. Note that the details of the trait members (e.g., the associated type declaration (`type Bar<U>`) are not contained in this type, and are represented separately (e.g., in `AssocTypeDatum`).

chalk-rust-ir/src/lib.rs
```diff
@@ -87,6 +87,10 @@ pub struct StructFlags {
...
+///
+/// Not to be confused with a rust_ir for a Trait Implementation, which is
+/// represented with ??? JRL
 pub struct TraitDatum<I: Interner> {
     pub id: TraitId<I>,
```
nikomatsakis: Doc-comment:

The id of this trait; could be used to load the trait datum by invoking the [`trait_datum`] method.

[`trait_datum`]: http://rust-lang.github.io/chalk/chalk_solve/trait.RustIrDatabase.html#tymethod.trait_datum

chalk-rust-ir/src/lib.rs
```diff
@@ -99,6 +103,15 @@ pub struct TraitDatum<I: Interner> {
 
     /// The id of each associated type defined in the trait.
     pub associated_ty_ids: Vec<AssocTypeId<I>>,
```
nikomatsakis: The ids for the associated type members of the trait. The details of each type can be found by invoking the [`associated_type_data`](http://rust-lang.github.io/chalk/chalk_solve/trait.RustIrDatabase.html#tymethod.associated_ty_data) method.

chalk-rust-ir/src/lib.rs
```diff
@@ -99,6 +103,15 @@ pub struct TraitDatum<I: Interner> {
...
+
+    /// A marker indicating if this trait definition represents one of the
+    /// various builtin traits (sized, copy, etc)
+    pub well_known: Option<WellKnownTrait>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
nikomatsakis: A list of the traits that are "well known" to chalk, which means that the chalk-solve crate has special, hard-coded impls for them.

chalk-rust-ir/src/lib.rs
```diff
@@ -130,8 +143,15 @@ pub struct TraitDatumBound<I: Interner> {
 pub struct TraitFlags {
     pub auto: bool,
     pub marker: bool,
+    /// Indicate that a trait is defined upstream (in a dependency), used during
+    /// coherence checking.
     pub upstream: bool,
+    /// The trait equivalent of a struct fundamental, currently (2020-03-25)
```
nikomatsakis: A fundamental trait is a trait where adding an impl for an existing type is considered a breaking change. Examples of fundamental traits are the closure traits like `Fn` and `FnMut`. As of this writing, fundamental traits are declared by the unstable `#[fundamental]` attribute in rustc, and hence cannot appear outside of the standard library.

chalk-rust-ir/src/lib.rs
```diff
@@ -130,8 +143,15 @@ pub struct TraitDatumBound<I: Interner> {
 pub struct TraitFlags {
     pub auto: bool,
```
nikomatsakis: An "auto trait" is one that is "automatically implemented" for every struct, so long as no explicit impl is given. Examples are `Send` and `Sync`.

chalk-rust-ir/src/lib.rs
yaahc: The doc for `RustIrDatabase::trait_datum` says that this fn returns the trait datum for an impl, it seems like these comments are contradictory.

chalk-rust-ir/src/lib.rs
```diff
@@ -97,8 +125,20 @@ pub struct TraitDatum<I: Interner> {
...
 
-    /// The id of each associated type defined in the trait.
+    /// The ids for the associated type members of the trait. The details of
+    /// each type can be found by invoking the [`associated_ty_data`] method.
+    ///
+    /// [`associated_ty_data`]: ../chalk_solve/trait.RustIrDatabase.html#tymethod.associated_ty_data
     pub associated_ty_ids: Vec<AssocTypeId<I>>,
+
+    /// A list of the traits that are "well known" to chalk, which means that
+    /// the chalk-solve crate has special, hard-coded impls for them.
```
yaahc: I'm wondering if it would be better to put these comments on the definitions of the members' types rather than on the members themselves

chalk-rust-ir/src/lib.rs
nikomatsakis: yeah that's a copy and paste bug in the doc comment for `trait_datum`, good catch :)

chalk-rust-ir/src/lib.rs
nikomatsakis: This comment seems better for the `WellKnownTrait` type. I imagine the comment for the member would be:

`None` if this is a normal, user-defined trait; `Some` if this is a well-known trait.

yaahc:
Gotta resolve conflicts and I have a few doc comments I'm gonna update but I'll merge it in shortly

yaahc:
I also moved the comments for TraitID and AssocTyID to their struct definitions as well, and fixed the typo in RustIrDatabase.

PR #366, opened 2020-03-29
Use fallback debug impls instead of unimplemented

jackh726:
Also turn on deny(rust_2018_idioms)

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -23,78 +23,63 @@ impl<I: Interner> Debug for AssocTypeId<I> {
 
 impl<I: Interner> Debug for Ty<I> {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
-        I::debug_ty(self, fmt)
-            .unwrap_or_else(|| unimplemented!("cannot format Ty without setting Program in tls"))
+        I::debug_ty(self, fmt).unwrap_or_else(|| write!(fmt, "Ty(?)"))
     }
 }
 
 impl<I: Interner> Debug for Lifetime<I> {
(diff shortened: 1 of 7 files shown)


chalk-ir/src/debug.rs
```diff
@@ -23,78 +23,63 @@ impl<I: Interner> Debug for AssocTypeId<I> {
 
 impl<I: Interner> Debug for Ty<I> {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
-        I::debug_ty(self, fmt)
-            .unwrap_or_else(|| unimplemented!("cannot format Ty without setting Program in tls"))
+        I::debug_ty(self, fmt).unwrap_or_else(|| write!(fmt, "Ty(?)"))
```
nikomatsakis:
```suggestion
        I::debug_ty(self, fmt).unwrap_or_else(|| write!(fmt, "{:?}", self.interned))
```

Weren't we going to print the 'interned' format instead? Something like this?

chalk-ir/src/lib.rs
```diff
@@ -1325,6 +1343,10 @@ impl<I: Interner> Goal<I> {
         Self { interned }
     }
 
+    pub fn interned(&self) -> &I::InternedGoal {
```
nikomatsakis: I don't object to these methods, but we're not using them anywhere, right? And I don't really think they're *needed*

chalk-ir/src/debug.rs
jackh726: 🤦‍♀️I forgot those are `: Debug`

chalk-ir/src/lib.rs
jackh726: Well, I was using them in rustc for debug_* functions. Alternatively, we could pass `I::Interned*` to debug_* functions

PR #370, opened 2020-04-03
Intern Vec<ProgramClause<I>>

tirr-c:
This PR interns `ProgramClause` and `Vec<ProgramClause>`, and makes changes accordingly.

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -1244,33 +1244,40 @@ impl<I: Interner> ProgramClauseData<I> {
 
     pub fn intern(self, interner: &I) -> ProgramClause<I> {
         ProgramClause {
-            clause: interner.intern_program_clause(self),
+            interned: interner.intern_program_clause(self),
         }
     }
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasInterner)]
 pub struct ProgramClause<I: Interner> {
-    clause: I::InternedProgramClause,
+    interned: I::InternedProgramClause,
 }
 
 impl<I: Interner> ProgramClause<I> {
+    pub fn new(interner: &I, clause: ProgramClauseData<I>) -> Self {
+        let interned = interner.intern_program_clause(clause);
+        Self { interned }
+    }
+
     pub fn into_from_env_clause(self, interner: &I) -> ProgramClause<I> {
         let program_clause_data = self.data(interner);
         let new_clause = program_clause_data.clone().into_from_env_clause(interner);
-        ProgramClause {
-            clause: interner.intern_program_clause(new_clause),
-        }
+        Self::new(interner, new_clause)
+    }
+
+    pub fn interned(&self) -> &I::InternedProgramClause {
+        &self.interned
     }
 
     pub fn data(&self, interner: &I) -> &ProgramClauseData<I> {
-        interner.program_clause_data(&self.clause)
+        interner.program_clause_data(&self.interned)
     }
 }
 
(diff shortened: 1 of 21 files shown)


chalk-ir/src/lib.rs
```diff
@@ -1227,17 +1227,42 @@ impl<I: Interner> ProgramClauseImplication<I> {
...
+    pub fn intern(self, interner: &I) -> ProgramClause<I> {
+        ProgramClause {
+            clause: interner.intern_program_clause(self),
+        }
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasInterner)]
+pub struct ProgramClause<I: Interner> {
+    clause: I::InternedProgramClause,
```
nikomatsakis, line 1254: Nit: I think we we call this `interned` in all the other structs

chalk-ir/src/lib.rs
```diff
@@ -1227,17 +1227,42 @@ impl<I: Interner> ProgramClauseImplication<I> {
...
+}
+
+impl<I: Interner> ProgramClause<I> {
```
nikomatsakis, line 1257: Nit: We should probably have a `new` function, like [the `Goal::new`](http://rust-lang.github.io/chalk/chalk_ir/struct.Goal.html#method.new)

chalk-ir/src/lib.rs
```diff
@@ -1227,17 +1227,42 @@ impl<I: Interner> ProgramClauseImplication<I> {
...
+    pub fn into_from_env_clause(self, interner: &I) -> ProgramClause<I> {
+        let program_clause_data = self.data(interner);
+        let new_clause = program_clause_data.clone().into_from_env_clause(interner);
+        ProgramClause {
```
nikomatsakis, line 1261: then we can call `ProgramClause::new(...)`

chalk-ir/src/lib.rs
```diff
@@ -1265,6 +1265,56 @@ impl<I: Interner> ProgramClause<I> {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasInterner)]
+pub struct ProgramClauses<I: Interner> {
+    clauses: I::InternedProgramClauses,
```
nikomatsakis, line 1273: I expect we shoul call this `interned` too

chalk-ir/src/lib.rs
```diff
@@ -1227,17 +1227,42 @@ impl<I: Interner> ProgramClauseImplication<I> {
...
+        let new_clause = program_clause_data.clone().into_from_env_clause(interner);
+        ProgramClause {
+            clause: interner.intern_program_clause(new_clause),
+        }
+    }
+
+    pub fn data(&self, interner: &I) -> &ProgramClauseData<I> {
+        interner.program_clause_data(&self.clause)
+    }
 }
```
nikomatsakis, line 1269: Also I think we should add an `interned(&self) -> &I::InternedProgramClause` accessor

nikomatsakis:
Looks like [the tests need to be updated, too](https://github.com/rust-lang/chalk/runs/559004396#step:6:290)

tirr-c:
@nikomatsakis Done, pushed changes!

chalk-ir/src/lib.rs
jackh726: We call it `interned` except in `Goal` and `Goals` IIRC. I think those should be renamed and everything should be `interned`.

nikomatsakis: It looks like @tirr-c fixed the new cases in this PR, we can create an issue for the others

PR #371, opened 2020-04-04
Add new entries to `TypeName`

MarkMcCaskey:
Current progress is: 
- Function code threaded through system but missing important pieces
- Closure is only partially threaded through
- Scalar and tuple code on [this branch](https://github.com/rust-lang/chalk/compare/master...MarkMcCaskey:feature/add-scalar-to-typename?expand=1)

Part of #368

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -104,6 +104,12 @@ pub enum TypeName<I: Interner> {
     /// an associated type like `Iterator::Item`; see `AssociatedType` for details
     AssociatedType(AssocTypeId<I>),
 
+    /// TODO(markmccaskey):
+    FnDef(FnDefId<I>),
+
+    /// TODO(markmccaskey):
+    Closure(ClosureId<I>),
+
     /// This can be used to represent an error, e.g. during name resolution of a type.
     /// Chalk itself will not produce this, just pass it through when given.
     Error,
@@ -145,6 +151,12 @@ impl UniverseIndex {
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct StructId<I: Interner>(pub I::DefId);
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FnDefId<I: Interner>(pub I::DefId);
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ClosureId<I: Interner>(pub I::DefId);
 
 /// The id of a trait definition; could be used to load the trait datum by
 /// invoking the [`trait_datum`] method.
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -22,6 +22,7 @@ pub struct Program {
 pub enum Item {
     StructDefn(StructDefn),
     TraitDefn(TraitDefn),
+    FnDefn(FnDefn),
     Impl(Impl),
     Clause(Clause),
 }
@@ -39,6 +40,30 @@ pub struct StructDefn {
 pub struct StructFlags {
     pub upstream: bool,
     pub fundamental: bool,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FnDefn {
+    pub name: Identifier,
+    pub generics: Vec<ParameterKind>,
+    pub where_clauses: Vec<QuantifiedWhereClause>,
+    pub flags: FnFlags,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FnFlags {
+    pub upstream: bool,
+    pub fundamental: bool,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct ClosureDefn {
+    // What's a closure's name? Need to generate an `Identifier` or something
+    // TODO(markmccaskey): investigate what rustc does here
+    pub name: Identifier,
+    pub parameter_kinds: Vec<ParameterKind>,
+    pub where_clauses: Vec<QuantifiedWhereClause>,
+    pub flags: FnFlags,
 }
 
 #[derive(Clone, PartialEq, Eq, Debug)]
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -47,6 +47,7 @@ pub struct FnDefn {
     pub name: Identifier,
     pub generics: Vec<ParameterKind>,
     pub where_clauses: Vec<QuantifiedWhereClause>,
+    pub fields: Vec<Field>,
     pub flags: FnFlags,
 }
 
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -92,8 +92,7 @@ pub struct StructFlags {
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct FnDefDatumBound<I: Interner> {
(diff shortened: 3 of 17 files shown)


chalk-ir/src/lib.rs
```diff
@@ -96,6 +96,41 @@ impl<G: HasInterner> HasInterner for InEnvironment<G> {
...
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum FloatTy {
+    F32,
+    F64,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Scalar {
```
jackh726, line 126: I like this idea of grouping these into a single enum. 👍

chalk-ir/src/lib.rs
```diff
@@ -146,6 +193,12 @@ impl UniverseIndex {
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct StructId<I: Interner>(pub I::DefId);
 
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FnDefId<I: Interner>(pub I::DefId);
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ClosureId<I: Interner>(pub I::DefId);
```
jackh726, line 200: We had previously talked about moving all of these as associated types on `Interner`. This is making me somewhat think this is a good idea. But maybe not in this PR.

chalk-parse/src/ast.rs
```diff
@@ -41,6 +42,34 @@ pub struct StructFlags {
     pub fundamental: bool,
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FnDefn {
+    pub name: Identifier,
+    pub parameter_kinds: Vec<ParameterKind>,
+    pub where_clauses: Vec<QuantifiedWhereClause>,
+    pub params: Vec<Field>,
+    pub returns: Vec<Field>,
```
jackh726, line 51: This doesn't feel right to me. If anything, it should be a single `Field`.

chalk-parse/src/ast.rs
```diff
@@ -41,6 +42,34 @@ pub struct StructFlags {
...
+pub struct FnFlags {
+    pub upstream: bool,
+    pub fundamental: bool,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct ClosureDefn {
+    // What's a closure's name? Need to generate an `Identifier` or something
+    // TODO(markmccaskey): investigate what rustc does here
+    pub name: Identifier,
```
jackh726, line 65: Technically, closures aren't nameable. But I'm not sure if rustc generates a name.

chalk-parse/src/ast.rs
```diff
@@ -41,6 +42,34 @@ pub struct StructFlags {
     pub fundamental: bool,
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FnDefn {
+    pub name: Identifier,
+    pub parameter_kinds: Vec<ParameterKind>,
```
jackh726, line 48: Might be better to rename this "generics" or something.

chalk-rust-ir/src/lib.rs
```diff
@@ -90,6 +90,44 @@ pub struct StructFlags {
...
+    //pub binders: Binders<StructDatumBound<I>>,
+    pub id: ClosureId<I>,
+}
+
+impl<I: Interner> ClosureDatum<I> {
+    pub fn name(&self, interner: &I) -> TypeName<I> {
+        self.id.cast(interner)
+    }
+}
+
```
jackh726, line 130: Why are these here, but the others are in `chalk-ir`? Plus there's duplicate structs like `FnDefFlags`/`FnFlags`

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -275,6 +275,172 @@ impl<I: Interner> ToProgramClauses<I> for StructDatum<I> {
     }
 }
 
+impl<I: Interner> ToProgramClauses<I> for FnDefDatum<I> {
```
jackh726, line 278: This seems mostly duplicated from `StructDatum`?

chalk-parse/src/ast.rs
flodiebold: Do you actually need the function params and return at all? I don't see where they should matter to Chalk :thinking:

nikomatsakis: Well, we will *eventually* need this information. We are going to have to generate `FnMut` impls and the like.

chalk-parse/src/ast.rs
nikomatsakis: Yes, rustc has an internal name. I think giving them a dummy name is a fine thing to do for chalk tests.

chalk-parse/src/ast.rs
nikomatsakis: (But it might plausibly be done in a follow-up PR)

chalk-parse/src/ast.rs
nikomatsakis: That said, we'll probably want to (a) consolidate some of this with functions and (b) we may want other things eventually to better match rustc. But we can evolve this part.

chalk-solve/src/clauses/program_clauses.rs
nikomatsakis: Interesting stuff! We'll have to think about just which rules we want here...

chalk-rust-ir/src/lib.rs
MarkMcCaskey: The current implementation is pretty much just a duplication and slight modification of `Struct` as a starting point.  For example, [`chalk_ast::StructFlags`](http://rust-lang.github.io/chalk/chalk_parse/ast/struct.StructFlags.html) and [`chalk_rust_ir::StructFlags`](http://rust-lang.github.io/chalk/chalk_rust_ir/struct.StructFlags.html) turned into `FnDefFlags` and `FnFlags`.  And `FnDefDatum` is here because [`StructDatum`](http://rust-lang.github.io/chalk/chalk_rust_ir/struct.StructDatum.html) is also here!

I'm oscillating between thinking I understand the goal and the pieces I'm working with and thinking I don't understand it :smile: .  I'll go back to the basics and see if I can get a better understanding of the pieces involved here!  The review helped me realize that my working understanding of [`Field`](http://rust-lang.github.io/chalk/chalk_parse/ast/struct.Field.html) was off base, for example, and perhaps I've missed other important details!

MarkMcCaskey:
Thanks for the review!  I'm realizing that I've probably gone a bit too far into unknown territory, so I'm going to read through the docs on chalk and think about this some more to try to get a more solid foundation before continuing tomorrow night.

nikomatsakis:
@MarkMcCaskey perhaps, but actually I thnk the code you wrote is all basically "on target" -- that is, the details of closures/functions might be a bit off, but you've got basically the right pieces in basically the right places :)

MarkMcCaskey:
I added basic parsing for (empty) function definitions so we can have tests for this.

I noticed in the tests there's code like,

```rust
            impl<T> Print for OnlySized<T> {
                // fn print() {
                //     println!("{}", std::mem::size_of::<T>());
                // }
            }
``` 

Is it in the scope of this PR to make this work?

nikomatsakis:
@MarkMcCaskey definitely not =)

nikomatsakis:
Closing in favor of #394 -- let's revisit this stuff once that one lands.

Issue #351, opened 2020-03-17
experiment with resurrecting the recursive solver


nikomatsakis:
We should experiment with returning to a "recursive style" solver in chalk. As described in [this gist](https://gist.github.com/nikomatsakis/bfbdbe588d6fc61ecb09e3b51847fb7c), recursive solving does offer some potential advantages:

* if we extend the setup, it may resolve the ambiguous associated type problem (ironic since we previously removed the recursive solver in part because we couldn't make it handle associated types well)
* less overall work, at the cost of completeness
* cleaner handling for multiple threads doing queries at once without locking (the cache is separated from the main solver stack)
* smoother integration into systems like chalk

Work items:

* [x] Initial PR 
* [ ] Coinduction semantics https://github.com/rust-lang/chalk/issues/399
* [ ] Invoke `program_clauses_for_goal` on the canonicalized form vs instantiating the goal first ([comment](https://github.com/rust-lang/chalk/pull/372#discussion_r408038753))

This issue has been assigned to @flodiebold via [this comment](https://github.com/rust-lang/chalk/issues/351#issuecomment-614047090).

nikomatsakis:
@flodiebold is working on this

jackh726:
Dang it that didn't work.

jackh726:
I would say that the "experiment" is done. Any follow up issues (including #399) can just be separate.

PR #372, opened 2020-04-05
Recursive solver

flodiebold:
This is getting close to a state where I think it could be merged. Tests are run for both solvers unless otherwise specified.

To do:
 - [x] lowering tests only run with the default solver currently
 - [x] move the `Generalize` folder somewhere else
 - [x] clean up a bit how high-priority clauses override low-priority ones
 - [x] add Floundered error in `program_clauses` instead of using `Option`
 - [ ] maybe some other things I forgot


chalk-ir/src/lib.rs
```diff
@@ -836,6 +836,17 @@ impl<I: Interner> AliasTy<I> {
     pub fn intern(self, interner: &I) -> Ty<I> {
         Ty::new(interner, self)
     }
+
+    pub fn type_parameters<'a>(&'a self, interner: &'a I) -> impl Iterator<Item = Ty<I>> + 'a {
```
nikomatsakis, line 840: Is this only used to support `self_type_parameter` below? It's a bit of a dubious thing to request otherwise, so maybe we should remove it if it is dead. (It encourages you to ignore lifetimes/constants, we've seen in rustc that's that's a bit error-prone.)

chalk-solve/src/clauses.rs
```diff
@@ -139,17 +141,26 @@ fn program_clauses_that_could_match<I: Interner>(
...
 
     match goal {
         DomainGoal::Holds(WhereClause::Implemented(trait_ref)) => {
             let trait_id = trait_ref.trait_id;
 
+            let trait_datum = db.trait_datum(trait_id);
+
+            if trait_datum.is_non_enumerable_trait() || trait_datum.is_auto_trait() {
+                let self_ty = trait_ref.self_type_parameter(interner);
+                if self_ty.bound(interner).is_some() || self_ty.inference_var(interner).is_some() {
```
nikomatsakis, line 156: I don't think `bound` should be possible here -- did you see that in practice?

chalk-solve/src/clauses.rs
```diff
@@ -168,8 +179,8 @@ fn program_clauses_that_could_match<I: Interner>(
                             push_auto_trait_impls(builder, trait_id, struct_id);
                         }
                     }
-                    TyData::InferenceVar(_) => {
-                        panic!("auto-traits should flounder if nothing is known")
+                    TyData::InferenceVar(_) | TyData::BoundVar(_) => {
```
nikomatsakis, line 182: Seems like maybe you did... that seems like a bug somewhere -- oh, right I think you said you are invoking this in the recursive solver before doing instantiation? Not sure what I think of that but maybe it's ok.

chalk-solve/src/clauses.rs
```diff
@@ -222,19 +233,24 @@ fn program_clauses_that_could_match<I: Interner>(
...
-                    // ```
-                    // forall<'a> { Implemented(dyn Fn(&u8): Fn<(&'a u8)>) }
-                    // ```
-                    let qwc = exists_qwc.substitute(interner, &[self_ty.clone().cast(interner)]);
-
-                    builder.push_binders(&qwc, |builder, wc| {
-                        builder.push_fact(wc);
-                    });
-                }
+                let generalized_dyn_ty = generalize::Generalize::apply(db.interner(), dyn_ty);
```
nikomatsakis, line 236: What is this...? Needs a comment at least

chalk-solve/src/clauses/generalize.rs
```diff
@@ -0,0 +1,80 @@
+//! This gets rid of free variables in a type by replacing them by fresh bound
+//! ones. We use this when building clauses that contain types passed to
+//! `program_clauses`; these may contain variables, and just copying those
+//! variables verbatim leads to problems. Instead, we return a slightly more
+//! general program clause, with new variables in those places.
```
nikomatsakis, line 5: Hmm, ok, is this specific to `dyn` types? I think all the other program clauses are generated purely from declaration sites, but I guess those do kind of "mix" data from the goal, and I can see how that could lead to problems. It'd be good to give some examples.

chalk-solve/src/clauses.rs
nikomatsakis: OK, read the comments on the generalize module, I think I see what's going on here. I do think we should comment somewhere. If I understand, you might have a type like

```rust
dyn Foo<?X>
```

and we will "generalize" this to `exists<A> { dyn Foo<A> }` and then make a rule like

```
forall<A> {
    Implemented(dyn Foo<A>: Foo<A>)
}
```

which... does seem right.

chalk-solve/src/clauses.rs
```diff
@@ -277,6 +293,36 @@ fn program_clauses_that_could_match<I: Interner>(
             let associated_ty_datum = db.associated_ty_data(alias.associated_ty_id);
             let trait_id = associated_ty_datum.trait_id;
             let trait_parameters = db.trait_parameters_from_projection(alias);
+
+            if (alias
+                .self_type_parameter(interner)
+                .bound(interner)
```
nikomatsakis, line 299: it seems like passing in bound variables is causing a lot of code duplication. At minimum, we should make a helper, but I'm semi-inclined not to pass them in in the first place. But I'll have to see when I get to the code below. =)

chalk-solve/src/clauses.rs
```diff
@@ -277,6 +293,36 @@ fn program_clauses_that_could_match<I: Interner>(
...
+                    .inference_var(interner)
+                    .is_some())
+                && (ty.bound(interner).is_some() || ty.inference_var(interner).is_some())
+            {
+                return Err(Floundered);
+            }
+
+            let trait_datum = db.trait_datum(trait_id);
+
+            // FIXME
```
nikomatsakis, line 312: Why FIXME?

chalk-solve/src/clauses.rs
```diff
@@ -277,6 +293,36 @@ fn program_clauses_that_could_match<I: Interner>(
             let associated_ty_datum = db.associated_ty_data(alias.associated_ty_id);
             let trait_id = associated_ty_datum.trait_id;
             let trait_parameters = db.trait_parameters_from_projection(alias);
+
+            if (alias
```
nikomatsakis, line 297: Comment? I think:

// Flounder if both the self-type *and* the target type are unknown.
//
// e.g., `Normalize(<?X as Iterator>::Item = ?Y)`

chalk-solve/src/clauses.rs
nikomatsakis: // Flounder if the self-type is unknown and the trait is non-enumerable.
//
// e.g., `Normalize(<?X as Iterator>::Item = u32)`

chalk-solve/src/infer.rs
```diff
@@ -204,6 +204,36 @@ impl<I: Interner> InferenceTable<I> {
             InferenceValue::Bound(_) => panic!("var_universe invoked on bound variable"),
         }
     }
+
+    /// Check whether the given substitution is the identity substitution in this
```
nikomatsakis, line 208: I feel like we .. have code for this? But maybe not in the inference context

tests/test/cycle.rs
```diff
@@ -163,9 +163,15 @@ fn overflow() {
...
         goal {
             S<Z>: Q
-        } yields {
+        } yields[SolverChoice::slg(10, None)] {
             "Ambiguous; no inference guidance"
         }
+
+        goal {
+            S<Z>: Q
+        } yields[SolverChoice::recursive()] {
```
nikomatsakis, line 172: This surprises me a bit.

tests/test/projection.rs
```diff
@@ -76,6 +76,7 @@ fn normalize_basic() {
             "Unique"
         }
 
+        /* TODO: this doesn't seem to be correct, actually
```
nikomatsakis, line 79: Right, so the problem is that if `U = Vec<T>`, then we have to prove `Normalized(<T as Iterator>::Item -> T)`, which of course we can't do. I wonder why the SLG solver considers this ambiguous, actually? Maybe worth investigating?

I think I'd rather either delete the test or make it `yields[SLGSolver]` and open a FIXME issue to potentially investigate why it behaves this way.

nikomatsakis: Also a `yields[Recursive]` to record the recursive behavior.

tests/test/projection.rs
```diff
@@ -140,11 +142,101 @@ fn projection_equality() {
...
                 S: Trait2<U>
             }
-        } yields {
-            // FIXME(rust-lang/chalk#234) -- there is really only one
-            // *reasonable* solution here, which is `u32`, but we get
-            // confused because `(Trait1::Type)<S>` seems valid too.
-            "Ambiguous; no inference guidance"
+        } yields[SolverChoice::recursive()] {
+            "Unique; substitution [?0 := u32]"
+        }
```
nikomatsakis, line 147: can we include the yields[SLG] with the old result, to record it?

tests/test/projection.rs
```diff
@@ -444,6 +536,7 @@ fn normalize_under_binder() {
             }
         }
 
+        /* TODO: this doesn't seem to be correct, actually
```
nikomatsakis, line 539: Again, let's document the behavior of the two solvers -- what does recursive solver do here?

I think I expect... U = I32?

tests/test/projection.rs
```diff
@@ -464,6 +558,7 @@ fn normalize_under_binder() {
             "Unique; substitution [?0 := I32], lifetime constraints []"
         }
 
+        /* TODO: this doesn't seem to be correct, actually
```
nikomatsakis, line 561: As above

chalk-solve/src/clauses.rs
flodiebold: Exactly -- it's also kind of related to the fact that the recursive solver calls `program_clauses` before instantiation, the SLG solver would probably have inference variables there and not have a problem :thinking:

chalk-solve/src/clauses/generalize.rs
flodiebold: Yes, it's just `dyn` at the moment

chalk-solve/src/clauses.rs
flodiebold: Just FIXME because I went back and forth a few times on this and wasn't sure whether to keep it :sweat_smile:

chalk-solve/src/infer.rs
flodiebold: I don't know, though this function did exist before and got removed after the recursive solver got removed

chalk-solve/src/clauses/generalize.rs
nikomatsakis: Yeah I can't think of another case where it would arise -- it would basically be any case where we are generating rules *from the goal itself* and not from definitions that the goal references, but I think that's specific to `dyn` right now.

chalk-solve/src/clauses.rs
flodiebold: I've removed this one in favor of just the floundering on unknown self type.

tests/test/projection.rs
flodiebold: I think I misunderstood the test -- the recursive solver actually also gives ambiguous. The comment is wrong though, the alternative isn't `Vec<T>`, but `Vec<Iterator::Item<T>>`. I added back the test and fixed the comment

tests/test/cycle.rs
flodiebold: Yeah, I didn't fully investigate why it happens.

flodiebold:
I added back the commented-out tests and added a few comments as suggested.

nikomatsakis:
@flodiebold can you rename `recursive/mod.rs` to `recursive.rs`, since we've been using that module convention in chalk elsewhere?

chalk-solve/src/clauses.rs
nikomatsakis: OK, so, I understand why this is setup this way now. I'm a bit unsure about it. For one thing, I'm not sure whether bound variables in canonicalized goals should *exclusively* represent inference variables, though that's the case right now. In rustc, canonicalization also "renumbers and replaces" placeholders, and the more I think about it, the more I think that makes sense. 

I guess that to change this would require creating the inference context earlier and leveraging checkpoints. i.e., we would create the inference table in `solve_from_clauses` and then clone it for each `Fulfill::new` call (or else, if not cloning, using a snapshot).

nikomatsakis: But I think we could address this in a follow-up issue.

chalk-solve/src/recursive/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+
+        // Next, check if the goal is in the search tree already.
+        if let Some(dfn) = self.context.search_graph.lookup(&goal) {
+            // Check if this table is still on the stack.
+            if let Some(depth) = self.context.search_graph[dfn].stack_depth {
+                // Is this a coinductive goal? If so, that is success,
+                // so we can return normally. Note that this return is
+                // not tabled.
+                //
+                // XXX how does caching with coinduction work?
```
nikomatsakis: We need to open a follow-up issue for the failing coinduction tests.

chalk-solve/src/recursive/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+                return *minimums;
+            }
+
+            if (
+                self.context.search_graph[dfn].solution_priority,
+                current_prio,
+            ) == (ClausePriority::High, ClausePriority::Low)
+                && self.context.search_graph[dfn].solution.is_ok()
+            {
+                // TODO check solution inputs?
```
nikomatsakis: So what is the role of this code, @flodiebold? I see there is also the `combine_with_priorities` code -- it seems like we want to use that here.

Also, total nit, but I find this hard to read. The `is_ok()` check is kind of obscured, and the tuple makes me jump back and forth to figure out what is compared to what. I'd prefer:

```rust
if (
    current_prio = ClausePriority::Low &&
    self.context.search_graph[dfn].solution_priority == ClausePriority::High &&
    self.context.search_graph[dfn].solution.is_ok()
) {
```

flodiebold: I wasn't sure about using `combine_with_priorities` since the code generally doesn't combine solutions from earlier iterations here, it just overwrites them. But maybe that was equivalent so far :thinking:

flodiebold: Using `combine_with_priorities` here is actually kind of complicated, because `combine_with_priorities` only works with a `DomainGoal` and `Solution`s, and here we have a `Goal` and `Result<Solution>`s, so we have to handle a bunch more cases :thinking:

flodiebold: @nikomatsakis done, though as mentioned more complicated than it seemed at first...

tests/test/cycle.rs
flodiebold: This is actually fixed now that I had to make it flounder when truncation would happen.

chalk-solve/src/clauses.rs
nikomatsakis: added note to https://github.com/rust-lang/chalk/issues/351

PR #373, opened 2020-04-05
add `Visitor` and friends 

### LINKED ISSUE #333: introduce a `Visit` trait to go with `Fold`
Right now we have a `Fold` trait that is used to transform types and other bits of IR. But sometimes we don't need to recreate the IR, we instead want to compute a value (e.g., count all the types etc). You can model this as a folder with mutable state, but it'd be more efficient to have a "visitor" trait. 

Some examples where this would be useful:

* the truncator is right now only used to detect when terms get too big
* the `UCollector`
* the `needs_shift` method
* some upcoming work as well

Areredify:
Closes #333 
Is `needs_shift` documentation accurate? Implementation gave me the impression that it's the opposite: check if there is any free variables (instead of bound variables).

### DIFF
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -15,14 +15,25 @@ pub use visitors::VisitExt;
 pub trait VisitResult: Sized {
     fn new() -> Self;
 
-    fn and_then(self, op: impl FnOnce() -> Self) -> Self;
+    fn return_early(&self) -> bool;
+    fn combine(self, other: Self) -> Self;
+
+    fn and_then(self, op: impl FnOnce() -> Self) -> Self {
+        if self.return_early() {
+            self
+        } else {
+            self.combine(op())
+        }
+    }
 }
 
 impl VisitResult for () {
     fn new() -> () {}
-    fn and_then(self, op: impl FnOnce() -> ()) -> () {
-        op()
+
+    fn return_early(&self) -> bool {
+        false
     }
+    fn combine(self, _other: Self) {}
 }
 
 /// A "visitor" recursively folds some term -- that is, some bit of IR,
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -10,6 +10,8 @@ mod binder_impls;
 mod boring_impls;
 pub mod visitors;
 
+pub use visitors::VisitExt;
+
 pub trait VisitResult: Sized {
     fn new() -> Self;
 
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -2,8 +2,8 @@
 use std::fmt::Debug;
 
 use crate::{
-    AliasEq, BoundVar, DebruijnIndex, Goal, InferenceVar, Interner, Lifetime, LifetimeData,
-    PlaceholderIndex, ProgramClause, Ty, TyData,
+    BoundVar, DebruijnIndex, DomainGoal, Goal, InferenceVar, Interner, Lifetime, LifetimeData,
+    PlaceholderIndex, ProgramClause, Ty, TyData, WhereClause,
 };
 
 mod binder_impls;
@@ -83,6 +83,14 @@ where
         goal.super_visit_with(self.as_dyn(), outer_binder)
     }
 
+    fn visit_domain_goal(
+        &mut self,
+        domain_goal: &DomainGoal<I>,
+        outer_binder: DebruijnIndex,
+    ) -> Self::Result {
+        domain_goal.super_visit_with(self.as_dyn(), outer_binder)
+    }
+
     /// If overridden to return true, then visiting will panic if a
     /// free variable is encountered. This should be done if free
     /// type/lifetime variables are not expected.
@@ -159,12 +167,12 @@ where
         }
     }
 
-    fn visit_alias_eq(
+    fn visit_where_clause(
         &mut self,
-        alias_eq: &AliasEq<I>,
+        where_clause: &WhereClause<I>,
         outer_binder: DebruijnIndex,
     ) -> Self::Result {
-        alias_eq.super_visit_with(self.as_dyn(), outer_binder)
+        where_clause.super_visit_with(self.as_dyn(), outer_binder)
     }
 
     /// As with `visit_free_placeholder_ty`, but for lifetimes.
@@ -375,7 +383,7 @@ impl<I: Interner> Visit<I> for ProgramClause<I> {
     }
 }
 
-impl<I: Interner> Visit<I> for AliasEq<I> {
+impl<I: Interner> Visit<I> for WhereClause<I> {
     fn visit_with<'i, R: VisitResult>(
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
@@ -384,12 +392,12 @@ impl<I: Interner> Visit<I> for AliasEq<I> {
     where
         I: 'i,
     {
-        visitor.visit_alias_eq(self, outer_binder)
+        visitor.visit_where_clause(self, outer_binder)
     }
 }
 
-impl<I: Interner> SuperVisit<I> for AliasEq<I> {
-    fn super_visit_with<'i, R: VisitResult>(
+impl<I: Interner> Visit<I> for DomainGoal<I> {
+    fn visit_with<'i, R: VisitResult>(
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
@@ -397,8 +405,6 @@ impl<I: Interner> SuperVisit<I> for AliasEq<I> {
     where
         I: 'i,
     {
-        R::new()
-            .and_then(|| self.alias.visit_with(visitor, outer_binder))
-            .and_then(|| self.ty.visit_with(visitor, outer_binder))
+        visitor.visit_domain_goal(self, outer_binder)
     }
 }
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -12,12 +12,32 @@ pub mod visitors;
 
 pub use visitors::VisitExt;
 
+/// A "result type" that can be returned from a visitor. Visitors pick
+/// an appropriate result type depending on what sort of operation they
+/// are doing. A common choice is `FindAny`, which indicates that the visitor
+/// is searching for something and that the visitor should stop once it is found.
 pub trait VisitResult: Sized {
     fn new() -> Self;
 
+    /// Returns true if this result is "complete" and we can stop visiting any
+    /// further parts of the term. This is used by `FindAny`, for example, to
+    /// stop the search after a match has been found.
     fn return_early(&self) -> bool;
     fn combine(self, other: Self) -> Self;
 
+    /// Convenience helper for use in writing `Visitor` impls. Returns `self`
+    /// if `return_early()` is true, but otherwise combines `self` with the
+    /// result of invoking `op`.
+    ///
+    /// If you have a struct with two fields, `foo` and `bar`, you can
+    /// thus write code like
+    ///
+    /// ```rust,ignore
+    /// self.foo.visit_with(visitor, outer_binder)
+    ///     .and_then(|| self.bar.visit_with(visitor, outer_binder))
+    /// ```
+    ///
+    /// and `bar` will only be visited if necessary.
     fn and_then(self, op: impl FnOnce() -> Self) -> Self {
         if self.return_early() {
             self
@@ -27,6 +47,8 @@ pub trait VisitResult: Sized {
     }
 }
 
+/// Unit type for a visitor indicates a "side-effecting" visitor that
+/// should visit an entire term.
 impl VisitResult for () {
     fn new() -> () {}
 
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -8,6 +8,7 @@ use crate::{
 
 mod binder_impls;
 mod boring_impls;
+pub mod visitors;
 
 pub trait VisitResult: Sized {
     fn new() -> Self;
@@ -15,17 +16,16 @@ pub trait VisitResult: Sized {
     fn and_then(self, op: impl FnOnce() -> Self) -> Self;
 }
 
-
 impl VisitResult for () {
     fn new() -> () {}
-    fn and_then(self, op: impl FnOnce() -> () ) -> () {
+    fn and_then(self, op: impl FnOnce() -> ()) -> () {
         op()
     }
 }
 
 /// A "visitor" recursively folds some term -- that is, some bit of IR,
-/// such as a `Goal`, and computes a value as a result. 
-/// 
+/// such as a `Goal`, and computes a value as a result.
+///
 ///
 /// To **apply** a visitor, use the `Visit::visit_with` method, like so
 ///
@@ -34,7 +34,7 @@ impl VisitResult for () {
 /// ```
 pub trait Visitor<'i, I: Interner>
 where
-I: 'i
+    I: 'i,
 {
     type Result: VisitResult;
 
@@ -219,7 +219,9 @@ pub trait Visit<I: Interner>: Debug {
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
-    ) -> R where I: 'i;
+    ) -> R
+    where
+        I: 'i;
 }
 
 /// For types where "visit" invokes a callback on the `visitor`, the
@@ -230,7 +232,9 @@ pub trait SuperVisit<I: Interner>: Visit<I> {
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
-    ) -> R where I: 'i;
+    ) -> R
+    where
+        I: 'i;
 }
 
 /// "visiting" a type invokes the `visit_ty` method on the visitor; this
@@ -241,7 +245,10 @@ impl<I: Interner> Visit<I> for Ty<I> {
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
-    ) -> R where I: 'i {
+    ) -> R
+    where
+        I: 'i,
+    {
         visitor.visit_ty(self, outer_binder)
     }
 }
@@ -255,7 +262,10 @@ where
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
-    ) -> R where I: 'i {
+    ) -> R
+    where
+        I: 'i,
+    {
         let interner = visitor.interner();
         match self.data(interner) {
             TyData::BoundVar(bound_var) => {
@@ -280,7 +290,10 @@ impl<I: Interner> Visit<I> for Lifetime<I> {
         &self,
         visitor: &mut dyn Visitor<'i, I, Result = R>,
         outer_binder: DebruijnIndex,
-    ) -> R where I: 'i {
+    ) -> R
+    where
+        I: 'i,
+    {
         visitor.visit_lifetime(self, outer_binder)
     }
 }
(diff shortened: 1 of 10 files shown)


chalk-ir/src/visit.rs
```diff
@@ -0,0 +1,404 @@
...
+        _outer_binder: DebruijnIndex,
+    ) -> Self::Result {
+        if self.forbid_free_placeholders() {
+            panic!("unexpected placeholder type `{:?}`", universe)
+        } else {
+            Self::Result::new()
+        }
+    }
+
+    fn visit_alias_eq(
```
nikomatsakis, line 162: Hmm, do we really want this accessor? I think we don't have similar one for other kinds of specific goals

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,341 @@
...
+impl<T: Visit<I>, I: Interner> Visit<I> for &T {
+    fn visit_with<'i, R: VisitResult>(
+        &self,
+        visitor: &mut dyn Visitor<'i, I, Result = R>,
+        outer_binder: DebruijnIndex,
+    ) -> R
+    where
+        I: 'i,
+    {
+        (**self).visit_with(visitor, outer_binder)
```
nikomatsakis, line 24: Nit: I'm not sure what we wrote elsewhere, but I've migrated to writing code like this as `T::visit_with(self, visitor, outer_binder)`. This removes the need to have a bunch of `**`, which I always find kind of hard to mentally track, and makes it quite clear that we are invoking the `Visitor` defined for the referent type `T`

nikomatsakis: (I've had e.g. infinite loops because I wrote `(*self).` etc...there is of course a litn for that now, but the `T::` form seems more clear and secure.)

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,341 @@
...
+        &self,
+        visitor: &mut dyn Visitor<'i, I, Result = R>,
+        outer_binder: DebruijnIndex,
+    ) -> R
+    where
+        I: 'i,
+    {
+        let mut result = R::new();
+        for e in self {
+            result = result.and_then(|| e.visit_with(visitor, outer_binder))
```
nikomatsakis, line 39: Hmm, so, the one problem here is that we are not giving the `and_then` the ability to stop iterating through the vector. I think we'd have to write this to "unfold" across the stack like so:

```rust
fn unroll<I: Interner, T: Visit<I>, R: VisitResult>(
    visitor: &mut dyn Visitor<'_, I, Result = R>,
    iter: impl Iterator<Item = T>
) -> R {
    let mut iterator = self.iter();
    match iterator.next() {
        None => R::new(),
        Some(elem) => elem.visit_with(visitor, outer_binder).and_then(|| unroll(visitor, iter))
    }
})
```

But seeing how nasty this is makes me wonder if `and_then` is *really* the right primitive, versus something that returns e.g. a `Result`-like enum that means "keep going", or perhaps adding a boolean to the result like `return_early`, so that the loop would be like:

```rust
for e in self {
    result = result.combine(e.visit_with(visitor, outer_binder));
    if result.return_early() { return result; }
}
result
```

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,341 @@
...
+        visitor: &mut dyn Visitor<'i, I, Result = R>,
+        outer_binder: DebruijnIndex,
+    ) -> R
+    where
+        I: 'i,
+    {
+        let interner = visitor.interner();
+        let mut result = R::new();
+
+        for p in self.iter(interner) {
```
nikomatsakis, line 132: same as the for loop above; we should probably make a helper for `for` loop iterator regardless

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,341 @@
...
```
nikomatsakis, line 152: same as the for loop above; we should probably make a helper for `for` loop iterator regardless

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,341 @@
...
+        _outer_binder: DebruijnIndex,
+    ) -> R
+    where
+        I: 'i,
+    {
+        R::new()
+    }
+}
+
+impl<I: Interner, T, L> Visit<I> for ParameterKind<T, L>
```
nikomatsakis, line 239: can this not be derived..?

chalk-solve/src/wf.rs
```diff
@@ -41,63 +42,63 @@ pub struct WfSolver<'db, I: Interner> {
...
-        }
+    fn interner(&self) -> &'i I {
+        self.interner
     }
-}
 
-impl<I: Interner> FoldInputTypes for Substitution<I> {
-    fn fold(&self, interner: &I, accumulator: &mut Vec<Ty<I>>) {
-        self.parameters(interner).fold(interner, accumulator)
+    fn visit_alias_eq(&mut self, alias_eq: &AliasEq<I>, outer_binder: DebruijnIndex) {
```
nikomatsakis, line 70: I guess this is why you added the `AliasEq` callback. Ah, I was imagining before that we would intercept `AliasEq` when we are visiting *goals*, but in this code we are going to visit `WhereClause` values that are outside of any goal or program clause, right? Hmm, that's annoying!

Something about having `visit_alias_eq` doesn't feel right, it's kind of a "random level" in the type hierarchy, and we've kept the "callback points" relatively contained up until now. I guess that having a callback either for `WhereClause` (going "up a level") or for `AliasTy` (going "down a level") feels potentially a bit better to me. An `AliasTy` would also get invoked when walking types, I imagine, which in some sense fits well -- it just adds one more category of type that gets a special callback. (But weird in that it also gets invoked from other contexts...)

That said, I think I'd be ok landing this "as is" with a kind of FIXME to resolve the right callback here -- it's a bit tied into how we are handling implied bounds (that's what this code is doing), and I also think the right answer might become more clear over time.

chalk-ir/src/visit/boring_impls.rs
Areredify: Derive is only for types that have one type parameter that either has an interner or is the interner. I don't know why is it so strict, but I guess there was a reason for that?

chalk-ir/src/visit/boring_impls.rs
Areredify: I thought about this when implementing, and I chose to do it that way instead of unrolling because it matches the behavior of the regular struct Visit.
```
fn visit_with(...) -> R {
  R::new()
      .and_then(|| a.visit()
                            .and_then(|| b.visit()
                                                  .and_then(|| c.visit())))
}
```
would be the struct visit rewritten as unrolling, which is kind of a weird mix of right and left folding.

Replacing `and_then` with `combine` + boolean sounds very annoying, since you would need to call it after each visit of a struct member too. I would just add an `return_early` method as a perf hack for iterator things, which is a little inconvenient since it would require `and_then` and `return_early` behavior to match (but when implementing `VisitResult` you will probably use `return_early` in `and_then` anyway, so who cares). Alternatively, we can hack the existing implementation by doing 
```
for e in self {
    let visited = false;
    result = result.and_then(|| { e.visit_with(visitor, outer_binder); visited = true; })
    if !visited {
       break
    }
}
```
If we propose a contract that `and_then` not calling the closure means early return, then it kinda feels less hacky.

Areredify:
Oh, and also I forgot to ask about the truncator, should I just replace it with a visitor that checks term size?

chalk-solve/src/wf.rs
Areredify: I changed callbacks a little (and added visitor for EnvElaborator), take a look please

chalk-ir/src/visit/boring_impls.rs
nikomatsakis: I feel like I'd prefer to have one mechanism, let's call it `return_early`. If we want, we can build the `and_then` mechanism on top of that readily enough:

```rust
fn and_then(self, op: impl FnOnce() -> Self) -> Self {
    if self.return_early() { self } else { self.combine(op()) }
}
```

That said, I'd probably change the deriving at least to use `return_early` instead of `and_then` because it'll be more efficient at runtime and requires less compiler tricks to compile.

chalk-ir/src/visit/boring_impls.rs
nikomatsakis: Oh. Well, it's almost always sufficient I guess -- and the more general pattern isn't as obvious -- but I think we could extend it readily enough. e.g., to any number of type parameters that are all `HasInterner`, which would apply here, right? But if it's only the one impl, maybe not worth it (I imagine this is why we didn't bother the first time)

chalk-solve/src/wf.rs
nikomatsakis: Will do

nikomatsakis:
@Areredify re: truncator, yes, that'd be the idea -- we just want to know if *any* term exceeds a certain size

chalk-solve/src/wf.rs
```diff
@@ -41,63 +42,70 @@ pub struct WfSolver<'db, I: Interner> {
...
-impl<I: Interner> FoldInputTypes for Substitution<I> {
-    fn fold(&self, interner: &I, accumulator: &mut Vec<Ty<I>>) {
-        self.parameters(interner).fold(interner, accumulator)
+    fn visit_where_clause(&mut self, where_clause: &WhereClause<I>, outer_binder: DebruijnIndex) {
+        match where_clause {
+            WhereClause::AliasEq(alias_eq) => {
+                TyData::Alias(alias_eq.alias.clone())
+                    .intern(self.interner)
+                    .visit_with(self, outer_binder);
+                alias_eq.ty.visit_with(self, outer_binder);
```
nikomatsakis, line 76: I think this could be 

```rust
if let WhereClause::AliasEq(alias_eq) = self {
TyData::Alias(alias_eq.alias.clone())
                    .intern(self.interner)
                    .visit_with(self, outer_binder);
}

where_clause.super_visit_with(self, outer_binder);
```

Although maybe this code is better, actually, in the sense of being more robust against future changes that might add new cases we should consider.

chalk-solve/src/wf.rs
```diff
@@ -41,63 +42,70 @@ pub struct WfSolver<'db, I: Interner> {
...
+struct InputTypeCollector<'i, I: Interner> {
+    types: Vec<Ty<I>>,
+    interner: &'i I,
 }
 
-impl<T: FoldInputTypes> FoldInputTypes for [T] {
-    fn fold(&self, interner: &T::Interner, accumulator: &mut Vec<Ty<T::Interner>>) {
-        for f in self {
-            f.fold(interner, accumulator);
+impl<'i, I: Interner> InputTypeCollector<'i, I> {
```
nikomatsakis, line 50: Let's add a helper method like

```rust
fn types_in(interner: &'i I, value: impl Visit<I>) -> Vec<Ty<I>> {
    let mut collector = InputTypeCollector { types: vec![], interner };
    value.visit_with(&mut collector, DebruijnIndex::INNERMOST);
    collector.types
}
```

chalk-solve/src/wf.rs
```diff
@@ -198,12 +165,13 @@ where
...
-                    let mut input_types = Vec::new();
+                    let mut type_collector = InputTypeCollector::new(gb.interner());
+
                     // ...in a field type...
-                    fields.fold(gb.interner(), &mut input_types);
+                    fields.visit_with(&mut type_collector, DebruijnIndex::INNERMOST);
                     // ...in a where clause.
-                    where_clauses.fold(gb.interner(), &mut input_types);
-                    gb.all(input_types.into_iter().map(|ty| ty.well_formed()))
+                    where_clauses.visit_with(&mut type_collector, DebruijnIndex::INNERMOST);
```
nikomatsakis, line 173: then this can be

```rust
let types = InputTypeCollector::types_in(gb.interner(), (&fields, &where_clauses));
```

chalk-solve/src/wf.rs
```diff
@@ -287,13 +255,14 @@ fn impl_header_wf_goal<I: Interner>(
                 // we would retrieve `HashSet<K>`, `Box<T>`, `Vec<Box<T>>`, `(HashSet<K>, Vec<Box<T>>)`.
                 // We will have to prove that these types are well-formed (e.g. an additional `K: Hash`
                 // bound would be needed here).
-                let mut input_types = Vec::new();
-                where_clauses.fold(interner, &mut input_types);
+                let mut type_collector = InputTypeCollector::new(interner);
+                where_clauses.visit_with(&mut type_collector, DebruijnIndex::INNERMOST);
```
nikomatsakis, line 259:
```rust
let types = InputTypeCollector::types_in(gb.interner(), &where_clauses);
```

chalk-solve/src/wf.rs
```diff
@@ -329,10 +298,11 @@ fn impl_wf_environment<'i, I: Interner>(
     //     // Inside here, we can rely on the fact that `K: Hash` holds
     // }
     // ```
-    let mut header_input_types = Vec::new();
-    trait_ref.fold(interner, &mut header_input_types);
+    let mut type_collector = InputTypeCollector::new(interner);
+    trait_ref.visit_with(&mut type_collector, DebruijnIndex::INNERMOST);
```
nikomatsakis, line 302:
```rust
let types = InputTypeCollector::types_in(gb.interner(), trait_ref);
```

chalk-solve/src/wf.rs
```diff
@@ -433,11 +403,12 @@ fn compute_assoc_ty_goal<I: Interner>(
                         .cloned()
                         .map(|qwc| qwc.into_from_env_goal(interner)),
                     |gb| {
-                        let mut input_types = Vec::new();
-                        value_ty.fold(interner, &mut input_types);
+                        let mut type_collector = InputTypeCollector::new(interner);
+                        value_ty.visit_with(&mut type_collector, DebruijnIndex::INNERMOST);
```
nikomatsakis, line 407:
```rust
let types = InputTypeCollector::types_in(gb.interner(), value_ty);
```

chalk-solve/src/wf.rs
nikomatsakis: (In case unclear, just leave as is I think)

Areredify:
I'll file a separate pr with the truncator, let's merge this first

chalk-ir/src/visit/visitors.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.found
+    }
+    fn combine(self, other: Self) -> Self {
+        FindAny {
+            found: self.found || other.found,
+        }
+    }
+}
+
+pub struct FindFreeVarsVisitor<'i, I: Interner> {
```
nikomatsakis, line 41: I feel like this should be a private impl detail of the `has_free_vars` function

chalk-ir/src/visit/visitors.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl<T, I: Interner> VisitExt<I> for T where T: Visit<I> {}
+
+#[derive(Clone, Copy, Debug)]
+pub struct FindAny {
+    pub found: bool,
+}
+
+impl FindAny {
```
nikomatsakis, line 20: Maybe make an associated constant `const FOUND: FindAny = true;`

chalk-ir/src/visit/visitors.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn interner(&self) -> &'i I {
+        self.interner
+    }
+
+    fn visit_free_var_ty(
+        &mut self,
+        _bound_var: BoundVar,
+        _outer_binder: DebruijnIndex,
+    ) -> Self::Result {
+        FindAny { found: true }
```
nikomatsakis, line 61: then this can be `FindAny::FOUND`

chalk-ir/src/visit/visitors.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    fn visit_free_var_lifetime(
+        &mut self,
+        _bound_var: BoundVar,
+        _outer_binder: DebruijnIndex,
+    ) -> Self::Result {
+        FindAny { found: true }
```
nikomatsakis, line 69: and this

chalk-ir/src/visit.rs
```diff
@@ -0,0 +1,421 @@
...
+mod binder_impls;
+mod boring_impls;
+pub mod visitors;
+
+pub use visitors::VisitExt;
+
+pub trait VisitResult: Sized {
+    fn new() -> Self;
+
+    fn return_early(&self) -> bool;
```
nikomatsakis, line 18: Let's add a comment:

---

Returns true if this result is "complete" and we can stop visiting any further parts of the term. This is used by `FindAny`, for example, to stop the search after a match has been found.

chalk-ir/src/visit.rs
```diff
@@ -0,0 +1,421 @@
...
+    PlaceholderIndex, ProgramClause, Ty, TyData, WhereClause,
+};
+
+mod binder_impls;
+mod boring_impls;
+pub mod visitors;
+
+pub use visitors::VisitExt;
+
+pub trait VisitResult: Sized {
```
nikomatsakis, line 15: Let's add a comment:

---

A "result type" that can be returned from a visitor. Visitors pick an appropriate result type depending on what sort of operation they are doing. A common choice is `FindAny`, which indicates that the visitor is searching for something and that the visitor should stop once it is found.

chalk-ir/src/visit.rs
```diff
@@ -0,0 +1,421 @@
...
+    fn new() -> Self;
+
+    fn return_early(&self) -> bool;
+    fn combine(self, other: Self) -> Self;
+
+    fn and_then(self, op: impl FnOnce() -> Self) -> Self {
```
nikomatsakis, line 21: Comment:

Convenience helper for use in writing `Visitor` impls. Returns `self` if `return_early()` is true, but otherwise combines `self` with the result of invoking `op`.

If you have a struct with two fields, `foo` and `bar`, you can thus write code like

```rust,ignore
self.foo.visit_with(visitor, outer_binder)
    .and_then(|| self.bar.visit_with(visitor, outer_binder))
```

and `bar` will only be visited if necessary.

chalk-ir/src/visit.rs
```diff
@@ -0,0 +1,421 @@
...
+        if self.return_early() {
+            self
+        } else {
+            self.combine(op())
+        }
+    }
+}
+
+impl VisitResult for () {
```
nikomatsakis, line 30: Comment:

Unit type for a visitor indicates a "side-effecting" visitor that should visit an entire term.

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,380 @@
...
+
+use crate::{
+    AssocTypeId, DebruijnIndex, Goals, ImplId, Interner, Parameter, ParameterKind,
+    PlaceholderIndex, ProgramClause, ProgramClauseData, ProgramClauses, QuantifierKind, StructId,
+    Substitution, SuperVisit, TraitId, UniverseIndex, Visit, VisitResult, Visitor,
+};
+use chalk_engine::{context::Context, ExClause, FlounderedSubgoal, Literal};
+use std::{marker::PhantomData, sync::Arc};
+
+pub fn visit_iter<'i, T, I, IT, R>(
```
nikomatsakis, line 15: Comment:

Convenience function to visit all the items in the iterator `it`.

chalk-ir/src/visit/boring_impls.rs
```diff
@@ -0,0 +1,380 @@
...
+    it: IT,
```
nikomatsakis, line 16: Personally, I would use `impl Iterator<Item = T>`

nikomatsakis:
> I'll file a separate pr with the truncator, let's merge this first

Agreed -- shall we file an issue to track it?

Areredify:
I gotta up my comments game 😅

chalk-ir/src/visit.rs
```diff
@@ -12,12 +12,32 @@ pub mod visitors;
 
 pub use visitors::VisitExt;
 
+// A "result type" that can be returned from a visitor. Visitors pick
```
nikomatsakis, line 15: Argh -- text seems good, but these should be "doc comments", meaning they should use `///` so they show up in the rustdoc results

Areredify: done 😅

PR #374, opened 2020-04-06
add Sized well-known impl and wf check

### LINKED ISSUE #261: add a "well-known" Sized trait to `chalk_solve`
I think the `chalk-solve` crate ought to start encoding rules for "well-known" traits that require compiler support. A good place to start is the `Sized` trait. I'm opening this issue to remember the problem, it will be extended with mentoring instructions.

Areredify:
EDIT: doesnt actually close \\#261, we need #368 to land first and address it

### DIFF
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -15,11 +15,6 @@ use chalk_ir::{
 };
 use std::iter;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum LangItem {
-    SizedTrait,
-}
-
 /// Identifier for an "associated type value" found in some impl.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct AssociatedTyValueId<I: Interner>(pub I::DefId);
@@ -136,7 +131,7 @@ pub struct TraitDatum<I: Interner> {
 
 /// A list of the traits that are "well known" to chalk, which means that
 /// the chalk-solve crate has special, hard-coded impls for them.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash)]
 pub enum WellKnownTrait {
     SizedTrait,
     CopyTrait,
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -15,8 +15,10 @@ use chalk_ir::{
 };
 use std::iter;
 
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum LangItem {}
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum LangItem {
+    SizedTrait,
+}
 
 /// Identifier for an "associated type value" found in some impl.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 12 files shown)


chalk-solve/src/wf.rs
```diff
@@ -464,3 +478,27 @@ fn compute_assoc_ty_goal<I: Interner>(
         },
     ))
 }
+
+fn compute_struct_sized_constraint<I: Interner>(
```
nikomatsakis, line 482: We should add a comment to this function

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -16,8 +18,44 @@ pub fn add_builtin_program_clauses<I: Interner>(
...
 
     match well_known {
-        WellKnownTrait::SizedTrait => { /* TODO */ }
+        WellKnownTrait::SizedTrait => add_sized_program_clauses(db, builder, trait_ref, ty),
         WellKnownTrait::CopyTrait => { /* TODO */ }
         WellKnownTrait::CloneTrait => { /* TODO */ }
     }
 }
+
+fn add_sized_program_clauses<I: Interner>(
```
nikomatsakis, line 27: I think we should make `builtin_traits/sized.rs` and put this function in there

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -16,8 +18,44 @@ pub fn add_builtin_program_clauses<I: Interner>(
...
+        TyData::Apply(ApplicationTy {
+            name: TypeName::Struct(struct_id),
+            substitution,
+        }) => (*struct_id, substitution),
+        _ => return,
+    };
+
+    let struct_datum = db.struct_datum(struct_id);
+
+    if struct_datum.binders.map_ref(|b| b.fields.is_empty()).value {
```
nikomatsakis, line 45: Comment: 

---

Structs with no fields are always `Sized`

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -16,8 +18,44 @@ pub fn add_builtin_program_clauses<I: Interner>(
...
+        builder.push_fact(trait_ref.clone());
+        return;
+    }
+
+    let last_field_ty = struct_datum
```
nikomatsakis, line 50: Let's add a comment here -- something like this:

---

To check if a struct type `S<..>` is `Sized`, we only have to look at its last field. This is because the WF checks for structs require that all the other fields must be `Sized`.

Areredify:
Addressed the nits and rebased

chalk-solve/src/lib.rs
```diff
@@ -76,6 +76,9 @@ pub trait RustIrDatabase<I: Interner>: Debug {
         false
     }
 
+    /// Returns id of a trait lang item, if found
+    fn require_lang_item(&self, lang_item: LangItem) -> TraitId<I>;
```
nikomatsakis, line 80: i.e., we can rename this to `well_known_trait_id` or something

chalk-rust-ir/src/lib.rs
```diff
@@ -15,8 +15,10 @@ use chalk_ir::{
 };
 use std::iter;
 
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum LangItem {}
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
Areredify: Should I remove it altogether or leave it as it was before me?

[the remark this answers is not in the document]
nikomatsakis: Remove it altogether I think, I forgot it was there

Areredify:
All done and rebased

Areredify:
@nikomatsakis wait, why did I put the constraint check into the WfSolver? It should probably be in ToProgramClauses for StructDatum, right?

Areredify:
made a [zulip](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/extralogical.20reasoning) thread because I have many questions about this

PR #377, opened 2020-04-07
Rename field `goals` of type `Goals` to `interned`.

### LINKED ISSUE #376: rename field `goals` of type `Goals` to `interned`
A very simple first PR: we should rename the [field `goals`](http://rust-lang.github.io/chalk/chalk_ir/struct.Goals.html#structfield.goals) to `interned`, to mach the other types.

Split out from #370.

crlf0710:
Closes #376


nikomatsakis:
Thanks!

Issue #369, opened 2020-03-31
move remaining vectors, boxes in chalk-ir into associated types on `Interner`


nikomatsakis:
In general, we have been trying to remove hard-coded vectors and boxes from the chalk-ir APIs, and moving them instead to `Interner` associated types. See [the chalk book](http://rust-lang.github.io/chalk/book/types/role_of_interner.html) for more background on the `Interner` trait and its design. 

This issue tracks the removal of the final vectors etc:

* [x] `Vec<ProgramClause<I>>` in `Environment` -- introduce `ProgramClauses` -- done in #370
* [x] `Vec<QuantifiedWhereClause<I>>` in `DyTy` -- introduce `QuantifiedWhereClauses` -- done in #379
* [x] `Vec<Parameter>` in `Fn` (this should be changed to `Substitution`) -- done in #368 
* [ ] `Vec<ParameterKind<()>>` in `Binders`, `BindersIntoIterator` -- introduce `ParameterKinds` -- #386 
* [ ] `Vec<ParameterKind<Universe>>` in `Canonical` -- hmm, this one requires a bit of thought, because I think we want to reconcile other bits of canonicalization -- #386 

Here are some example PRs showing how this removal can be done:

* https://github.com/rust-lang/chalk/pull/316 converted a `Vec<Parameter>` into the `Substitution<I>` type
* https://github.com/rust-lang/chalk/pull/325 replaced `Vec<Goal<I>>` with `Goals<I>`

This issue has been assigned to @crlf0710 via [this comment](https://github.com/rust-lang/chalk/issues/369#issuecomment-612989155).

crlf0710:
Will have a try this weekend.

crlf0710:
Edit: Seems https://github.com/rust-lang/chalk/pull/370 is in flight.

nikomatsakis:
@crlf0710 there are still other issues to claim -- e.g., `QuantifiedWhereClauses`, `ParameterKinds`, etc. Feel free to claim one of those.

crlf0710:
Changed `Vec<Parameter>` to `Substitution` in `Fn` in #378. 

I'll have a try on `Vec<QuantifiedWhereClause<I>>` next.

crlf0710:
Introduced `QuantifiedWhereClauses` in #379.

crlf0710:
Given the rest a try in #386

PR #379, opened 2020-04-07
Intern `Vec<QuantifiedWhereClause<I>>`.

crlf0710:
cc #369 .

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -543,7 +543,7 @@ impl DebruijnIndex {
 /// level of binder).
 #[derive(Clone, PartialEq, Eq, Hash, Fold)]
 pub struct DynTy<I: Interner> {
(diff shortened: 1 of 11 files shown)


chalk-ir/src/lib.rs
```diff
@@ -1033,6 +1033,62 @@ impl<I: Interner> QuantifiedWhereClause<I> {
...
+        use crate::cast::Caster;
+        QuantifiedWhereClauses {
+            interned: I::intern_quantified_where_clauses(
+                interner,
+                clauses.into_iter().casted(interner),
+            ),
+        }
+    }
+
+    pub fn from_fallible<E>(
```
nikomatsakis, line 1063: This is getting to be an annoying amount of boilerplate -- not to be addressed in this PR, but I wonder if we can cleanup and encapsulate this pattern better.

PR #384, opened 2020-04-10
refactored binders to have a private value field

### LINKED ISSUE #375: refactor Binders API to have a private value field

Areredify:
Closes #375

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -1144,7 +1144,7 @@ impl<I: Interner> HasInterner for AliasEq<I> {
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct Binders<T> {
     pub binders: Vec<ParameterKind<()>>,
-    pub value: T,
+    value: T,
 }
 
 impl<T: HasInterner> HasInterner for Binders<T> {
@@ -1152,6 +1152,35 @@ impl<T: HasInterner> HasInterner for Binders<T> {
 }
 
 impl<T> Binders<T> {
+    pub fn new(binders: Vec<ParameterKind<()>>, value: T) -> Self {
+        Self { binders, value }
+    }
+
+    /// Skips the binder and returns the "bound" value. This is a
+    /// risky thing to do because it's easy to get confused about
+    /// De Bruijn indices and the like. It is usually better to
+    /// discharge the binder using `no_bound_vars` or something
+    /// like that. `skip_binder` is only valid when you are either
+    /// extracting data that has nothing to do with bound vars, you
+    /// are doing some sort of test that does not involve bound
+    /// regions, or you are being very careful about your depth
+    /// accounting.
+    ///
+    /// Some examples where `skip_binder` is reasonable:
+    ///
+    /// - extracting the `TraitId` from a TraitRef;
+    /// - checking if there are any fields in a StructDatum
+    pub fn skip_binders(&self) -> &T {
+        &self.value
+    }
+
+    pub fn as_ref(&self) -> Binders<&T> {
+        Binders {
+            binders: self.binders.clone(),
+            value: &self.value,
+        }
+    }
+
     pub fn map<U, OP>(self, op: OP) -> Binders<U>
     where
         OP: FnOnce(T) -> U,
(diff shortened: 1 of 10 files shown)


chalk-solve/src/solve/slg/resolvent.rs
```diff
@@ -485,7 +485,7 @@ impl<'i, I: Interner> Zipper<'i, I> for AnswerSubstitutor<'i, I> {
         T: Zip<I> + Fold<I, Result = T>,
     {
         self.outer_binder.shift_in();
-        Zip::zip_with(self, &answer.value, &pending.value)?;
+        Zip::zip_with(self, &answer.skip_binders(), &pending.skip_binders())?;
```
detrumi, line 488:
```suggestion
        Zip::zip_with(self, answer.skip_binders(), pending.skip_binders())?;
```

chalk-solve/src/coherence/solve.rs
```diff
@@ -85,9 +85,12 @@ impl<I: Interner> CoherenceSolver<'_, I> {
 
         let interner = self.db.interner();
 
+        let (lhs_binders, lhs_bound) = lhs.binders.as_ref().into();
+        let (rhs_binders, rhs_bound) = rhs.binders.as_ref().into();
```
detrumi, line 89: This is slightly misleading, because `as_ref` is cloning the binders, whereas it wasn't before.

Areredify: binders were cloned later anyway, but as_ref is kind of a bad name (I just copied it from rustc), should I change it?

Areredify: Oh, you mean the names, hmm, ok

chalk-ir/src/lib.rs
```diff
@@ -1194,11 +1219,38 @@ impl<T> Binders<T> {
...
+    /// no vars bound by this binder. (In other words, if this binder --
+    /// and indeed any enclosing binder -- doesn't bind anything at
+    /// all.) Otherwise, returns `None`.
+    ///
+    /// (One could imagine having a method that just unwraps a single
+    /// binder, but permits late-bound vars bound by enclosing
+    /// binders, but that would require adjusting the debruijn
+    /// indices, and given the shallow binding structure we often use,
+    /// would not be that useful.)
+    pub fn no_bound_vars<I: Interner>(&self, interner: &I) -> Option<&T>
```
jackh726, line 1232: Do we actually use this anywhere? I don't see it. If not, let's just not add it. I know rustc has it, but I'm not sure that's a valid reason on its own. Also remember to update the doc comment for skip_binders to not mention this.

chalk-solve/src/coherence/solve.rs
jackh726: I think `as_ref` is fine. Just be sure the doc comments are clear. (The binders will almost certainly end up being an interned type instead of a vec anyways, so the clone should be cheap).

chalk-rust-ir/src/lib.rs
```diff
@@ -413,8 +408,7 @@ impl<I: Interner> AssociatedTyDatum<I> {
     /// these quantified where clauses are in the scope of the
     /// `binders` field.
     pub fn bounds_on_self(&self, interner: &I) -> Vec<QuantifiedWhereClause<I>> {
-        let Binders { binders, value } = &self.binders;
-
+        let (binders, assoc_ty_datum) = self.binders.as_ref().into();
```
jackh726, line 411: This does technically add a clone to binders. But I don't think it matters.

chalk-solve/src/coherence/solve.rs
```diff
@@ -268,11 +261,10 @@ impl<I: Interner> CoherenceSolver<'_, I> {
...
 }
 
-fn params<'a, I: Interner>(interner: &I, impl_datum: &'a ImplDatum<I>) -> &'a [Parameter<I>] {
-    impl_datum
-        .binders
-        .value
-        .trait_ref
-        .substitution
-        .parameters(interner)
+fn params<'a, I: Interner>(
```
jackh726, line 264: This is such a short function now. And since we have that safety comment about skipping binders, let's just inline this where it's used.

Issue #363, opened 2020-03-27
Extend Chalk with support for well-known, builtin traits


nikomatsakis:
This is a meta issue for the work of extending chalk to support the various built-in traits like `Sized`, `Copy`, and so forth. 

* [ ] Extend `TraitDefn` with concept of "well-known trait" identifier #356 
* [ ] Add basic support for the `Sized` trait https://github.com/rust-lang/chalk/issues/261
* [ ] Add support for a built-in `Copy` trait
* [ ] Add support for a built-in `Clone` trait
* [ ] Add support for built-in `Fn` traits
* [ ] Add support for built-in `Unsize` traits

nikomatsakis:
I would also say that this issue is fairly 'parallelizable', so if others want to pick up some aspects of it, that's not a problem. e.g. I think we could add code for `Copy` and the like. Maybe it's a good idea to break those out into separate issues and write-up some mentoring instructions?

nikomatsakis:
Although they may be somewhat blocked on https://github.com/rust-lang/chalk/pull/371 -- having more details of Rust types would be useful.

Areredify:
So if someone wants to pick it up, I am done with Copy, Drop and Clone (as far as I understood Clone introduces no hard coded impls and/or wf checks?) modulo #371 of course, but it will be easy to address when it comes to it.

Areredify:
Im already working on `CoerceUnsized`

memoryleak47:
I'll try `Unpin`

PR #385, opened 2020-04-10
add `Copy`, `Clone` and `Drop` traits support

Areredify:
So I raised some concerns about this implementation in [this zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/extralogical.20reasoning), how grounded are they?

cc #363

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -415,23 +415,32 @@ impl<I: Interner> ToProgramClauses<I> for TraitDatum<I> {
             // conditions.
             let type_parameters: Vec<_> = trait_ref.type_parameters(interner).collect();
 
-            // Add all cases for potential downstream impls that could exist
-            for i in 0..type_parameters.len() {
-                builder.push_clause(
-                    trait_ref.clone(),
-                    where_clauses
-                        .iter()
-                        .cloned()
-                        .casted(interner)
-                        .chain(iter::once(DomainGoal::Compatible(()).cast(interner)))
-                        .chain((0..i).map(|j| {
-                            DomainGoal::IsFullyVisible(type_parameters[j].clone()).cast(interner)
-                        }))
-                        .chain(iter::once(
-                            DomainGoal::DownstreamType(type_parameters[i].clone()).cast(interner),
-                        ))
-                        .chain(iter::once(GoalData::CannotProve(()).intern(interner))),
-                );
+            // Drop trait can't have downstream implementation because it can only
+            // be implemented with the same genericity as the struct definition,
+            // i.e. Drop implementation for `struct S<T: Eq> {}` is forced to be
+            // `impl Drop<T: Eq> for S<T> { ... }`. That means that orphan rules
+            // prevent Drop from being implemented in downstream crates.
+            if self.well_known != Some(WellKnownTrait::DropTrait) {
+                // Add all cases for potential downstream impls that could exist
+                for i in 0..type_parameters.len() {
+                    builder.push_clause(
+                        trait_ref.clone(),
+                        where_clauses
+                            .iter()
+                            .cloned()
+                            .casted(interner)
+                            .chain(iter::once(DomainGoal::Compatible(()).cast(interner)))
+                            .chain((0..i).map(|j| {
+                                DomainGoal::IsFullyVisible(type_parameters[j].clone())
+                                    .cast(interner)
+                            }))
+                            .chain(iter::once(
+                                DomainGoal::DownstreamType(type_parameters[i].clone())
+                                    .cast(interner),
+                            ))
+                            .chain(iter::once(GoalData::CannotProve(()).intern(interner))),
+                    );
+                }
             }
 
             // Orphan rules:
--- a/chalk-solve/src/wf.rs
+++ b/chalk-solve/src/wf.rs
@@ -164,7 +164,8 @@ where
             let interner = gb.interner();
 
             // struct is well-formed in terms of Sized
-            let sized_constraint_goal = compute_struct_sized_constraint(gb.db(), fields);
+            let sized_constraint_goal =
+                WellKnownConstraints::struct_sized_constraint(gb.db(), fields);
 
             // (FromEnv(T: Eq) => ...)
             gb.implies(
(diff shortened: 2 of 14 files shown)


chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
-/// Computes a goal to prove Sized constraints on a struct definition.
-/// Struct is considered well-formed (in terms of Sized) when it either
-/// has no fields or all of it's fields except the last are proven to be Sized.  
-fn compute_struct_sized_constraint<I: Interner>(
-    db: &dyn RustIrDatabase<I>,
-    fields: &[Ty<I>],
-) -> Option<Goal<I>> {
-    if fields.len() <= 1 {
-        return None;
+struct WellKnownConstraints {}
```
nikomatsakis, line 495: Can you add some comments to this struct and its methods below expressing what they are doing?

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+
+impl WellKnownConstraints {
```
nikomatsakis, line 497: I think maybe this should be called `WfWellKnownGoals` or something? It's basically specifying the well-formedness goals that must be proven for well-known traits?

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
-    let interner = db.interner();
+    pub fn outside_impl<I: Interner>(
+        db: &dyn RustIrDatabase<I>,
+        impl_datum: &ImplDatum<I>,
+    ) -> Option<Goal<I>> {
+        let interner = db.interner();
+
+        match db.trait_datum(impl_datum.trait_id()).well_known? {
+            // You can't add a manual implementation of Sized
+            WellKnownTrait::SizedTrait => Some(GoalData::CannotProve(()).intern(interner)),
```
nikomatsakis, line 518: This seems like it will ultimately have to change, since the resulting UX would be terrible, no?

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+                        )
+                    },
+                )
+            });
+
+        let impl_self_ty = impl_datum
+            .binders
+            .map_ref(|b| b.trait_ref.self_type_parameter(interner));
+
+        let eq_goal = gb.forall(
```
nikomatsakis, line 694: it'd be nice to have some inline comments saying what's happening here, I think

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+
+        let struct_datum = db.struct_datum(struct_id);
+        let struct_name = struct_datum.name(interner);
+
+        let impl_fields = impl_datum
+            .binders
+            .map_ref(|v| (&v.trait_ref, &v.where_clauses));
+
+        let implied_by_struct_def_goal =
+            gb.forall(&impl_fields, (), |gb, _, (trait_ref, where_clauses), ()| {
```
nikomatsakis, line 672: I'd appreciate inline comments... something like

---

// forall<ImplP1...ImplPn> { .. }

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+                let interner = gb.interner();
+
+                gb.implies(
+                    iter::once(
```
nikomatsakis, line 676: // WF(SelfType) => ...

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+                        FromEnv::Ty(trait_ref.self_type_parameter(interner))
+                            .cast::<DomainGoal<I>>(interner),
+                    ),
+                    |gb| {
+                        gb.all(
```
nikomatsakis, line 681: // All(ImplWhereClauses)

chalk-solve/src/wf.rs
nikomatsakis: Something like

// forall<StructP1..StructPN> {...}

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+                let interner = gb.interner();
+
+                let def_struct: Ty<I> = ApplicationTy {
+                    name: struct_name,
+                    substitution,
+                }
+                .cast(interner)
+                .intern(interner);
+
+                gb.exists(
```
nikomatsakis, line 707: // exists<ImplP1..ImplPn> {...}

chalk-solve/src/wf.rs
```diff
@@ -486,29 +492,233 @@ fn compute_assoc_ty_goal<I: Interner>(
...
+                    &impl_self_ty,
+                    def_struct,
+                    |gb, _, impl_struct, def_struct| {
+                        let interner = gb.interner();
+                        GoalData::EqGoal(EqGoal {
```
nikomatsakis, line 712: // StructName<StructP1..StructPn> = ImplSelfType

chalk-solve/src/wf.rs
Areredify: By UX you mean writing tests and using repl?

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -415,23 +415,32 @@ impl<I: Interner> ToProgramClauses<I> for TraitDatum<I> {
...
-                            DomainGoal::DownstreamType(type_parameters[i].clone()).cast(interner),
-                        ))
-                        .chain(iter::once(GoalData::CannotProve(()).intern(interner))),
-                );
+            // Drop trait can't have downstream implementation because it can only
+            // be implemented with the same genericity as the struct definition,
+            // i.e. Drop implementation for `struct S<T: Eq> {}` is forced to be
+            // `impl Drop<T: Eq> for S<T> { ... }`. That means that orphan rules
+            // prevent Drop from being implemented in downstream crates.
+            if self.well_known != Some(WellKnownTrait::DropTrait) {
```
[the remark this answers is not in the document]
Areredify: No, Copy is a regular trait in that regard

chalk-solve/src/wf.rs
Areredify: `CannotProve` is not the best way to report errors, but it I guess we'll change that when we add proper error reporting

nikomatsakis: By UX I mean "when rustc runs this code" -- the vision eventually is that rustc will run these WF goals and report errors from the results. But I agree we can address it later.

chalk-solve/src/wf.rs
```diff
@@ -609,7 +617,7 @@ impl WellKnownConstraints {
             ..
         }) = impl_datum
             .binders
-            .value
+            .skip_binders()
             .trait_ref
             .self_type_parameter(interner)
```
nikomatsakis, line 622: Side note that I think we should add some accessor here to `ImplDatum` that encapsulates this `skip_binders` call. Something like

```rust
/// If the self-type for this impl is a struct, returns its id.
fn self_type_struct_id(&self) -> Option<StructId>
```

I think it's probably useful in other places (e.g., the auto-trait code potentially?) and regardless I generally prefer to have `skip_binders` logic encapsulated so it be more easily reviewed. But maybe we should add it as a separate "cleanup item", because I think there are other `skip_binders` calls scattered about that could use similar treatment.

Areredify: I agree, we have a number of "get trait id" and "get struct id" uses of `skip_binders()` that can be moved to helper methods

chalk-solve/src/clauses/program_clauses.rs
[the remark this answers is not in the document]
jackh726: Ah, this makes sense.

PR #392, opened 2020-04-13
Stop leaking truncated goal

### LINKED ISSUE #306: deferred: can we / should we truncate the environment, and how to avoid loops?
In https://github.com/rust-lang/chalk/pull/294, we opted to stop truncating the environment. The problem was due to an infinite loop that arose in some rust-analyzer cases. The scenario [was described on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/design.20meeting.202019.2E12.2E09/near/182989871) recently but is also covered in #280 and #289.

In short, we would truncate a projection `Foo::Bar` from the environment and replace it with an inference variable `?X`. When we get back an answer, we would unify `Foo::Bar` with the value inferred for `?X` -- which would up being a fresh inference variable. This would create a projection goal, but that projection goal includes the same environment that we started with, and hence that same environment would get truncated again. This would result in the cycle beginning anew.

For now, we opted to just stop truncating the environment, because it isn't really needed at the moment -- Rust doesn't generate queries that grow the environment in indefinite ways. But we may want to revisit this with another fix, or -- depending on how we resolve normalization -- this may become a non-issue in the future. 

This issue has been assigned to @zaharidichev via [this comment](https://github.com/rust-lang/chalk/issues/306#issuecomment-612038103).

zaharidichev:
Closes #306 

As noted in this [Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/Contribution.20to.20chalk/near/193511696), it makes sense to return `Floundered` if we were to truncate and avoid leaking the truncated result. If the approach seems fine, I will make sure to change the comments where needed.

### DIFF
--- a/tests/test/projection.rs
+++ b/tests/test/projection.rs
@@ -712,7 +712,7 @@ fn rust_analyzer_regression() {
                     PI: ParallelIterator
                 }
             }
-        } yields_all[SolverChoice::slg(4, None)] {
+        } yields_all[SolverChoice::slg(5, None)] {
             "substitution [], lifetime constraints []"
         }
     }
--- a/tests/test/projection.rs
+++ b/tests/test/projection.rs
@@ -712,8 +712,8 @@ fn rust_analyzer_regression() {
                     PI: ParallelIterator
                 }
             }
-        } yields_all[SolverChoice::slg(5, None)] {
-            "substitution [], lifetime constraints []"
+        } yields_first[SolverChoice::slg(4, None)] {
+            "Floundered"
         }
     }
 }
--- a/chalk-solve/src/solve/slg.rs
+++ b/chalk-solve/src/solve/slg.rs
@@ -348,9 +348,7 @@ impl<I: Interner> TruncatingInferenceTable<I> {
 
 impl<I: Interner> context::TruncateOps<SlgContext<I>> for TruncatingInferenceTable<I> {
     fn goal_needs_truncation(&mut self, interner: &I, subgoal: &InEnvironment<Goal<I>>) -> bool {
-        // We only want to check the goal itself. We keep the environment intact.
-        // See rust-lang/chalk#280
-        truncate::needs_truncation(interner, &mut self.infer, self.max_size, &subgoal.goal)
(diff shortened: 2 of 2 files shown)


nikomatsakis:
The main change here is moving from `Option` to `Result`? Seems fine to me -- though it needs to be rebased, I'm not sure if this was invalidated by the work to use a visitor for truncation.

zaharidichev:
Yes I will rebase. Most likely it was invalidated..

jackh726:
@zaharidichev yes, I think these changes are superseded by the previous changes.

I think I was a bit unclear on Zulip. The primary change (to close #306) would be change it so we return `true` here https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/solve/slg.rs#L353 when the environment needs truncation. (The comment there is no longer relevant since that error only matters if we're actually *return* a truncated goal.) This change will *probably* cause some tests to fail (solution -> ambiguous), but depending on how bad it is, that *might* be okay, given that the alternative is no protections against environment size.

If you want to make that change, feel free to either do it in this PR or close this and open a new one.

If it turns out that there is too much breakage with truncating the environment, then #306 goes back to being a design question. (e.g. How do we handle very large environments?)

zaharidichev:
@jackh726 that makes sense. Thanks a bunch. Made the changes here in place. This only test that fails is the [rust analyzer regression](https://github.com/rust-lang/chalk/blob/master/tests/test/projection.rs#L673). Bumping the max size to 5 fixes it. Were you expecting others to fail as well?

jackh726:
That sounds about right in terms of breakage. Instead of changing it to 5, can you instead keep it at 4, but change the output?

tests/test/projection.rs
```diff
@@ -712,7 +712,7 @@ fn rust_analyzer_regression() {
                     PI: ParallelIterator
                 }
             }
-        } yields_all[SolverChoice::slg(4, None)] {
+        } yields_all[SolverChoice::slg(5, None)] {
             "substitution [], lifetime constraints []"
         }
```
jackh726, line 717: Yeah, keep size at 4, and change output to new expected.

PR #393, opened 2020-04-13
const generics support

Areredify:


Areredify:
@nikomatsakis I guess this is ready for review, I can't really think of something to make it nicer. As for the parser, I want #394 to land before I add proper name resolution to lowering.

chalk-ir/src/lib.rs
```diff
@@ -585,6 +593,81 @@ pub struct Fn<I: Interner> {
...
+    /// If this is a `ConstData::InferenceVar(d)`, returns `Some(d)` else `None`.
+    pub fn inference_var(&self, interner: &I) -> Option<InferenceVar> {
+        if let ConstData::InferenceVar(iv) = self.data(interner) {
+            Some(*iv)
+        } else {
+            None
+        }
+    }
+
+    /// True if this lifetime is a "bound" lifetime, and hence
```
detrumi, line 634: Should be bound const/constant

chalk-ir/src/visit.rs
```diff
@@ -180,6 +188,29 @@ where
         Self::Result::new()
     }
 
+    fn visit_free_var_const(
```
detrumi, line 191: At the moment, we could merge all these `visit_free_var_*` functions into a single `visit_free_var` function, as we don't handle them differently (and the same with `visit_bound_var_*` functions). Though maybe these might be useful in the future? Not sure.

chalk-solve/src/infer/instantiate.rs
```diff
@@ -95,6 +95,7 @@ impl<I: Interner> InferenceTable<I> {
                         lt.cast(interner)
                     }
                     ParameterKind::Ty(()) => placeholder_idx.to_ty(interner).cast(interner),
+                    ParameterKind::Const(_) => placeholder_idx.to_const(interner).cast(interner),
```
detrumi, line 98: Should probably match on `Const(())`, instead of any inner value?

chalk-ir/src/lib.rs
```diff
@@ -585,6 +593,81 @@ pub struct Fn<I: Interner> {
...
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasInterner)]
+pub struct ConcreteConst<I: Interner> {
+    pub interned: I::InternedConcreteConst,
+}
+
+impl<I: Interner> ConcreteConst<I> {
+    pub fn c_eq(&self, other: &ConcreteConst<I>, interner: &I) -> bool {
```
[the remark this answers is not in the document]
Areredify: My assumption was that `c_eq` wants to have access to the interner

chalk-solve/src/solve/slg.rs
```diff
@@ -564,6 +567,38 @@ impl<I: Interner> MayInvalidate<'_, I> {
         true
     }
 
+    // Returns true if the two consts could be unequal.
+    fn aggregate_consts(&mut self, new: &Const<I>, current: &Const<I>) -> bool {
```
jackh726, line 571: This is duplicated in `aggregate`? (Or it's similar)

tests/test/projection.rs
```diff
@@ -232,6 +232,49 @@ fn normalize_gat2() {
     }
 }
 
+#[test]
+fn normalize_gat_const() {
```
jackh726, line 236: This is a good start, but there's a few more tests that I would like to see before this lands. We can discuss on Zulip, but generally I'm interested that testing *specific* values works.

Areredify: Yeah, I was a little bit lazy with the tests, I just wanted to confirm I took the right path with the implementation?

chalk-ir/src/fold.rs
```diff
@@ -419,6 +473,61 @@ where
...
+        folder.fold_const(self, outer_binder)
+    }
+}
+
+impl<I, TI> SuperFold<I, TI> for Const<I>
+where
+    I: Interner,
+    TI: TargetInterner<I>,
+{
+    #[allow(unused_variables)]
```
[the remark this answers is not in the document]
nikomatsakis: The main reason to use `#[allow]` is on trait declarations where there are default definitions -- in that context, the parameter name shows up in the rustdoc.

chalk-ir/src/lib.rs
```diff
@@ -585,6 +585,49 @@ pub struct Fn<I: Interner> {
...
+    pub fn data(&self, interner: &I) -> &ConstData<I> {
+        I::const_data(interner, &self.interned)
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, HasInterner)]
+pub struct ConstData<I: Interner> {
+    // Should chalk store the type???
+    // I feel like we need it type to deal with 
+    // universally qualified stuff?
```
nikomatsakis: [rustc has it](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Const.html#structfield.ty), so it seems right in chalk too

chalk-ir/src/zip.rs
```diff
@@ -20,13 +20,14 @@ use std::sync::Arc;
...
+    /// Indicates that the two types `a` and `b` were found in matching spots.
     fn zip_tys(&mut self, a: &Ty<I>, b: &Ty<I>) -> Fallible<()>;
 
-    /// Indicates that the two lifetimes `a` and `b` were found in
-    /// matching spots, beneath `binders` levels of binders.
+    /// Indicates that the two lifetimes `a` and `b` were found in matching spots.
     fn zip_lifetimes(&mut self, a: &Lifetime<I>, b: &Lifetime<I>) -> Fallible<()>;
+    
+    /// Indicates that the two consts `a` and `b` were found in matching spots.
+    //fn zip_lifetimes(&mut self, a: &Const<I>, b: &Const<I>) -> Fallible<()>;
```
nikomatsakis: Nit: `zip_consts`

chalk-ir/src/zip.rs
```diff
@@ -397,16 +398,18 @@ impl<I: Interner> Zip<I> for GoalData<I> {
...
+impl<T: Zip<I>, L: Zip<I>, C: Debug, I: Interner> Zip<I> for ParameterKind<T, L, C> {
     fn zip_with<'i, Z: Zipper<'i, I>>(zipper: &mut Z, a: &Self, b: &Self) -> Fallible<()>
     where
         I: 'i,
     {
         match (a, b) {
             (ParameterKind::Ty(a), ParameterKind::Ty(b)) => Zip::zip_with(zipper, a, b),
             (ParameterKind::Lifetime(a), ParameterKind::Lifetime(b)) => Zip::zip_with(zipper, a, b),
-            (ParameterKind::Ty(_), _) | (ParameterKind::Lifetime(_), _) => {
+            _ => {
```
nikomatsakis: This is error-prone -- if we add a new variant, we get no error... I'd prefer to see

```rust
(ParameterKind::Ty(_), _) |
(ParameterKind::Const(_), _) |
(ParameterKind::Lifetime(_), _) => {
}
```

chalk-ir/src/visit.rs
nikomatsakis: Hmm, good point. I agree it's not clear if anything in the future would ever handle these differently. Perhaps not.

chalk-solve/src/infer/unify.rs
```diff
@@ -333,6 +333,87 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
+
+                self.unify_var_const(var, a)
+            }
+
+            (&ConstData::Placeholder(p1), &ConstData::Placeholder(p2)) => {
+                Zip::zip_with(self, &p1, &p2)
+            }
+
+            (&ConstData::Evaluated(ref ev1), &ConstData::Evaluated(ref ev2)) => {
+                if ev1.c_eq(ev2, interner) {
```
nikomatsakis, line 385: can we give this method a less abbreviated name, like `const_eq` or something?

chalk-ir/src/debug.rs
```diff
@@ -39,7 +39,7 @@ impl<I: Interner> Debug for Const<I> {
     }
 }
 
-impl<I: Interner> Debug for ConstEvaluated<I> {
+impl<I: Interner> Debug for ConcreteConst<I> {
```
nikomatsakis, line 42: Hmm, Rustc calls this [`ConstKind::Value`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.ConstKind.html#variant.Value), maybe we want to take that name

chalk-solve/src/infer.rs
```diff
@@ -118,73 +118,47 @@ impl<I: Interner> InferenceTable<I> {
...
-        match self.unify.probe_value(var) {
-            InferenceValue::Unbound(_) => None,
-            InferenceValue::Bound(ref val) => {
-                let ty = val.as_ref(interner).ty().unwrap().clone();
-                assert!(!ty.needs_shift(interner));
-                Some(ty)
-            }
-        }
+    pub(crate) fn normalize_ty_shallow(&mut self, interner: &I, leaf: &Ty<I>) -> Option<Ty<I>> {
+        self.probe_var(leaf.inference_var(interner)?)
```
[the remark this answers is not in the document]
nikomatsakis: That's ok, I think, cloning a `Ty` should be cheap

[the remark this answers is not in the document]
nikomatsakis: Though this assertion feels sort of gratuitous. I guess we'd need `assert_ty` helpers or something.

chalk-ir/src/zip.rs
Areredify: This is an outdated piece of code

chalk-ir/src/debug.rs
Areredify: I thought about naming it `ConstValue`, my main reasoning against it was that `ConstData` and `ConstValue` are too similar, `ConcreteConst` was one of the names eddyb proposed

chalk-solve/src/solve/slg.rs
Areredify: It's similar

chalk-solve/src/infer.rs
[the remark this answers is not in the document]
Areredify: It's not extra, it was in the previous version as well (`val.as_ref(interner).ty().unwrap().clone()`) and is required because we get `&ParameterData` from `Parameter::data`.

chalk-solve/src/infer/unify.rs
```diff
@@ -333,6 +333,87 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
 
+    fn unify_const_const<'a>(&mut self, a: &'a Const<I>, b: &'a Const<I>) -> Fallible<()> {
+        let interner = self.interner;
+
+        if let Some(n_a) = self.table.normalize_const_shallow(interner, a) {
+            return self.unify_const_const(&n_a, b);
+        } else if let Some(n_b) = self.table.normalize_const_shallow(interner, b) {
+            return self.unify_const_const(a, &n_b);
+        }
+
```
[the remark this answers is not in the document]
Areredify: I was puzzled by this construction as well, but forgot about it. I guess it would make sense if you can bind an inference var to an inference var (as opposed to unifying them), but that shouldn't happen as far as I know. @nikomatsakis what do you think?

Areredify:
We have two things we haven't decided on what to do with:
a) [this](https://github.com/rust-lang/chalk/pull/393/files#r410943055)
b) `ConcreteConst` vs `ConstValue`

chalk-solve/src/infer/unify.rs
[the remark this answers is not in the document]
nikomatsakis: You can do the `unwrap_or`. I'm not sure why I wrote it that way tbh. I guess I was just thinking "functionally" -- i.e,. of an inference rule like

```
Bound(V) = T'
Unify(T', T)
-----------------
Unify(V, T)
```

chalk-ir/src/debug.rs
nikomatsakis: OK. I don't care too much which name we adopt, we'll have to reconcile them eventually. I can see why `ConstData` and `ConstValue` seem close -- I wonder if this would feel better if we used `ConstKind` instead of `ConstData` (and similarly `TyKind` instead of `TyData` etc)...I was thinking of adopting that rename on the basis of "seems about equivalent and it's easier to port chalk than rustc".

nikomatsakis:
Work list:

* [ ] adding `ty` back  to definition of const
* [ ] adding tests [listed here](https://hackmd.io/Ah_J6nFQSbuUiVOsH5acLg)
* [ ] various nits 
* [ ] nit: https://github.com/rust-lang/chalk/pull/393/files#r410943055
* [ ] `ConcreteConst` vs `ConstValue`, naming choice

chalk-ir/src/debug.rs
Areredify: Well, now we have `ConstValue` and `ConstData` and `ConcreteConst` all at once :upside_down_face:

Areredify:
So, I implemented const types by adding `Ty` to `ParameterKinds::Const` and to `ConstData`, moving the old contents into `ConstValue`. This design has some drawbacks: mainly, it has a lot of duplication. 

a) `ParameterData` aka `ParameterKind<Ty, Lifetime, Const>` which has type of constant both in `ParameterKinds::Const` and in `Const`.
b) all bound vars, inference vars and placeholder vars store their types in both vars themselves and in binders and universes and inference tables.

b) is not really "fixable" I think;
in a) `ParameterData` should really be `ParameterKind::<Ty, Lifetime, ConstValue>` from the type standpoint, but that requires(?) to intern `ConstValue` and that is too many internings imo.

If you have ideas on how to design this better, I would like to hear them.

@jackh726 @detrumi I would appreciate if you took a second look at the pr if you have time

Areredify:
One more of my concerns is const type unification: it can push lifetime constraints if types are not equal-equal, which is not a big deal maybe, worst case scenario it's kicking the can down the road. I decided to defer dealing with it until subtyping is implemented

nikomatsakis:
@Areredify I'm kind of confused -- I would expect to have a type like

```rust
pub struct ConstData<I> {
    ty: Ty<I>,
    kind: ConstKind<I>,
}

pub enum ConstKind<I> {
    Variable, Bound, Value, etc
}
```

and then `type Parameter<I> = ParameterKind<Ty<I>, Lifetime<I>, Const<I>>`. 

Was there something with this approach that didn't work? (Modulo having to find a good name for `ConstKind`...) I feel like that's the approach you had early on in the PR history.

Areredify:
@nikomatsakis That's the current setup + `ParameterKind` is defined as 
```
pub enum ParameterKind<I, T, L, C> {
    Ty(T),
    Lifetime(L),
    Const {
        ty: Ty<I>,
        value: C
    } 
}
```

nikomatsakis:
@Areredify I'm not sure what you mean by "current", I guess, but it seems like the setup I laid out is not what you described. In particular, I don't understand why the `ParameterKind::Const` variant should be "different" than the others, in that it contains both a type *and* a constant. I think the `ty` field should live in the "constant", as I described.

Areredify:
@nikomatsakis Because we use `ParameterKind<T>`  in a lot of places (`ParameterKind<()>` in binders, `ParameterKind<EnaVariable>` in unification) we need it to carry type of the constant with it, that's why my current setup is:
```
pub enum ParameterKind<I, T, L, C> {
    Ty(T),
    Lifetime(L),
    Const { ty: Ty<I>, value: C },
}

pub struct ConstData<I> {
    ty: Ty<I>,
    kind: ConstKind<I>,
}

pub enum ConstKind<I> {
    Variable, Bound, Value, etc
}

type Parameter<I> = ParameterKind<Ty<I>, Lifetime<I>, Const<I>>;
```

nikomatsakis:
@Areredify Why do we need `ParameterKind<()>` to carry the type of the constant along with it?

Areredify:
To instantiate universal binders correctly, for example

Areredify:
rebased against the `GenericArg` pr

tests/test/constants.rs
```diff
@@ -0,0 +1,138 @@
...
+            "Ambiguous; no inference guidance"
+        }
+
+        goal {
+            forall<const N> {
+                S<N>: Trait
+            }
+        } yields {
+            "No possible solution"
+        }
```
nikomatsakis, line 61: can we test also

```rust
goal {
    S<3>: Trait
}
```

?

tests/test/projection.rs
```diff
@@ -457,6 +457,48 @@ fn normalize_gat2() {
...
+        }
+
+        goal {
+            forall<const N, T> {
+                exists<U> {
+                    Normalize(<StreamIterMut<T> as StreamingIterator<T>>::Item<N> -> U)
+                }
+            }
+        } yields {
+            "Unique; substitution [?0 := Span<!1_0, !1_1>], lifetime constraints []"
```
nikomatsakis, line 479: I guess we don't have any sigil for "constant placeholders"? Seems consistent, I suppose

tests/test/constants.rs
nikomatsakis: and maybe `S<5>: Trait` just for fun

chalk-solve/src/solve/slg/aggregate.rs
```diff
@@ -425,7 +434,53 @@ impl<I: Interner> AntiUnifier<'_, '_, I> {
         }
     }
 
-    fn new_variable(&mut self) -> Ty<I> {
+    fn aggregate_consts(&mut self, c1: &Const<I>, c2: &Const<I>) -> Const<I> {
+        let interner = self.interner;
+
+        // FIXME actually check that c1 and c2 have the same type
```
nikomatsakis, line 440: Is there a reason we can't invoke `let c_ty = self.aggregate_types(..)` or something like that? 

Well, I guess that's not quite what we want probably. We don't want to replace the types with inference variables. I guess that Rust's type system generally guarantees that the types here will be equivalent, so this is more of a "validation" check, right? (In which case, this isn't really a *FIXME* so much as "hmm this would be nice"?)

Areredify: Hmm, you are right, I'll change the wording then

Areredify:
@nikomatsakis locked and loaded

PR #394, opened 2020-04-14
Add scalars to `TypeName`

MarkMcCaskey:
Part of #368 ; spun off of #371


chalk-parse/src/ast.rs
```diff
@@ -290,3 +292,47 @@ pub enum Goal {
     // Additional kinds of goals:
     Leaf(LeafGoal),
 }
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+/// This allows users to add arbitrary `A :- B` clauses into the
+/// logic; it has no equivalent in Rust, but it's useful for testing.
```
nikomatsakis, line 298: This comment is surely wrong

chalk-parse/src/parser.lalrpop
```diff
@@ -217,6 +218,82 @@ Field: Field = {
...
+            "i128" => ScalarValue::Int(IntValue::I128(
+                digits.parse::<i128>().expect("failed to parse i128"),
+            )),
+            "isize" => ScalarValue::Int(IntValue::Isize(
+                digits.parse::<isize>().expect("failed to parse isize"),
+            )),
+            otherwise => unreachable!("Unknown numeric type suffix found: `{}`", otherwise),
+        }
+    } else {
+        // TODO: try various parsers/do whatever rustc does here
```
nikomatsakis, line 290: rustc uses inference that falls back to `i32`. I think we should either default to `i32` or else give an error and require people to be explicit.

chalk-parse/src/ast.rs
nikomatsakis: But I'm not convinced there's a role for this struct regardless -- I guess the intention was to give some way to "declare" scalar types? I think we should just build them in.

chalk-parse/src/ast.rs
```diff
@@ -24,6 +24,8 @@ pub enum Item {
     TraitDefn(TraitDefn),
     Impl(Impl),
     Clause(Clause),
+    Scalar(Scalar),
+    // TODO: tuple
```
nikomatsakis, line 28: I don't think we need these here -- there are no "items" corresponding to scales and tuples. Items are declarations like `struct Foo { .. }`

chalk-parse/src/ast.rs
```diff
@@ -290,3 +292,47 @@ pub enum Goal {
...
+    // TODO: review this
+    pub name: Option<Identifier>,
+    // Do we want to store the value too?
+    // storing the value is simple if we're already parsing. TODO: is this useful?
+    pub value: ScalarValue,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+/// The inner value and type of the scalar value.
+pub enum ScalarValue {
```
nikomatsakis, line 310: I would leave out scalar values for now -- we won't need these until the const generics stuff lands, and I think it'd be better to add in that PR

MarkMcCaskey:
Thank you for the very helpful review!

Here's the status of the PR:
- `chalk_solve::clauses::match_type_name` does a no-op for scalars and tuples right now, this probably isn't what we want -- I didn't get around to adding tests that would cause this to fail yet but I imagine it won't be hard to do
- scalar type name and tuple parsing now works: it may be a bit sub-optimal in places because I expected `(A, B,)` to not parse, but I just tested it now and it works!  Which means that my implementation, which special cases the single tuple case of `(A,)`, may be overly complex.  Also I used right-recursion for something when some of the existing parsing code was using left-recursion which is a red flag to me that I may have made a mistake.  It's been a bit since I did BNF-style parsing and I don't remember the details -- I'll look into this tomorrow.
- I updated all the tests that were using names that collide with the new scalar values to use different names.  The reason I chose different names instead of removing the declarations was to future-proof in case traits are automatically implemented for the scalar types and that interferes with the functioning of the tests.  That said, I did notice that `#[auto]` on a trait did _not_ apply to the scalar values -- another thing that should be tested!
- It's a lexing/regex error if the tests use something like `struct i32 { }` now which is probably not what we want.  Ideally there'd be a more clear error message there.

So from my perspective there's 3 things left to do:
- [ ] Address feedback and fix anything I've missed here
- [ ] Add more interesting tests and improve the internals to make that work
- [ ] Resolve the merge conflict implementing `Visit` for scalars and tuples

I'll try to get those done tomorrow if I'm able to!

chalk-parse/src/ast.rs
```diff
@@ -175,6 +175,47 @@ pub enum Ty {
         lifetime_names: Vec<Identifier>,
         ty: Box<Ty>,
     },
+    Tuple {
+        arity: usize,
```
nikomatsakis, line 179: This should be 

```rust
Tuple {
    types: Vec<Box<Ty>>
}
```

chalk-integration/src/lowering.rs
```diff
@@ -1074,6 +1074,17 @@ impl LowerTy for Ty {
                 };
                 Ok(chalk_ir::TyData::Function(function).intern(interner))
             }
+            Ty::Tuple { arity } => Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
```
nikomatsakis, line 1077: If this were `types` (and not `arity`), then you would have...

chalk-integration/src/lowering.rs
```diff
@@ -1074,6 +1074,17 @@ impl LowerTy for Ty {
...
+                name: chalk_ir::TypeName::Tuple(arity),
```
nikomatsakis, line 1078: `s/arity/types.len()/`

chalk-integration/src/lowering.rs
```diff
@@ -1074,6 +1074,17 @@ impl LowerTy for Ty {
...
+                substitution: chalk_ir::Substitution::empty(interner),
```
nikomatsakis, line 1079: `chalk_ir::Substition::from_fallible(interner, types.iter().map(...)) // as in the code above`

chalk-parse/src/parser.lalrpop
```diff
@@ -188,7 +189,33 @@ TyWithoutFor: Ty = {
...
+   "i128" => ScalarType::Int(IntTy::I128),
+   "isize" => ScalarType::Int(IntTy::Isize),
+   "f32" => ScalarType::Float(FloatTy::F32),
+   "f64" => ScalarType::Float(FloatTy::F64),
+   "bool" => ScalarType::Bool,
+   "char" => ScalarType::Char,
+};
+
+TupleOrParensInner: Ty = {
+    <l:Comma2<Ty>> => Ty::Tuple { arity: l.len() },
```
nikomatsakis, line 215: Mm yeah it'd nice to not have `Comma2`.

Try this:

```
TupleOrParensInner: Ty = {
    <Ty>,
    <Ty> "," => Ty::Tuple { types: vec[<>] },
    <a:Ty> "," <types:Comma<Ty>>  => {
        let mut types  = types;
        types.insert(0, a);
        Ty::Tuple { types }
    }
```

(we can always extend `Comma` to accept trailing commas, too, but that's separate)

nikomatsakis: not mega efficient but who cares

tests/lowering/mod.rs
```diff
@@ -229,9 +225,9 @@ fn check_parameter_kinds() {
     lowering_error! {
         program {
             struct Foo<'a> { }
-            struct i32 { }
+            struct Myi32 { }
             trait Bar { }
-            impl Bar for Foo<i32> { }
+            impl Bar for Foo<Myi32> { }
```
nikomatsakis, line 230: could just use `i32` here, too, but this is fine

tests/lowering/mod.rs
```diff
@@ -425,3 +421,58 @@ fn fundamental_multiple_type_parameters() {
...
+            impl Foo for char { }
+        }
+    }
+
+    lowering_error! {
+        program {
+            struct i32 { }
+        }
+
+        // TODO: improve this error message
```
nikomatsakis, line 473: seems fine, our parse errors stink

chalk-integration/src/lowering.rs
```diff
@@ -1074,6 +1074,21 @@ impl LowerTy for Ty {
...
+                substitution: chalk_ir::Substitution::from_fallible(
+                    interner,
+                    types.iter().map(|t| Ok(t.lower(env)?)),
+                )?,
+            })
+            .intern(interner)),
+
+            Ty::Scalar { ty } => Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
+                name: chalk_ir::TypeName::Scalar(ast_scalar_to_chalk_scalar(ty)),
+                // substitution: chalk_ir::Substitution::from_fallible(interner, ty.lower(env)?)?,
```
nikomatsakis, line 1088: Nit: remove this comment
```suggestion
```

tests/lowering/mod.rs
```diff
@@ -425,3 +421,138 @@ fn fundamental_multiple_type_parameters() {
...
+    }
+}
+
+#[test]
+fn tuple_trait_impl() {
+    let db = ChalkDatabase::with(
+        "
+trait Foo { }
+struct S1 { }
+impl Foo for (S1, S1) { }
```
nikomatsakis, line 485: Are these tests trying to test the solver? If so, they should move to a file like tests/test/tuples.rs or something like that, and they use the `test!` macro like you see here:

https://github.com/rust-lang/chalk/blob/28cef6ff403d403e6ad2f3d27d944e9ffac1bce8/tests/test/auto_traits.rs#L5-L54

tests/lowering/mod.rs
```diff
@@ -425,3 +421,138 @@ fn fundamental_multiple_type_parameters() {
...
+    db.with_program(|_| {
+        assert_eq!(
+            format!("{:?}", goal),
+            "Implemented(3<Int(I32), Int(I32), 1<Int(I32)>>: Foo)"
+        );
+    });
+}
+
+#[test]
+fn scalar_trait_impl() {
```
nikomatsakis, line 510: similarly, move this to a file like tests/test/scalars.rs

tests/test/scalars.rs
```diff
@@ -0,0 +1,87 @@
...
+            impl Foo for u64 { }
+            impl Foo for u128 { }
+            impl Foo for usize { }
+            impl Foo for f32 { }
+            impl Foo for f64 { }
+            impl Foo for bool { }
+            impl Foo for char { }
+        }
+
+        goal { i8: Foo } yields { "Unique" }
```
nikomatsakis, line 70: Just an old rule of thumb, but maybe good to test something that's *not true* as well... e.g., `goal { i8: Bar }` (if you added a trait `Bar` with impls only for unsigned ints or something :)

MarkMcCaskey: Good catch -- thanks!

chalk-parse/src/parser.lalrpop
```diff
@@ -204,7 +204,6 @@ TyWithoutFor: Ty = {
     },
     <n:Id> "<" <a:Comma<Parameter>> ">" => Ty::Apply { name: n, args: a },
     <p:ProjectionTy> => Ty::Projection { proj: p },
-    <a:AliasTy> => Ty::Alias { alias: a },
```
nikomatsakis: why did you remove this?

MarkMcCaskey: I accidentally left it in from the merge above!  It was removed in #324 [here](https://github.com/rust-lang/chalk/pull/324/files#diff-6fa8bab6c97bead41db277caab4db7a2L193) and is gone on master now

chalk-integration/src/lowering.rs
```diff
@@ -21,6 +21,7 @@ use crate::{Identifier as Ident, RawId, TypeKind, TypeSort};
 type StructIds = BTreeMap<Ident, chalk_ir::StructId<ChalkIr>>;
 type TraitIds = BTreeMap<Ident, chalk_ir::TraitId<ChalkIr>>;
 type OpaqueTyIds = BTreeMap<Ident, chalk_ir::OpaqueTyId<ChalkIr>>;
+type ScalarIds = BTreeMap<chalk_ir::Scalar, RawId>;
```
nikomatsakis, line 24: I don't think these ids should be needed

chalk-solve/src/clauses.rs
```diff
@@ -416,6 +416,12 @@ fn match_ty<I: Interner>(
 ) -> Result<(), Floundered> {
     let interner = builder.interner();
     Ok(match ty.data(interner) {
+        TyData::Apply(ApplicationTy {
+            name: TypeName::Scalar(_),
+            ..
+        }) => {
+            builder.push_fact(WellFormed::Ty(ty.clone()));
```
nikomatsakis, line 423: I think this logic should move into `match_type_name`

chalk-solve/src/clauses.rs
```diff
@@ -454,6 +460,8 @@ fn match_type_name<I: Interner>(builder: &mut ClauseBuilder<'_, I>, name: TypeNa
             .db
             .associated_ty_data(type_id)
             .to_program_clauses(builder),
+        TypeName::Scalar(_) => (),
```
nikomatsakis, line 463: i.e., right here

chalk-solve/src/clauses.rs
```diff
@@ -454,6 +460,8 @@ fn match_type_name<I: Interner>(builder: &mut ClauseBuilder<'_, I>, name: TypeNa
...
+        TypeName::Tuple(_) => (),
```
nikomatsakis, line 464: I think this is right, though it may change as we tweak our def'n of implied bounds

MarkMcCaskey:
Thanks!  I moved the logic but had to change the type signature of `match_type_name` to take an `ApplicationTy` and an `Interner` to be able to get a `Ty` to be able to add the fact.  Let me know if there's a better way to do this!  I wasn't able to find anything much better by looking through the docs.

A somewhat reasonable alternative is to pass `Ty` and an `Interner` instead: we can avoid the cost of interning the `ApplicationTy` again at the cost of having to handle a failure case and do the branch on `ty.data` again.

nikomatsakis:
It seems fine as is to me, thanks!

PR #400, opened 2020-04-16
Improve opaque type debug output

detrumi:
Properly show things like `AliasEq(T = !T)` and `AliasEq(T = Ty) :- Reveal` in debug output, instead of `AliasEq(impl Ty = Ty)` and `AliasEq(impl Ty = Ty) :- Reveal`


nikomatsakis:
yay! 🎉

PR #401, opened 2020-04-16
Semantic equality to syntactic equality

Areredify:


Areredify:
coinductive test is failing, I don't know enough about it pls help

chalk-solve/src/clauses/builder.rs
```diff
@@ -76,18 +77,17 @@ impl<'me, I: Interner> ClauseBuilder<'me, I> {
...
-            );
-        }
+            ProgramClauseData::ForAll(Binders::new(
+                ParameterKinds::from(interner, self.binders.clone()),
+                clause,
+            ))
+            .intern(interner)
+        };
+
+        self.clauses.push(syn_eq_lower(interner, &clause));
```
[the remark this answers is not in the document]
Areredify: Yeah, I thought about making `syn_eq_lower` to return a `SynLowered<Fold::Result>`

Areredify:
Woah you can push to my branch

chalk-solve/src/clauses/builder.rs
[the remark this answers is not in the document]
nikomatsakis: Not sure I follow -- what is `SynLowered`, just a newtype?

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+            _ => Ok(ty.super_fold_with(self, outer_binder)?),
+        }
+    }
+
+    fn fold_program_clause(
+        &mut self,
+        clause: &ProgramClause<I>,
+        outer_binder: DebruijnIndex,
+    ) -> Fallible<ProgramClause<I>> {
+        let interner = self.interner;
```
nikomatsakis, line 67: I think we should either 

* save the old values of `new_params`/`new_goals`, and restore before we return
* or assert that they are empty (if that is known to be true)

I don't see why they would be known to be empty, though, so I think we should save the values of the vector. The alternative would be to rewrite the code to not touch pre-existing things and just save the lengths.

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        let ((binders, implication), in_binders) = match clause.data(interner) {
+            ProgramClauseData::ForAll(for_all) => (for_all.clone().into(), true),
+            // introduce a dummy binder and shift implication through it
+            ProgramClauseData::Implies(implication) => (
```
nikomatsakis, line 72: I keep meaning to factor away `ProgramClauseData::Implies`, and just *always* have a binder.

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+                (
+                    ParameterKinds::new(interner),
+                    implication.shifted_in(interner),
+                ),
+                false,
+            ),
+        };
+        let mut binders: Vec<_> = binders.as_slice(interner).clone().into();
+
+        let outer_binder = outer_binder.shifted_in();
```
nikomatsakis, line 82: Comment: 

// Adjust the outer binder to account for the binder in the program clause

bjorn3:
@Areredify There is a checkbox "Allow edits and access to secrets by maintainers" both when opening a PR and when you opened it at the bottom of the sidebar. It is checked by default.

chalk-solve/src/clauses/syntactic_eq.rs
nikomatsakis: Probably we need to save `binders_len` too

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        self.binders_len = binders.len();
```
nikomatsakis, line 84: Comment here, referencing the function comment:

```
// First lower the "consequence" -- in our example that is 
//
// ```
// Implemented(<X as Iterator>::Item: Debug)
// ```
//
// then save out the `new_params` and `new_goals` vectors to store any new variables created as a result. In this case, that would be the `Y` parameter and `AliasEq(<X as Iterator>::Item, Y)` goals. 
//
// Note that these new parameters will have indices that come after the existing parameters, so any references to existing parameters like `X` in the "conditions" are still valid even if we insert new parameters.
```

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+        let consequence = implication.consequence.fold_with(self, outer_binder)?;
+        // Immediately move `new_params` out of of the folder struct so it's safe
+        // to call `.fold_with` again
```
nikomatsakis, line 87: (then I think you can remove this comment, it's explained more usefully above)

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+        let new_params = replace(&mut self.new_params, vec![]);
+        let new_goals = replace(&mut self.new_goals, vec![]);
+
+        let mut conditions = implication.conditions.fold_with(self, outer_binder)?;
```
nikomatsakis, line 91: Comment:

// Now fold the conditions (in our example, `Implemented(X: Debug)`. The resulting list might be expanded to include new `AliasEq` goals.

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+        if new_params.is_empty() && !in_binders {
+            // shift the clause out since we didn't use the dummy binder
```
nikomatsakis, line 93: imo not worth this special case :)

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+        Ok(ProgramClauseData::ForAll(Binders::new(
+            ParameterKinds::from(interner, binders),
+            ProgramClauseImplication {
+                consequence,
+                conditions,
+                priority: implication.priority,
+            },
+        ))
+        .intern(interner))
+    }
```
nikomatsakis, line 121: (we would restore the saved fields here)

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+                self.binders_len += 1;
+                ty.super_fold_with(self, outer_binder)?;
+                Ok(new_ty)
+            }
+            TyData::Function(_) => Ok(ty.clone()),
+            _ => Ok(ty.super_fold_with(self, outer_binder)?),
+        }
+    }
+
+    fn fold_program_clause(
```
nikomatsakis, line 62: I would include a comment here that gives the overall strategy, which the function *body* can reference. Of course this was outlined in great detail in the module comment, so something like this would work. In these cases, I like to give very concrete example so that, in the code comments, I can refer to very specific things. e.g.,

```
/// Convert a program clause to rem
///
/// Consider this (nonsense) example:
///
/// ```notrust
/// forall<X> { 
///     Implemented(<X as Iterator>::Item>: Debug) :- 
///         Implemented(X: Debug)
/// }
/// ```
///
/// we would lower this into:
///
/// ```notrust
/// forall<X, Y> { 
///     Implemented(Y: Debug) :- 
///         AliasEq(<X as Iterator>::Item>, Y),
///         Implemented(X: Debug)
/// }
/// ```
```

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+            ProgramClauseImplication {
+                consequence,
+                conditions,
+                priority: implication.priority,
+            },
+        ))
+        .intern(interner))
+    }
+
+    fn fold_goal(&mut self, goal: &Goal<I>, outer_binder: DebruijnIndex) -> Fallible<Goal<I>> {
```
nikomatsakis, line 123: add comments with a concrete example, siimlar to above

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+        assert!(self.new_params.is_empty(), true);
+
+        let interner = self.interner;
+        match goal.data(interner) {
+            GoalData::DomainGoal(_) | GoalData::EqGoal(_) => (),
+            _ => return goal.super_fold_with(self, outer_binder),
+        };
+
+        self.binders_len = 0;
+        // shifted in because we introduce a new binder
```
nikomatsakis, line 133: then here we can talk specifically about what binder you are talking about. I presume it's the new `exists<..>` binder we plan to introduce

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn fold_goal(&mut self, goal: &Goal<I>, outer_binder: DebruijnIndex) -> Fallible<Goal<I>> {
+        assert!(self.new_params.is_empty(), true);
+
+        let interner = self.interner;
+        match goal.data(interner) {
+            GoalData::DomainGoal(_) | GoalData::EqGoal(_) => (),
+            _ => return goal.super_fold_with(self, outer_binder),
+        };
+
+        self.binders_len = 0;
```
nikomatsakis, line 132: comment: 

// Set `binders_len` to zero as in the `exists<..>` binder we will create, there are no existing variables.

chalk-solve/src/clauses/syntactic_eq.rs
nikomatsakis, line 84: Comment: 

Set binders_len to `binders.len` since new parameters will be added into the existing `forall<...>` binder on the program clause.

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+    };
+
+    clause
+        .fold_with(&mut folder, DebruijnIndex::INNERMOST)
+        .unwrap()
+}
+
+struct SynEqFolder<'i, I: Interner> {
+    interner: &'i I,
+    new_params: Vec<ParameterKind<()>>,
```
nikomatsakis, line 28: comments on these fields -- what do they store?

My suggestions:

Store the kinds of any new parameters that are introduced during folding. These parameters will either be added to an enclosing `exists` binder (when lowering a goal) or an enclosing `forall` binder (when lowering a program clause).

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+    new_goals: Vec<Goal<I>>,
```
nikomatsakis, line 29: For each new parameter `X`, introduces an `AliasEq` goal that defines it, like `AliasEq(<T as Iterator>::Item, X)`

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+    binders_len: usize,
```
nikomatsakis, line 30: Stores the number of variables in the binder that we are adding parameters into thus far. Incremented for each new variable we add.

chalk-solve/src/clauses/syntactic_eq.rs
```diff
@@ -0,0 +1,165 @@
...
+use chalk_ir::{
+    cast::Cast,
+    fold::{shift::Shift, Fold, Folder, SuperFold},
+    interner::Interner,
+    AliasEq, AliasTy, Binders, BoundVar, DebruijnIndex, Goal, GoalData, Goals, ParameterKind,
+    ParameterKinds, ProgramClause, ProgramClauseData, ProgramClauseImplication, QuantifierKind, Ty,
+    TyData,
+};
+
+pub fn syn_eq_lower<I: Interner, T: Fold<I>>(interner: &I, clause: &T) -> <T as Fold<I>>::Result {
```
nikomatsakis, line 13: Needs a comment. Also, I wonder if it's worth adding some *other* trait since you only really want to fold program clauses and goals here..? Nah, probably not, but we should add a comment about it.

chalk-solve/src/clauses/syntactic_eq.rs
Areredify: We know them to be empty because only domain goals and program clauses contain types (and EqGoal, but it is unique), and when folding domain goals we are guaranteed to not encounter another goal or clause.  If we restrict input types of `syn_lower` to only be `ProgramClause` and `Goal`, then the lowering should be correct.

chalk-solve/src/clauses/syntactic_eq.rs
Areredify: I think making it into `syn_lower_clause` and `syn_lower_goal` should be enough

nikomatsakis: Yeah, two separate functions would be fine

chalk-solve/src/clauses/syntactic_eq.rs
nikomatsakis: ok, then add assertions -- I guess it's true that we only have to worry about *domain goal*, and it cannot contain clauses. Right. I was thinking about how goals can contain clauses in general (and vice versa, of course).

jackh726:
Closing in favor of #589

PR #402, opened 2020-04-17
[WIP] Opaque type generics

detrumi:
Opaque types should accept generic parameters, probably something like this:
```
program {
    trait Iterator { type Item; }

    struct Vec<T> { }
    struct u32 { }
    impl<T> Iterator for Vec<T> {
        type Item = T;
    }

    opaque type Foo<X>: Iterator<Item = X> = Vec<X>;
}

goal {
    Foo<u32>: Iterator<Item = u32>
} yields {
    "Unique; substitution []"
}
```


detrumi:
Closing in favor of #464

PR #403, opened 2020-04-17
Make it possible to cache the result of env elaboration

flodiebold:
`program_clauses_for_env` can take a huge amount of time -- rust-analyzer is unusably slow on the Chalk codebase itself, spending most of that time in env elaboration because one `I: Interner` results in ~300 clauses, which gets repeated tens of times per query.

This adds a method to `RustIrDatabase` that should basically just be a pass-through for `program_clauses_for_env`, but which makes it possible to cache the results in Salsa.

The effect of this is a ~20% speedup of analysis of rust-analyzer on itself, and turning analysis of the Chalk codebase from seemingly forever to ~1 minute.

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -127,8 +127,13 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
     let mut vec = vec![];
     vec.extend(db.custom_clauses());
     program_clauses_that_could_match(db, environment, goal, &mut vec)?;
-    program_clauses_for_env(db, environment, &mut vec);
     vec.retain(|c| c.could_match(interner, goal));
+    vec.extend(
+        db.program_clauses_for_env(environment)
+            .iter(interner)
+            .filter(|c| (*c).could_match(interner, goal))
+            .cloned(),
+    );
 
     debug!("vec = {:#?}", vec);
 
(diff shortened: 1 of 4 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -127,8 +127,13 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
     let mut vec = vec![];
     vec.extend(db.custom_clauses());
     program_clauses_that_could_match(db, environment, goal, &mut vec)?;
-    program_clauses_for_env(db, environment, &mut vec);
     vec.retain(|c| c.could_match(interner, goal));
+    vec.extend(
+        db.program_clauses_for_env(environment)
```
jackh726, line 132: Why not just move this above and keep the `retain` after the `extend` here? I guess it theoretically uses more memory, but it's cleaner.

bjorn3: It would also be a bit slower, as you would have to move elements back after you inserted them.

flodiebold: Yeah, I just didn't want to insert hundreds of elements just to remove them immediately afterwards, considering this actually gets called a lot.

jackh726: Theoretically we could also use `.chain().filter()` for everything. But I'm gonna just add a FIXME and land this.

PR #407, opened 2020-04-19
Add integration test for unwind safety

### LINKED ISSUE #260: Make chalk_solve::Solver UnwindSafe

marmeladema:
This currently does _not_ work because when the panic triggers, the solver lock is poisoned and i am not aware of a way to force un-poisoning it.

One approach that i've tried and that work, is to lock the solver outside of `catch_unwind` but this requires to use `AssertUnwindSafe` to use the lock guard inside `catch_unwind` and that kind of defeats the whole purpose of that test.

Fixes #260

### DIFF
--- a/tests/integration/panic.rs
+++ b/tests/integration/panic.rs
@@ -1,0 +1,160 @@
+use chalk_integration::db::ChalkDatabase;
+use chalk_integration::query::LoweringDatabase;
+use chalk_ir::interner::ChalkIr;
+use chalk_ir::AssocTypeId;
+use chalk_ir::Goal;
+use chalk_ir::ImplId;
+use chalk_ir::InEnvironment;
+use chalk_ir::OpaqueTyId;
+use chalk_ir::Parameter;
+use chalk_ir::ProgramClause;
+use chalk_ir::StructId;
+use chalk_ir::TraitId;
+use chalk_ir::TypeName;
+use chalk_ir::UCanonical;
+use chalk_rust_ir::AssociatedTyDatum;
+use chalk_rust_ir::AssociatedTyValue;
+use chalk_rust_ir::AssociatedTyValueId;
+use chalk_rust_ir::ImplDatum;
+use chalk_rust_ir::OpaqueTyDatum;
+use chalk_rust_ir::StructDatum;
+use chalk_rust_ir::TraitDatum;
+use chalk_rust_ir::WellKnownTrait;
+use chalk_solve::RustIrDatabase;
+use chalk_solve::Solution;
+use chalk_solve::SolverChoice;
+use std::sync::Arc;
+
+#[derive(Debug, Default)]
+struct MockDatabase {
+    chalk_db: ChalkDatabase,
+    panic: bool,
+}
+
+impl MockDatabase {
+    pub fn with(program_text: &str, solver_choice: SolverChoice) -> Self {
+        Self {
+            chalk_db: ChalkDatabase::with(program_text, solver_choice),
+            panic: false,
+        }
+    }
+
+    pub fn solve(
+        &self,
+        goal: &UCanonical<InEnvironment<Goal<ChalkIr>>>,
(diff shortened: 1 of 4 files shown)


tests/integration/panic.rs
```diff
@@ -0,0 +1,160 @@
...
+use chalk_rust_ir::TraitDatum;
+use chalk_rust_ir::WellKnownTrait;
+use chalk_solve::RustIrDatabase;
+use chalk_solve::Solution;
+use chalk_solve::SolverChoice;
+use std::sync::Arc;
+
+#[derive(Debug, Default)]
+struct MockDatabase {
+    chalk_db: ChalkDatabase,
```
jackh726, line 30: I don't think we actually want to wrap an actual database.

tests/integration/panic.rs
```diff
@@ -0,0 +1,160 @@
...
+        &self,
+        goal: &UCanonical<InEnvironment<Goal<ChalkIr>>>,
+    ) -> Option<Solution<ChalkIr>> {
+        let solver = self.chalk_db.solver();
+        let solution = solver.lock().unwrap().solve(self, goal);
+        solution
+    }
+}
+
+impl RustIrDatabase<ChalkIr> for MockDatabase {
```
jackh726, line 52: For these methods, I was thinking just hardcode what we want to return for `trait Foo` and `struct Bar`

tests/integration/panic.rs
```diff
@@ -0,0 +1,160 @@
...
+#[test]
+fn unwind_safety() {
+    use self::MockDatabase;
+    use chalk_integration::lowering::LowerGoal;
+    use chalk_integration::query::LoweringDatabase;
+    use chalk_solve::ext::GoalExt;
+    use std::panic;
+
+    // lower program
+    let mut db = lower_program_with_db! {
```
jackh726, line 130: So, we won't actually "lower" anything, it'll just be hardcoded.

tests/integration/panic.rs
```diff
@@ -0,0 +1,160 @@
...
+            trait Bar { }
+            impl Bar for Foo { }
+        }
+        database MockDatabase
+    };
+
+    let program = db.chalk_db.checked_program().unwrap();
+
+    // lower goal
+    let goal = lower_goal! {
```
jackh726, line 142: Similarly here. We aren't going to "lower" this, we'll just manually create the `UCanonicalGoalInEnvironment` for `Foo: Bar`

tests/integration/panic.rs
```diff
@@ -0,0 +1,160 @@
...
+    // solve goal but this will panic
+    db.panic = true;
+    let result = panic::catch_unwind(|| {
+        db.solve(&peeled_goal);
+    });
+    assert!(result.is_err() == true);
+
+    // solve again but without panicking this time
+    db.panic = false;
+    assert!(db.solve(&peeled_goal).is_some());
```
jackh726, line 159: This is exactly what I was thinking :)

tests/test_util.rs
```diff
@@ -1,15 +1,35 @@
 #![allow(unused_macros)]
 
-macro_rules! lowering_success {
-    (program $program:tt) => {
+macro_rules! lower_program_with_db {
```
jackh726, line 3: I don't think we want to change/use anything here.

jackh726:
Closing in favor of https://github.com/rust-lang/chalk/pull/462

Thanks for the initial work here though :)

PR #409, opened 2020-04-22
Re: Introduce Tracing into Chalk

nolanderc:
Replaces the old logging system that printed directly to stderr with macros from the [`tracing`](https://github.com/tokio-rs/tracing/tree/master/tracing) crate.


chalk-engine/src/logic.rs
```diff
@@ -112,7 +112,7 @@ impl<C: Context> Forest<C> {
         mut test: impl FnMut(&C::InferenceNormalizedSubst) -> bool,
     ) -> bool {
         if let Some(answer) = self.tables[table].answer(answer) {
-            info!("answer cached = {:?}", answer);
+            info!(?answer, "answer cached");
```
nikomatsakis, line 115: Woah! what is happening here. I guess this is a different for how tracing works?

nolanderc: Yup, in tracing the logging macros generate `Event`s which may contain a number of fields. The `?` sigil is a shorthand for creating a new field using the `Debug` implementation of `answer`.

nolanderc:
`?foo` is a shorthand for creating a new field on the logging `Event` using the `Debug` implementation for `foo`. In contrast to `log`, `tracing` prefers to keep parameters separate from the message string. I'll see what I can do regarding the chalk book.

I'm a new contributor to the Rust project, so I'm not familiar with what the policy around adding new dependencies is. But for what it is worth, `tracing` has a feature flag that forces it to _only_ emit `log` records: https://tracing.rs/tracing/#emitting-log-records. Would that be sufficient to resolve your concern?

detrumi:
> `tracing` has a feature flag that forces it to _only_ emit `log` records: https://tracing.rs/tracing/#emitting-log-records. Would that be sufficient to resolve your concern?

I think the problem is that it adds a new dependency in `Cargo.toml`, which increases build times for rustc. A cargo feature allows you to opt-out of that dependency.

nikomatsakis:
Well, there are two concerns:

* build times
* licensing and general "exposure"

We have a "accept-list" of dependencies it would have to be added to, IIRC, and we'd also want to check the license.

As it happens I'd like to port rustc over to using tracing, but that's a job in and of itself.

jackh726:
Ok, @nolanderc think you can rebase this? And maybe add docs in the chalk for how to view logs.

@nikomatsakis Now that chalk-ir doesn't depend on chalk-engine (the other way around), then tracing won't be a dependency for `librustc_middle`, which is what the whitelist is for. So good on that front.

Other than that, based on discussion on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Tracing.20in.20rustc), I think this is fine :)

nikomatsakis:
@jackh726 

> Now that chalk-ir doesn't depend on chalk-engine (the other way around), then tracing won't be a dependency for librustc_middle, which is what the whitelist is for. So good on that front.

I don't quite understand that. Can you spell it out a bit more?

jackh726:
@nikomatsakis So, I realized that the debug macros are in chalk-ir now, so maybe that statement was confusing. But there are no calls that use them in `chalk-ir`. So, if we switch to using `tracing`, we can just make `chalk-engine` and `chalk-solve` have direct dependencies on it separately.

The whitelist in rustc is only for crates that librustc_middle depends on, directly or transitively. librustc_middle only depends on `chalk-ir` and `chalk-derive`, so this should be it won't depend on `tracing`.

nikomatsakis:
> The whitelist in rustc is only for crates that librustc_middle depends on, directly or transitively. 

What? This sounds like a bug in rustc.

jackh726:
See https://github.com/rust-lang/rust/blob/ad4bc3323b9299d867697e9653dcea1b5e1ad283/src/tools/tidy/src/deps.rs#L59

jackh726:
Closing in favor of https://github.com/rust-lang/chalk/pull/525

Sorry we didn't get this pulled in @nolanderc

PR #410, opened 2020-04-23
introduce new high-level summary using mermaid

nikomatsakis:
Removed some old material, moved the rest to walkthrough, it's probably debated and somewhat overlaps.

r? @jackh726


nikomatsakis:
Some screen shots:

image

image

book/src/what_is_chalk/walkthrough.md
```diff
@@ -0,0 +1,218 @@
...
+*See also: [The SLG Solver][slg]*
+
+## Crates
+
+Chalk's functionality is broken up into the following crates:
+- [**chalk_engine**][chalk_engine]: Defines the core [SLG solver][slg].
+- [**chalk_rust_ir**][chalk_rust_ir], containing the "HIR-like" form of the AST
+- [**chalk_ir**][chalk_ir]: Defines chalk's internal representation of
+  types, lifetimes, and goals.
+- [**chalk_solve**][chalk_solve]: Combines `chalk_ir` and `chalk_engine`,
```
jackh726, line 139: We should open a followup issue to make sure we document where the recursive solver fits into all of this. I don't think it needs to go in this PR, since we are considering splitting it out of `chalk-solve` anyways.

PR #411, opened 2020-04-24
Use FxHashMap/FxHashSet and add well-formed clause for tuples

jackh726:


bjorn3:
Using `FxHashMap` and `FxHashSet` makes it harder to detect cases where the output depends on the iteration order, as they don't use a random number in their hashes, unlike `HashMap` and `HashSet`, which means that simply rerunning won't show the problem.

jackh726:
Unfortunately, we can't build through rustc without using them. But I'm not sure if we really prefer to randomize order anyways.

bjorn3:
Maybe do it behind a feature flag that is enabled on CI, but disabled for rustc. That way the rustc lint won't find it, but CI will still catch being dependent on iteration order.

jackh726:
We already use FxHashMap elsewhere. And I wouldn't want to rely on there *maybe* being a different order. It's not a bad idea to have tests where we which around iteration order. But I would rather that be explicit.

flodiebold:
I agree, I originally intended to change this to `FxHashSet` anyway but forgot.

PR #413, opened 2020-04-25
Handle functions, scalars and tuples for the builtin `Sized` trait

marmeladema:
I tried to follow @Areredify advises on zulip and tried to handle scalars and tuples for the builtin trait `Sized`. I hope its not too wrong.

I am unsure if goal
```Rust
(dyn Copy, u8): Sized
```
should really have a solution?

### DIFF
--- a/chalk-solve/src/clauses/builtin_traits.rs
+++ b/chalk-solve/src/clauses/builtin_traits.rs
@@ -1,6 +1,6 @@
 use super::builder::ClauseBuilder;
 use crate::{Interner, RustIrDatabase, TraitRef, WellKnownTrait};
-use chalk_ir::TyData;
+use chalk_ir::{Substitution, Ty, TyData};
 
 mod clone;
 mod copy;
@@ -29,4 +29,28 @@ pub fn add_builtin_program_clauses<I: Interner>(
         // Drop impls are provided explicitly
         WellKnownTrait::DropTrait => (),
     }
+}
+
+/// Given a trait ref `T0: Trait` and a list of types `U0..Un`, pushes a clause of the form
+/// `Implemented(T0: Trait) :- Implemented(U0: Trait) .. Implemented(Un: Trait)`
+pub fn needs_impl_for_tys<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    trait_ref: &TraitRef<I>,
+    tys: impl Iterator<Item = Ty<I>>,
+) {
+    // The trait must take one parameter (a type)
+    debug_assert_eq!(
+        db.trait_datum(trait_ref.trait_id)
+            .binders
+            .len(db.interner()),
+        1,
+    );
+    builder.push_clause(
+        trait_ref.clone(),
+        tys.map(|ty| TraitRef {
+            trait_id: trait_ref.trait_id,
+            substitution: Substitution::from1(db.interner(), ty),
+        }),
+    );
 }
--- a/chalk-solve/src/clauses/builtin_traits/copy.rs
+++ b/chalk-solve/src/clauses/builtin_traits/copy.rs
@@ -1,6 +1,32 @@
+use crate::clauses::builtin_traits::needs_impl_for_tys;
 use crate::clauses::ClauseBuilder;
 use crate::{Interner, RustIrDatabase, TraitRef};
-use chalk_ir::TyData;
+use chalk_ir::{ApplicationTy, Substitution, TyData, TypeName};
+
+fn push_tuple_copy_conditions<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    trait_ref: &TraitRef<I>,
+    arity: usize,
+    substitution: &Substitution<I>,
+) {
+    // Empty tuples are always Copy
+    if arity == 0 {
+        builder.push_fact(trait_ref.clone());
+        return;
+    }
+
+    let interner = db.interner();
+
+    needs_impl_for_tys(
+        db,
+        builder,
+        trait_ref,
+        substitution
+            .iter(interner)
+            .map(|param| param.ty(interner).unwrap().clone()),
+    );
+}
 
 pub fn add_copy_program_clauses<I: Interner>(
     db: &dyn RustIrDatabase<I>,
@@ -11,6 +37,12 @@ pub fn add_copy_program_clauses<I: Interner>(
     let _interner = db.interner();
 
     match ty {
+        TyData::Apply(ApplicationTy { name, substitution }) => match name {
+            TypeName::Tuple(arity) => {
+                push_tuple_copy_conditions(db, builder, trait_ref, *arity, substitution)
+            }
+            _ => return,
+        },
         TyData::Function(_) => builder.push_fact(trait_ref.clone()),
         // TODO(areredify)
         // when #368 lands, extend this to handle everything accordingly
--- a/chalk-solve/src/clauses/builtin_traits/sized.rs
+++ b/chalk-solve/src/clauses/builtin_traits/sized.rs
@@ -77,6 +77,7 @@ pub fn add_sized_program_clauses<I: Interner>(
             }
             _ => return,
         },
+        TyData::Function(_) => builder.push_fact(trait_ref.clone()),
         // TODO(areredify)
         // when #368 lands, extend this to handle everything accordingly
         _ => return,
--- a/chalk-solve/src/clauses/builtin_traits/sized.rs
+++ b/chalk-solve/src/clauses/builtin_traits/sized.rs
@@ -1,27 +1,17 @@
 use std::iter;
 
+use crate::clauses::builtin_traits::needs_impl_for_tys;
 use crate::clauses::ClauseBuilder;
 use crate::{Interner, RustIrDatabase, TraitRef};
-use chalk_ir::{ApplicationTy, Substitution, TyData, TypeName};
+use chalk_ir::{ApplicationTy, StructId, Substitution, TyData, TypeName};
 
-pub fn add_sized_program_clauses<I: Interner>(
+fn push_struct_sized_conditions<I: Interner>(
     db: &dyn RustIrDatabase<I>,
     builder: &mut ClauseBuilder<'_, I>,
     trait_ref: &TraitRef<I>,
-    ty: &TyData<I>,
+    struct_id: StructId<I>,
+    substitution: &Substitution<I>,
 ) {
-    let interner = db.interner();
-
-    let (struct_id, substitution) = match ty {
-        TyData::Apply(ApplicationTy {
-            name: TypeName::Struct(struct_id),
-            substitution,
-        }) => (*struct_id, substitution),
-        // TODO(areredify)
-        // when #368 lands, extend this to handle everything accordingly
-        _ => return,
-    };
-
     let struct_datum = db.struct_datum(struct_id);
 
     // Structs with no fields are always Sized
@@ -30,6 +20,8 @@ pub fn add_sized_program_clauses<I: Interner>(
         return;
     }
 
+    let interner = db.interner();
+
     // To check if a struct type S<..> is Sized, we only have to look at its last field.
     // This is because the WF checks for structs require that all the other fields must be Sized.
     let last_field_ty = struct_datum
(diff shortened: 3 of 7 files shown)


chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -43,3 +34,53 @@ pub fn add_sized_program_clauses<I: Interner>(
...
+            ..
+        }) => builder.push_fact(trait_ref.clone()),
+        TyData::Apply(ApplicationTy {
+            name: TypeName::Tuple(arity),
+            substitution,
+        }) => handle_tuple(db, builder, trait_ref, *arity, substitution),
+        // TODO(areredify)
+        // when #368 lands, extend this to handle everything accordingly
+        _ => return,
+    }
```
Areredify, line 85: I feel like this match is a little too verbose because we are decomposing the same variant in each of them. Also, I just realized that `TyData::Function` should be Sized too

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -43,3 +34,53 @@ pub fn add_sized_program_clauses<I: Interner>(
     };
     builder.push_clause(trait_ref.clone(), iter::once(last_field_sized_goal));
 }
+
+fn handle_tuple<I: Interner>(
```
Areredify, line 38: Handle is a little generic, maybe `push_tuple_sized_conditions` or something like that?

Areredify: Same for `handle_struct`

bjorn3:
`(dyn Copy, u8)` is not WF. Only the last field of a struct/tuple can be unsized.

```rust
error[E0277]: the size for values of type `dyn std::marker::Copy` cannot be known at compilation time
 --> src/main.rs:2:12
  |
2 |     let _: &(dyn Copy, u8) = loop {};
  |            ^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
  |
  = help: the trait `std::marker::Sized` is not implemented for `dyn std::marker::Copy`
  = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
  = note: only the last element of a tuple may have a dynamically sized type

error[E0038]: the trait `std::marker::Copy` cannot be made into an object
 --> src/main.rs:2:12
  |
2 |     let _: &(dyn Copy, u8) = loop {};
  |            ^^^^^^^^^^^^^^^ the trait `std::marker::Copy` cannot be made into an object
  |
  = note: the trait cannot be made into an object because it requires `Self: Sized`

```

If I swap `dyn Copy` and `u8`:

```rust
error[E0038]: the trait `std::marker::Copy` cannot be made into an object
 --> src/main.rs:2:12
  |
2 |     let _: &(u8, dyn Copy) = loop {};
  |            ^^^^^^^^^^^^^^^ the trait `std::marker::Copy` cannot be made into an object
  |
  = note: the trait cannot be made into an object because it requires `Self: Sized`

error: aborting due to previous error
```

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -43,3 +34,53 @@ pub fn add_sized_program_clauses<I: Interner>(
...
+
+    let interner = db.interner();
+
+    // To check if a tuple is Sized, we only have to look at its last element.
+    // This is because the WF checks for tuples require that all the other elements must be Sized.
+    let last_elem_sized_goal = TraitRef {
+        trait_id: trait_ref.trait_id,
+        substitution: Substitution::from1(interner, substitution.iter(interner).last().unwrap()),
+    };
+    builder.push_clause(trait_ref.clone(), iter::once(last_elem_sized_goal));
```
Areredify, line 59: this part is the same for both `handle`s, moreover I feel like this pattern will repeat in `Copy` impls. The pattern I mean is "to prove `trait_ref`, you need to prove the same trait for types `T1..Tn`", which is actually how rustc handles these things. Maybe it's better to encapsulate it in a helper function, something like:
```
fn needs_impl_for_tys(builder: &dyn RustIrDatabase<I>, trait_ref: &TraitRef<I>, tys: impl Iterator<Item = Ty<I>>) {
 ...
}
```

marmeladema:
> (dyn Copy, u8) is not WF
I agree it should not, but right now well-formedness of tuples in chalk is not handled properly. I guess that should be solved eventually, but i think its out of scope for this PR.

chalk-solve/src/clauses/builtin_traits/sized.rs
marmeladema: Where do you think this helper function should be defined?

Areredify: In `builtin_traits.rs`, I guess?

chalk-solve/src/clauses/builtin_traits/sized.rs
marmeladema: I've added a nested match statement for `TypeName`. I'd rather handle `Function` in another PR if possible.

Areredify: Why? Functions are always Sized, you just need to `push_fact` or whatever it's called

marmeladema: Because i don't know how to write tests for it

Areredify:
```
goal {
   fn(()): Sized
}
```
I believe this is syntax for a function that takes a unit type as input

marmeladema: Done!

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -30,3 +30,19 @@ pub fn add_builtin_program_clauses<I: Interner>(
         WellKnownTrait::DropTrait => (),
     }
 }
+
+/// Check the same trait for a list of types.
+pub fn needs_impl_for_tys<I: Interner>(
```
jackh726, line 35: I have mixed feelings about this helper function. While in theory it sounds helpful, there's a couple problems:
1) It's not even in the same file as its only *2* uses.
2) Both uses only use one clause.

Instead, I would just move this to `builtin_traits/sized.rs` and make it take just a single `Ty`

tests/test/tuples.rs
```diff
@@ -34,3 +34,51 @@ fn tuple_trait_impl() {
...
+    test! {
+        program {
+            #[lang(sized)]
+            trait Sized { }
+
+            trait Copy {}
+        }
+
+        goal {
+            (dyn Copy,): Sized
```
jackh726, line 49: This should soon not pass WF checks

tests/test/tuples.rs
```diff
@@ -34,3 +34,51 @@ fn tuple_trait_impl() {
...
+        }
+
+        goal {
+            (u8, dyn Copy): Sized
+        } yields {
+            "No possible solution"
+        }
+
+        goal {
+            (dyn Copy, u8): Sized
```
jackh726, line 61: Same here.

chalk-solve/src/clauses/builtin_traits.rs
marmeladema: It was requested by @Areredify because it's going to be used later (once this is merged) for clone and copy traits.

tests/test/tuples.rs
marmeladema: Yes, I added them so that they are actually "fixed" when tuples well-formed-ness is properly handled. This way won't forget to add such tests.

chalk-solve/src/clauses/builtin_traits.rs
jackh726: Ah, fair enough.

tests/test/tuples.rs
jackh726: So, we should probably rename `Copy` in these tests to something like `Foo`. Just so we don't get confused, because this is *not* the `Copy` trait.

Also, let's remove any instances where we're using `dyn Foo` where it's invalid. (i.e. not the last field of a tuple or struct). Really, we don't/aren't going to (at least not soon) check that the *usages* of types is WF, only definitions.

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -30,3 +30,19 @@ pub fn add_builtin_program_clauses<I: Interner>(
         WellKnownTrait::DropTrait => (),
     }
 }
+
+/// Check the same trait for a list of types.
```
nikomatsakis, line 34: This comment and function name doesn't mean much to me, I'm afraid. It'd be good to give an example of how/why it is used.

chalk-solve/src/clauses/builtin_traits/copy.rs
```diff
@@ -1,6 +1,32 @@
...
+
+    let interner = db.interner();
+
+    needs_impl_for_tys(
+        db,
+        builder,
+        trait_ref,
+        substitution
+            .iter(interner)
+            .map(|subs| subs.ty(interner).unwrap().clone()),
```
nikomatsakis, line 27: Nit: This should be called not `subs` but maybe `arg` or `param` something like that -- it represents a single value from the substitution.

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -30,16 +20,66 @@ pub fn add_sized_program_clauses<I: Interner>(
...
+    // Empty tuples are always Sized
+    if arity == 0 {
+        builder.push_fact(trait_ref.clone());
+        return;
+    }
+
+    let interner = db.interner();
+
+    // To check if a tuple is Sized, we only have to look at its last element.
+    // This is because the WF checks for tuples require that all the other elements must be Sized.
```
nikomatsakis, line 51: 👍 to comment!

nikomatsakis: It seems like `push_tuple_sized_condition` and `push_struct_sized_conditions` could be coallesced into one function, though. Something that takes a type plus a list of "constituent types" and picks the last one from the list.

tests/test/tuples.rs
```diff
@@ -34,3 +34,101 @@ fn tuple_trait_impl() {
...
+        goal {
+            (u8, dyn Foo): Sized
+        } yields {
+            "No possible solution"
+        }
+
+        goal {
+            (dyn Foo, u8): Sized
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
```
nikomatsakis, line 63: probably worth a comment here about how this is not a well-formed type

chalk-solve/src/clauses/builtin_traits/sized.rs
marmeladema: For the comment, i copied the one for struct made by @Areredify . He is the one to be thanked for :+1:

chalk-solve/src/clauses/builtin_traits/copy.rs
marmeladema: I chose param since `Substituation::iter` yields `Parameter`s

chalk-solve/src/clauses/builtin_traits/sized.rs
marmeladema: About coalescing both functions, i feel like `needs_impl_for_tys` is doing a good job here already. What would be the prototype of the function you have in mind?

chalk-solve/src/clauses/builtin_traits.rs
marmeladema: I am still not very at ease with chalk terminology, that's why I came up with such a poor comment. @Areredify could you help me out providing a better description for the helper function? Thanks!

Areredify: Uhh, something like "pushes a clause of the form `Implemented(T: Trait) :- Implemented(T1: Trait), ... Implemented(Tn: Trait)`, i.e. to prove `trait_ref` you need to prove the same trait bound on `tys`."

nikomatsakis: OK, maybe something like:

Given a trait ref `T0: Trait` and a list of types `U0..Un`, pushes a clause of the form

```
Implemented(T0: Trait) :- Implemented(U0: Trait) .. Implemented(Un: Trait)
```

chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -30,3 +30,19 @@ pub fn add_builtin_program_clauses<I: Interner>(
...
+pub fn needs_impl_for_tys<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    trait_ref: &TraitRef<I>,
+    tys: impl Iterator<Item = Ty<I>>,
+) {
```
nikomatsakis, line 40: can we add an assertion here like

```rust
// The trait must take one parameter (a type)
debug_asser_eqt!(
    db.trait_datum(trait_ref.trait_id).binders.len(),
    1,
);
```

chalk-solve/src/clauses/builtin_traits/sized.rs
nikomatsakis: Yeah, ok, I take it back. I think there *is* a kernel of logic you could extract, but it's not worth it.

chalk-solve/src/clauses/builtin_traits.rs
marmeladema: Done!

chalk-solve/src/clauses/builtin_traits.rs
marmeladema: Done!

tests/test/tuples.rs
marmeladema: I added:
```
        // It should not be well-formed because for tuples, only
        // the last element is allowed not to be Sized.
```

tests/test/tuples.rs
```diff
@@ -34,3 +34,103 @@ fn tuple_trait_impl() {
...
+
+        goal {
+            (u8, u8): Sized
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
+        }
+    }
+}
+
+#[test]
```
nikomatsakis, line 88: Can we get a similar test for `Clone`?

marmeladema: Done!

marmeladema:
CI failed because of chalk book:
```
error: Unable to retrieve "https://help.github.com/articles/about-collaborative-development-models/": https://help.github.com/articles/about-collaborative-development-models/: timed out

    ┌── contribution_guide.md:52:34 ───
    │
 52 │ We use the "fork and pull" model [described here](https://help.github.com/articles/about-collaborative-development-models/), where
    │                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ https://help.github.com/articles/about-collaborative-development-models/: timed out
    │

Error: ErrorMessage { msg: "One or more incorrect links" }
2020-04-28 20:48:39 [ERROR] (mdbook::renderer): Renderer exited with non-zero return code.
2020-04-28 20:48:39 [ERROR] (mdbook::utils): Error: Rendering failed
2020-04-28 20:48:39 [ERROR] (mdbook::utils): 	Caused By: The "linkcheck" renderer failed
##[error]Process completed with exit code 101.
```

PR #415, opened 2020-04-26
Make `dyn Trait` implement its super traits as well

flodiebold:
This implements the impl rules from #203, as far as I understood them. It doesn't do the well-formedness or object safety rules.

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -1250,6 +1250,14 @@ impl<T: HasInterner> Binders<T> {
         Self { binders, value }
     }
 
+    /// Wraps the given value in a binder without variables, i.e. `for<>
+    /// (value)`. Since our deBruijn indices count binders, not variables, this
+    /// is sometimes useful.
+    pub fn empty(interner: &T::Interner, value: T) -> Self {
+        let binders = ParameterKinds::new(interner);
+        Self { binders, value }
+    }
+
     /// Skips the binder and returns the "bound" value. This is a
     /// risky thing to do because it's easy to get confused about
     /// De Bruijn indices and the like. `skip_binder` is only valid
@@ -1287,12 +1295,39 @@ impl<T: HasInterner> Binders<T> {
         }
     }
 
+    /// Transforms the inner value according to the given function; returns
+    /// `None` if the function returns `None`.
+    pub fn filter_map<U, OP>(self, op: OP) -> Option<Binders<U>>
+    where
+        OP: FnOnce(T) -> Option<U>,
+        U: HasInterner<Interner = T::Interner>,
+    {
+        let value = op(self.value)?;
+        Some(Binders {
+            binders: self.binders,
+            value,
+        })
+    }
+
     pub fn map_ref<'a, U, OP>(&'a self, op: OP) -> Binders<U>
     where
         OP: FnOnce(&'a T) -> U,
         U: HasInterner<Interner = T::Interner>,
     {
         self.as_ref().map(op)
+    }
+
+    /// Creates a `Substitution` containing bound vars such that applying this
+    /// substitution will not change the value, i.e. `^0.0, ^0.1, ^0.2` and so
+    /// on.
+    pub fn identity_substitution(&self, interner: &T::Interner) -> Substitution<T::Interner> {
+        Substitution::from(
+            interner,
+            self.binders
+                .iter(interner)
+                .enumerate()
+                .map(|(i, pk)| (pk, i).to_parameter(interner)),
+        )
     }
 
     /// Creates a fresh binders that contains a single type
@@ -1314,6 +1349,36 @@ impl<T: HasInterner> Binders<T> {
 
     pub fn len(&self, interner: &T::Interner) -> usize {
         self.binders.len(interner)
+    }
+}
+
+impl<T, I> Binders<Binders<T>>
+where
+    T: Fold<I, I> + HasInterner<Interner = I>,
+    T::Result: HasInterner<Interner = I>,
+    I: Interner,
+{
+    /// This turns two levels of binders (`for<A> for<B>`) into one level (`for<A, B>`).
+    pub fn fuse_binders(self, interner: &T::Interner) -> Binders<T::Result> {
+        let num_binders = self.len(interner);
+        // generate a substitution to shift the indexes of the inner binder:
+        let subst = Substitution::from(
+            interner,
+            self.value
+                .binders
+                .iter(interner)
+                .enumerate()
+                .map(|(i, pk)| (pk, i + num_binders).to_parameter(interner)),
+        );
+        let value = self.value.substitute(interner, &subst);
+        let binders = ParameterKinds::from(
+            interner,
+            self.binders
+                .iter(interner)
+                .chain(self.value.binders.iter(interner))
+                .cloned(),
+        );
+        Binders { binders, value }
     }
 }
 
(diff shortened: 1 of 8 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -262,8 +261,19 @@ fn program_clauses_that_could_match<I: Interner>(
                         let qwc =
                             exists_qwc.substitute(interner, &[self_ty.clone().cast(interner)]);
 
-                        builder.push_binders(&qwc, |builder, wc| {
-                            builder.push_fact(wc);
+                        builder.push_binders(&qwc, |builder, wc| match &wc {
+                            // For the implemented traits, we need to elaborate super traits and add where clauses from the trait
+                            WhereClause::Implemented(trait_ref) => {
```
nikomatsakis, line 266: Hmm, I wonder if the actual flaw here is that for a type `dyn Foo`, we generate the fact `Implemented(dyn Foo: Foo)` instead of `FromEnv(dyn Foo: Foo)`. The latter would -- through implied bounds -- mean that `FromEnv(dyn Foo: Bar)` and so forth would come for free.

Although, I'm remembering now that there were some subtle correctness conditions to consider here -- @scalexm outlined a proposal in https://github.com/rust-lang/chalk/issues/203 that aimed to address exactly this. 

Regardless, I think that the approach in this PR is working harder than we exactly want.

flodiebold: Well yeah, scalexm's rules from #203 are what I tried to implement here :sweat_smile:

flodiebold: Using `FromEnv` would only work if we then do env elaboration on it, right?

nikomatsakis: Maybe I'm missing something. I think that if we generate the rule

```
FromEnv(dyn Foo: Foo) // rule 1
```

for any `dyn`-self type, then this would combine with the rules we genreate for the trait `Foo`:

```
// given `trait Foo: Bar`...
Implemented(?T: Foo) :- FromEnv(?T: Foo) // rule 2
FromEnv(?T: Bar) :- FromEnv(?T: Foo) // rule 3
```

such that we can prove both

```
Implemented(dyn Foo: Foo) // because rule 1 and then rule 2
```

and

```
Implemented(dyn Foo: Bar) // because rule 1, rule 3, and then rule 2
```

nikomatsakis: However, I better re-read #203 in full, because I am concerned this would also allow us to prove too much, and I forget exactly what was in there. :)

flodiebold: Oh, I think I misunderstood what role env elaboration plays, `FromEnv` would probably work then. I could try it.

But yeah, #203 suggests more complicated rules, so I thought it couldn't be that simple :wink:

nikomatsakis: In particular I think a few is something like

```rust
trait Foo<T: Copy> { ... }
```

where you might have a type like `dyn Foo<Box<u32>>`, and you want to be sure that just *seeing* a type like this isn't going to somehow let you start proving things like `Box<u32>: Copy`. 

The question kind of comes back to "who is ultimately proving that type is well-formed, and does that proof require truly proving `Box<u32>: Copy` (which would fail).

nikomatsakis: OK, re-reading the issue, I think that this is indeed probably what @scalexm meant. The rule I proposed (adding a `FromEnv` into the environment for any `dyn Foo: Foo` goal...) doesn't really make sense. The whole point of `FromEnv` is that it's supposed to be matched with some `WellFormed` obligation proved somewhere else, and in this case -- who is proving it? Nobody.

nikomatsakis:
So I've decided that this *is* going down the right direction after all. Before landing, though, we should probably convert #203 into a kind of "checklist of rules" to add.

Also, I think we should probably be extending the chalk book with documenting on these rules we're adding here.

chalk-solve/src/clauses.rs
```diff
@@ -262,8 +261,19 @@ fn program_clauses_that_could_match<I: Interner>(
...
+                                let mut seen_traits = FxHashSet::default();
```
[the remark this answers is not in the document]
flodiebold: Hmm this query would have to return something like `Binders<Vec<Binders<TraitRef>>>` (outer binders for the original trait's parameters, inner binders for `for<'a>` super trait bounds), right?

tests/test/existential_types.rs
```diff
@@ -121,6 +121,66 @@ fn dyn_Foo_Bar() {
...
+        program {
+            trait Foo<T> where Self: Bar<T> {}
+            trait Bar<T> where Self: Foo<T> {}
+
+            struct A {}
+            struct B {}
+        }
+
+        // It doesn't matter that much what comes out here since such a cycle is
+        // disallowed; mostly this is making sure that we don't crash
```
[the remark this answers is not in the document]
flodiebold: I didn't really investigate, but my guess is because you get a cycle of having to prove `dyn Foo<T>: Bar<T>` to prove `dyn Foo<T>: Foo<T>` to prove `dyn Foo<T>: Bar<T>` and so on... But I don't 100% understand how cycles in the logic are handled, so maybe that should actually be proveable...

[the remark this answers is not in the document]
nikomatsakis: Oh, yeah, of course, that makes sense. That sort of cycle would be considered unprovable today at least.

chalk-solve/src/clauses.rs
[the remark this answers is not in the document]
nikomatsakis: Yes... in rustc, we skip the outer binders, but you're correct that ideally it'd have a `Binders` type. Perhaps -- just for the purposes of this PR -- you could move this logic to a method on `TraitDatum`?

chalk-solve/src/clauses.rs
```diff
@@ -355,6 +365,70 @@ fn program_clauses_that_could_match<I: Interner>(
...
+    builder: &mut ClauseBuilder<'_, I>,
+    trait_ref: TraitRef<I>,
+    seen_traits: &mut FxHashSet<TraitId<I>>,
+) {
+    // We have some `dyn Trait`, and some `trait SuperTrait: WC`
+    // which is a super trait of `Trait` (including actually
+    // just being the same trait); then we want to push
+    // `Implemented(dyn Trait: SuperTrait) :- WC`.
+    let trait_datum = db.trait_datum(trait_ref.trait_id);
+    let wc = trait_datum
```
nikomatsakis, line 382: Speaking of helper methods, maybe it'd be nice to have e.g.

```rust
trait_datum.where_clauses().substitute(...)
```

where `where_clauses` returns `Binders<&...>`  (basically encapsulating the `binders.map_ref(..)` call).

(I think this would be useful elsewhere, too.)

chalk-solve/src/clauses.rs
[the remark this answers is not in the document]
flodiebold: Well, this turned out to be more interesting than expected, because if we have `Foo: for<'a> Bar<'a>` and `Bar<'a>: for<'b> Baz<'a, 'b>` we need to collapse the binders.

I've extracted it to a function for now; I'm not sure whether it makes sense to be on `TraitDatum`, but if you want I can move it there.

chalk-solve/src/clauses.rs
```diff
@@ -355,6 +365,70 @@ fn program_clauses_that_could_match<I: Interner>(
...
+        })
+        .substitute(db.interner(), &trait_ref.substitution);
+
+    // Add our clause: (here because we still needed the substitution)
+    builder.push_clause(trait_ref, wc);
+
+    for q_super_trait_ref in super_trait_refs {
+        // XXX really not sure about the handling of binders
+        // here, maybe it would be more correct to skip where
+        // clauses with actual binders?
```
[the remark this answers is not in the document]
flodiebold: Ok, added the test.

chalk-ir/src/lib.rs
```diff
@@ -1250,6 +1250,11 @@ impl<T: HasInterner> Binders<T> {
         Self { binders, value }
     }
 
+    pub fn empty(interner: &T::Interner, value: T) -> Self {
```
nikomatsakis, line 1253: Can you add a doc comment explaining what this does?

chalk-ir/src/lib.rs
```diff
@@ -1287,6 +1292,18 @@ impl<T: HasInterner> Binders<T> {
         }
     }
 
+    pub fn filter_map<U, OP>(self, op: OP) -> Option<Binders<U>>
```
nikomatsakis, line 1295: Doc comment

chalk-rust-ir/src/lib.rs
```diff
@@ -167,6 +167,10 @@ impl<I: Interner> TraitDatum<I> {
     pub fn is_coinductive_trait(&self) -> bool {
         self.flags.coinductive
     }
+
+    pub fn where_clauses(&self) -> Binders<&Vec<QuantifiedWhereClause<I>>> {
```
nikomatsakis, line 171: Doc comment

chalk-solve/src/clauses/dyn_ty.rs
```diff
@@ -0,0 +1,181 @@
...
+    let mut trait_refs = Vec::new();
+    go(db, trait_ref, &mut seen_traits, &mut trait_refs);
+
+    fn go<I: Interner>(
+        db: &dyn RustIrDatabase<I>,
+        trait_ref: Binders<TraitRef<I>>,
+        seen_traits: &mut FxHashSet<TraitId<I>>,
+        trait_refs: &mut Vec<Binders<TraitRef<I>>>,
+    ) {
+        let interner = db.interner();
```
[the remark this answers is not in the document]
nikomatsakis: (this doesn't really have to be changed...just a passing thought)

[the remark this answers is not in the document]
flodiebold: Yeah, I mostly put it down below because that avoids collecting and substituting those traits, but that's probably not a very important consideration...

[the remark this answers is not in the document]
nikomatsakis: Meh, efficiency is a good reason. =

flodiebold:
Rebased and added doc comments.

PR #417, opened 2020-04-28
Copy derives on intern types, from_fallible change, and linkcheck

jackh726:


.github/workflows/ci.yml
```diff
@@ -85,3 +78,25 @@ jobs:
...
+      
+      - name: Install mdbook-linkcheck
+        run: |
+          cd book
+          curl -L https://github.com/Michael-F-Bryan/mdbook-linkcheck/releases/download/v0.5.0/mdbook-linkcheck-v0.5.0-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          # Add the book directory to the $PATH
+          echo "::add-path::$GITHUB_WORKSPACE/book"
+
+      - name: Build Chalk book
+        run: cd book && ./mdbook build
```
nikomatsakis, line 102: So I think we can make this run only on PRs via something like this?

```
if: matrix.rust == 'stable' && github.ref != 'refs/heads/master'
```

but maybe that won't work the way we really want it to, I'm not sure

nikomatsakis:
feel free to merge, @jackh726, once CI is happy, I have to run for a bit

PR #422, opened 2020-04-30
Add publishing steps to book

jackh726:


book/src/publishing.md
```diff
@@ -0,0 +1,34 @@
...
+## Publishing
+- For each crate in the order above, run `cargo publish`
+    - You will probably have to wait a couple seconds between each to let the index update
+
+## Post-publish
+- Bump the minor version in each `cargo.toml` and add a `-dev` suffix
+- Bump the dependency version for each crate
+- Add an `Unreleased` section in the `RELEASES.md`
+- Run `cargo check`
+- Tag release commit on github
```
nikomatsakis, line 33: Can we add an example of what the tag looks like? I can never remember e.g. whether to use `v0.1.0` or `0.1.0`

jackh726: Added an example :)

PR #428, opened 2020-05-01
add WellKnownTraits chapter to the book

### LINKED ISSUE #416: document the "built-in trait impls" and where they live
Issues #363 and #368 describe adding support for various bits of "well-known, builtin" stuff, but it's hard to track completion with two independent lists. What we really need is a matrix. Moreover, it'd be useful for that matrix to be in the chalk book, and able to track not only *what is done* but *where the various impls and things live*.

This issue is for adding something to the chalk book (it will require a touch of research) that has roughly this form:

| Type | Copy | Clone | Sized | Unsize | Unpin | auto-traits |
| --- | --- | --- | --- | --- | --- | --- |
| tuple types | 
| scalar types |
| closures |
| ... |
| user-defined structs, enums |

with the following key:

* ✅ -- in chalk, fully implemented
* ⚪ -- in chalk, but not implemented
* 📚  -- in the stdlib, no built-in impls needed  (book == library, get it!)

OK, you can make up your own key I suppose. =) But I do like emojis!

Areredify:
I don't know what to use for n/a so I chose 🗿, if you have a better idea I'll change it
closes #416
please be(!) harsh, English is not my strongest suit

### DIFF
--- a/book/src/clauses/well_known_traits.md
+++ b/book/src/clauses/well_known_traits.md
@@ -1,0 +1,57 @@
+# Well known traits
+
+Not all traits can be encoded in Rust's type system: special traits
+like `Sized`, `Drop` or `Unsize` need additional compiler support in order to
+function properly. To address this, chalk introduces a notion of `WellKnownTrait`s:
+a subset of rustc's trait lang items that need special handling in trait system logic.
+
+As an example, consider the following two aspects of `Sized` logic:
+  1) In order to prove that a struct implements `Sized`, we need to prove 
+     that the last field of that struct is `Sized`.  
+  2) Structs need all of their fields, except, maybe, the last one to be `Sized`.
+    
+Neither of those aspects are expressable in Rust, so chalk generates 
+special clauses used to encode them. These examples illustrate two main 
+places that deal with well known traits: 
+1) [`chalk-solve\clauses\builtin_traits`][builtin_traits_mod], which generates 
+   requirements for proving that a given type implements a well known trait.
+2) [well-formedness](wf.md) checks, some of which need to know about well known traits.
+
+[builtin_traits_mod]: https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/clauses/builtin_traits.rs
+
+# Auto traits
+
+Auto traits are another kind of well known traits.
+The idea is that the type implements an auto trait if all data owned by that type implements it,
+with an ability to opt-out via special syntax:
+```rust,ignore
+impl !AutoTrait for Foo {}
+```
+Common examples of auto trais are `Send` and `Sync`. Since this semantic is not expressable with 
+"regular" impls, it needs special support in chalk too.
+
+# Current state 
+| Type            | Copy | Clone | Sized | Unsize | Drop | Fn  | Unpin  | Generator | auto traits |
+| ---             | ---  | ---   | ---   | ---    | ---  | --- | ---    |  ---      |  ---        |
+| tuple types     | ✅    | ✅    | ✅     | ✅     | 🗿    | 🗿  |  🗿      |  🗿       |   ❌         |
+| structs         | 🗿    | 🗿    |  ✅    | ✅     | 🗿    | 🗿  |  🗿      |  🗿       |   ✅         |
+| scalar types    | 📚    | 📚    | ✅     | 🗿     | 🗿   |  🗿  |  🗿     |  🗿       |    ❌        |
+| trait objects   | 🗿    | 🗿    | 🗿     |  ✅    | 🗿    | 🗿   | 🗿      |  🗿       |    🗿        |
+| functions       | ✅    | ✅    | ✅     | 🗿     | 🗿    | ❌   | 🗿      |  🗿       |    ❌         |
+| arrays❌         | ❌     | ❌    | ❌     | ❌      | 🗿   | 🗿   | 🗿      |  🗿       |    ❌        |
+| slices❌         | ❌     | ❌    | 🗿     | ❌      | 🗿   | 🗿   | 🗿      |  🗿       |    ❌       |
+| closures❌       | ❌     | ❌    | ❌     | 🗿      | 🗿   | ❌   | 🗿      |  🗿       |    ❌        |
+| generators❌     |  🗿    |  🗿  | ❌     |  🗿     | 🗿    | 🗿  | ❌      |   ❌       |    ❌       |
+| gen. witness❌   |  🗿    |   🗿  |  🗿   |   🗿    |  🗿   |  🗿 |  🗿    |   🗿       |    ❌       |
+| -----------     |       |      |       |        |      |     |        |           |             |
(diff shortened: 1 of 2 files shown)


book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
+# Well known traits
+
+Not all traits can be encoded in Rust's type system: special traits
```
jackh726, line 3: I think a better phrasing would be: "Not all trait logic can be encoded in Rust types: ..."

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+like `Sized`, `Drop` or `Unsize` need additional compiler support in order to
+function properly. To address this, chalk introduces a notion of `WellKnownTrait`s:
+a subset of rustc's trait lang items that need special handling in trait system logic.
+
+As an example, consider the following two aspects of `Sized` logic:
+  1) In order to prove that a struct implements `Sized`, we need to prove 
+     that the last field of that struct is `Sized`.  
+  2) Structs need all of their fields, except, maybe, the last one to be `Sized`.
```
jackh726, line 11: While these two points are valid, the flow is a bit weird. I would say something like
> As an example, consider the logic for `Sized` in regards to structs: A struct can have at most one `!Sized` field, and it must be the last. And the last field isn't `Sized`, then neither is the struct itself.

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+# Auto traits
+
+Auto traits are another kind of well known traits.
+The idea is that the type implements an auto trait if all data owned by that type implements it,
+with an ability to opt-out via special syntax:
+```rust,ignore
+impl !AutoTrait for Foo {}
+```
+Common examples of auto trais are `Send` and `Sync`. Since this semantic is not expressable with 
+"regular" impls, it needs special support in chalk too.
```
jackh726, line 31: Mmm, this is...not exactly correct. At least, not by the definition we've been using for well known traits. Instead I would say something like
> Auto traits, while not exactly well known traits, do also have special logic. The idea is that the type implements an auto trait if all data owned by that type implements it, with an ability to specifically opt-out or opt-in. Additionally, auto traits are *coinductive*. Some common examples of auto traits are `Send` and `Sync`.

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+| arrays❌         | ❌     | ❌    | ❌     | ❌      | 🗿   | 🗿   | 🗿      |  🗿       |    ❌        |
+| slices❌         | ❌     | ❌    | 🗿     | ❌      | 🗿   | 🗿   | 🗿      |  🗿       |    ❌       |
+| closures❌       | ❌     | ❌    | ❌     | 🗿      | 🗿   | ❌   | 🗿      |  🗿       |    ❌        |
+| generators❌     |  🗿    |  🗿  | ❌     |  🗿     | 🗿    | 🗿  | ❌      |   ❌       |    ❌       |
+| gen. witness❌   |  🗿    |   🗿  |  🗿   |   🗿    |  🗿   |  🗿 |  🗿    |   🗿       |    ❌       |
+| -----------     |       |      |       |        |      |     |        |           |             |
+| well-formedness |  ✅   |  🗿   | ✅     | 🗿     | ✅    |  🗿  | 🗿      |  🗿       |   🗿         |
+
+legend:  
+🗿 - not applicable
```
jackh726, line 50: I think maybe just leave this empty. A bit less clutter.

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+| well-formedness |  ✅   |  🗿   | ✅     | 🗿     | ✅    |  🗿  | 🗿      |  🗿       |   🗿         |
+
+legend:  
+🗿 - not applicable  
+✅ - implemented  
+📚 - implementation provided in libcore  
+❌ - not implemented  
+❌ in the column type means that type is not yet in chalk
+
+The list of types not yet in chalk is not full, but auto traits/`WellKnownTrait`s
```
jackh726, line 56: What do you mean?

book/src/clauses/well_known_traits.md
nikomatsakis: What about `⚬`? Makes it clear we didn't forget, but is unobtrusive.

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+"regular" impls, it needs special support in chalk too.
+
+# Current state 
+| Type            | Copy | Clone | Sized | Unsize | Drop | Fn  | Unpin  | Generator | auto traits |
+| ---             | ---  | ---   | ---   | ---    | ---  | --- | ---    |  ---      |  ---        |
+| tuple types     | ✅    | ✅    | ✅     | ✅     | 🗿    | 🗿  |  🗿      |  🗿       |   ❌         |
+| structs         | 🗿    | 🗿    |  ✅    | ✅     | 🗿    | 🗿  |  🗿      |  🗿       |   ✅         |
+| scalar types    | 📚    | 📚    | ✅     | 🗿     | 🗿   |  🗿  |  🗿     |  🗿       |    ❌        |
+| trait objects   | 🗿    | 🗿    | 🗿     |  ✅    | 🗿    | 🗿   | 🗿      |  🗿       |    🗿        |
+| functions       | ✅    | ✅    | ✅     | 🗿     | 🗿    | ❌   | 🗿      |  🗿       |    ❌         |
```
nikomatsakis, line 40: Does this refer to "function pointers" or `FnDef` types?

book/src/clauses/well_known_traits.md
```diff
@@ -0,0 +1,57 @@
...
+```rust,ignore
+impl !AutoTrait for Foo {}
+```
+Common examples of auto trais are `Send` and `Sync`. Since this semantic is not expressable with 
+"regular" impls, it needs special support in chalk too.
+
+# Current state 
+| Type            | Copy | Clone | Sized | Unsize | Drop | Fn  | Unpin  | Generator | auto traits |
+| ---             | ---  | ---   | ---   | ---    | ---  | --- | ---    |  ---      |  ---        |
+| tuple types     | ✅    | ✅    | ✅     | ✅     | 🗿    | 🗿  |  🗿      |  🗿       |   ❌         |
```
nikomatsakis, line 36: I wonder if these can link to the relevant enum variant, where applicable

book/src/clauses/well_known_traits.md
Areredify: Hmm, it refers to `TyData::Function` which is presume is `FnDef`

nikomatsakis: No, that's function pointer, and should be renamed.

book/src/clauses/well_known_traits.md
jackh726: That works :)

PR #430, opened 2020-05-03
.chalk file syntax writer

super-tuple:
This adds rendering lowered chalk data structure to the .chalk file syntax. 

See discussion in Zulip: https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/.2Echalk.20file.20writer

Work in progress, the remaining tasks are:

- [x] make pull request
- [x] sort produced items by `DefId` and interleave them
- [x] refactor tests
    - [x] move tests into test crate
    - [x] modularize tests
- [x] fix TODO in `ImplDatum` about missing `WriterState::add_debrujin_index` call
- [x] test formatting
- [x] disambiguate between same-name structs and traits
- [x] write macro interface to `reparse_test` function to match existing tests
- [x] split display logic into submodules
- [x] add comment justifying reparse test with different output
- [x] determine fate of `RustIrDatabase::name_*` methods
- [x] implement the `LoggingRustIrDatabase` definition-collecting intercept type
    - [x] implement `LoggingRustIrDatabase`
    - [x] test `LoggingRustIrDatabase`
    - [x] ensure we test all constructs
    - [x] fix failing tests!
- [ ] update / add support and tests for constructs since we started
    - [x] opaque type definitions
    - [x] tuples (`TypeName::Tuple`)
    - [x] scalars (`TypeName::Scalar`)
    - [x] raw (`TypeName::Raw`)
    - [x] str literals
    - [x] ref types
    - [x] slices
    - [x] const type vars
    - [ ] const type values
        - [x] test
    - [ ] opaque type usage (`TypeName::OpaqueType`)
        - [x] test
    - [ ] `AliasTy::Opaque`
        - [x] test
    - [x] `TypeName::FnDef`
    - [x] `WhereClause::LifetimeOutlives`
    - [x] `TypeName::Array`
    - [ ] Closure types
- [x] extend display & logging to include solver goals, or decide to punt to future PR - Decided to punt!
- [x] remove `dbg!()` statements
- [x] combine the two `write_program` methods
- [x] more tests for `fn()` return types (regular, w/ generics, etc.)
- [x] add more `fn()` tests involving generics
- [x] ~~investigate~~ add support for local/external `ImplType`
- [x] fix stub type picking up to fix remaining ignored logging_db tests
- [x] flags on structs and impls
- [x] support `#[repr]` on structs.
- [x] support `#[lang(*)]` on traits.
- [x] support `#[object_safe]` on traits.


chalk-solve/src/split.rs
```diff
@@ -165,6 +165,41 @@ pub trait Split<I: Interner>: RustIrDatabase<I> {
...
+    /// # Returns
+    ///
+    /// Returns the tuple of:
+    ///
+    /// * the parameters for the impl (`[Y]`, in our example)
+    /// * the parameters for the associated type value (`['a]`, in our example)
+    fn split_associated_ty_parameters<'p, P>(
+        &self,
+        parameters: &'p [P],
+        associated_ty_datum: &AssociatedTyDatum<I>,
```
nikomatsakis, line 195: Maybe `split_projection` should call this function? Seems like some duplication.

chalk-integration/src/program_writer.rs
```diff
@@ -0,0 +1,980 @@
...
+        });
+        self.impl_data.values().for_each(|datum| {
+            lines.push(datum.display(ws).to_string());
+        });
+        lines.join("\n")
+    }
+}
+
+#[cfg(test)]
+mod test {
```
nikomatsakis, line 27: These should probably live in the [`tests`](https://github.com/rust-lang/chalk/tree/master/tests) directory

chalk-integration/src/program_writer.rs
```diff
@@ -0,0 +1,980 @@
+use chalk_solve::display::{RenderAsRust, WriterState};
+
+use crate::program::Program;
+
+pub trait WriteProgram {
```
nikomatsakis, line 5: I'm not really clear on what the role of this trait is

chalk-integration/src/program_writer.rs
daboross: Thanks! Right now this file is mostly just a stopgap so we could write tests before figuring out the tests directory structure.

chalk-integration/src/program_writer.rs
daboross: It's purely for the tests, and should be removed as we create the logging database

chalk-solve/src/split.rs
daboross: Thanks for catching that!

daboross:
Thank you for the quick review! As you could probably tell, there are quite a few things left in the codebase that are holdovers from WIP code. I've replied to the individual notes.

> Gave a quick skim -- one thing I was expecting to see which I didn't yet see is something like
> 
> ```rust
> /// Wraps another `RustIrDatabase` (`DB`) and monitors the calls that occur.
> /// When the `LoggingRustIrDatabase` is dropped, it writes a serialized
> /// version of Rust program to `W`.
> struct LoggingRustIrDatabase<I, W, DB>
> where DB: RustIrDatabase<I>, W: Write, I: Interner
> {
> 
> }
> ```

This is on our todo list! We found it easier to start by implementing & testing the display code separately, but this is the end goal.

If all goes well, we should be pushing up a first implementation tonight.

super-tuple:
We are mostly complete, but there are a few outstanding design decisions, which we have described on zulip:

*  `RustIrDatabase::name_*` methods: https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/.2Echalk.20file.20writer/near/197021015

 * Wrapping `ChalkDatabase`: https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/.2Echalk.20file.20writer/near/197826479

Thanks for all the assistance, we are really excited to contribute!

daboross:
I think we've made good progress on this today, and with this commit we've fixed all but two of the outstanding issues - the remaining two being `*_name` methods, and a bug in the `LoggerIrDatabase` (more details in chalk).

There are still features missing, mostly because of the backlog of dealing with new features added to chalk. But I think these will continue to pile up, and our strategy for implementing them shouldn't be dramatically different from the existing code.

Besides those issues, though, this should be ready. If either of you have time this week, do you think you could look over it?

tests/test_util/display.rs
```diff
@@ -0,0 +1,166 @@
...
+    };
+    (program $program:tt formatting matches $res:literal) => {
+        crate::test_util::display::test_formatting(
+            crate::test_util::strip_leading_trailing_braces(stringify!($program)),
+            $res,
+        )
+    };
+}
+
+pub fn write_program(program: &Program) -> String {
```
jackh726, line 26: This looks pretty similar to the `write_program` function in `chalk-solve/display`?

tests/test_util/display.rs
```diff
@@ -0,0 +1,166 @@
...
+    for datum in program.impl_data.values() {
+        display::write_top_level(&mut out, ws, &**datum).unwrap();
+    }
+    for datum in program.opaque_ty_data.values() {
+        display::write_top_level(&mut out, ws, &**datum).unwrap();
+    }
+    out
+}
+
+fn program_diff(original: &impl Debug, produced: &impl Debug) -> String {
```
jackh726, line 44: Wait are we actually diffing the program text? I don't think this is really helpful. Moreso, we just want the goals to be solvable (or fail) in the same way.

tests/test/mod.rs
```diff
@@ -1,300 +1,12 @@
...
-                        }
-                        None => false,
-                    });
-                    assert!(expected.next().is_none(), "Not enough solutions found");
-                }
-            }
-        });
-    }
-}
+pub use super::test_util::{test::*, *};
```
jackh726, line 9: Why move these? I'm not entirely against this change, but I think this PR is big enough that if you want to do this, splitting it into a separate PR would be helpful.

tests/test_util/logging_db.rs
```diff
@@ -0,0 +1,100 @@
...
+macro_rules! logging_db_output_sufficient {
+    ($($arg:tt)*) => {{
+        use chalk_solve::SolverChoice;
+        use crate::test_util::test::*;
+        let (program, goals) = parse_test_data!($($arg)*);
+        crate::test_util::logging_db::logging_db_output_sufficient(program, goals)
+    }};
+}
+
+pub fn logging_db_output_sufficient(
```
jackh726, line 20: A little bit of a verbose function body, but it gets the job done.😂

tests/logging_db/mod.rs
```diff
@@ -0,0 +1,164 @@
+#[test]
```
jackh726, line 1: Other tests I would like to see:
- Coverage of all the builtin types (e.g. scalars, str, never, etc.)
- Coverage of functions (fn ptr and fn def)

It might be useful to "wrap" the main tests locally just to have a broader test set, to see if we missed anything.

tests/display/where_clauses.rs
```diff
@@ -0,0 +1,111 @@
...
+            trait Bax<T> { type BaxT; }
+            trait Test {
+                type Assoc<T>: Bar + Baz<Foo> + Bax<T, BaxT=T>
+                    where
+                        Foo: Bax<T, BaxT=T>,
+                        Foo: Bar,
+                        dyn Bar: Baz<Foo>;
+            }
+        }
+        produces {
```
jackh726, line 17: Theoretically, if we *do* want to do this, why not just check that the program outputs itself?

tests/logging_db/mod.rs
jackh726: Welp I just scrolled up and saw these are coveraged. Well done :)

chalk-solve/src/logging_db.rs
```diff
@@ -0,0 +1,432 @@
...
+}
+
+/// Wraps a [`RustIrDatabase`], and, when dropped, writes out all used
+/// definition to the given file.
+///
+/// Uses [`LoggingRustIrDatabase`] internally.
+///
+/// Uses a separate type, `P`, for the database stored inside to account for
+/// `Arc` or wrapping other storage mediums.
+pub struct WriteOnDropRustIrDatabase<I, W, DB, P = DB>
```
jackh726, line 210: Why do we need this? I think it's okay to just make the user output before the LoggingDatabase is dropped manually.

chalk-solve/src/lib.rs
```diff
@@ -101,6 +103,19 @@ pub trait RustIrDatabase<I: Interner>: Debug {
 
     /// Check if a trait is object safe
     fn is_object_safe(&self, trait_id: TraitId<I>) -> bool;
+
+    /// Retrieves a trait's original name. No uniqueness guarantees.
+    /// TODO: remove this, use only interner debug methods
```
jackh726, line 108: I think this is probably fine to stay. At least for this PR. (Maybe we can have a followup PR that makes these optional and adds fallback support)

chalk-solve/src/logging_db.rs
super-tuple: In a [previous comment](https://github.com/rust-lang/chalk/pull/430#pullrequestreview-405228281) @nikomatsakis suggested we make the logging database write on drop. We opted to refactor this behavior out into a separate wrapper since we felt that it was distinct, and there are likely use-cases for just logging. Do you think it's worth having?

tests/test_util/display.rs
daboross: It's diffing the `Debug` output of `Program`! I'll add some comments to it at least. This particular function is just nice output if we fail - it isn't a test. The actual test _is_ a `old_lowered_program == new_lowered_program` test, though.

We decided to use an exact comparison between `Program`s because it allowed us to really easily write a bunch of different tests without having to specify goals for every one, and that let us cover a lot more ground with less test code. With this, we can just test that a struct with fields outputs the same struct with fields, without having to figure out what exactly in chalk uses fields and to rely on that code as well.

It seems reasonable that the end goal is to keep goals solveable the same way. To that end, I'd consider these as unit tests. They test the exact behavior of this specific `display` code, and then the `logging_db` tests can assume that we render items correctly, and can focus on ensuring we get everything necessary for goals to pass (or fail).

Does that seem reasonable? We could also revise this testing strategy, but it would mean rewriting most of the `display` tests so that they have reasonable goals to follow.

tests/test_util/display.rs
super-tuple: It is similar - good catch, we missed that one. One operates on the interface provided by `Program`, the other `RustIrLoggingDb`. I think we can write a function that transforms the data in a `Program` into a list of `RecordedItemIds` similar to what we have in the logging db, and then provide that to `chalk-solve::display::write_program`, eliminating the duplication of the writing logic. I'll add this to the pull request task list.

tests/test_util/display.rs
jackh726: Yeah, honestly the test coverage is really good. And it seems pretty reasonable to just check that `old_lowered_program == new_lowered_program`. I was originally skeptical about this, because I thought it would end up being dificult. But it doesn't seem so.

chalk-solve/src/logging_db.rs
jackh726: Hmm, I guess it's simple enough to keep.

tests/logging_db/mod.rs
daboross: > It might be useful to "wrap" the main tests locally just to have a broader test set, to see if we missed anything.

We tried to do this at first, but found difficulty hooking into the existing test function which used `ChalkDatabase` to run everything - we talked a bit about this on zulip. It'd be possible to write this in `solve_goal`, and use some of the suggestions (like maybe putting a `logging` flag in `ChalkDatabase`). However, I thought that it would lead to diminishing returns compared to tests directed at `LoggingIrDatabase`, and the huge amount of extra test harness code wouldn't be worth it.

My biggest objection was that running it on the existing test suite wouldn't hit too many extra edge cases, since we'd be running it on tests which weren't designed to test `LoggingIrDatabase` edge cases. If the cost is too great, like changing things in `ChalkDatabase`, then it'd be a bust.

But thinking about this again, there might be a reasonable strategy. We could just run our test harness at the end of `solve_goal`, rather than trying to integrate our tests into `solve_goal`. To run this on all tests, we would need to essentially fully duplicate `solve_goal` in our test code. But our current code still partially duplicates it, and the only reason it isn't full duplication is to stay minimal.

If it's OK that we fully duplicate `solve_goal`'s functionality, then we'll do that.

tests/display/where_clauses.rs
daboross: The problem here is that going through the loop always adds additional clauses to an `AliasEq` where clause. Lowering always transforms `Foo: Bax<T, BaxT=T>` into `Foo: Bax<T, BaxT=T>, Foo: Bax<T>`, and trying to undo that seemed like more work than it'd be worth.

Even if we included the output program as the source, it'd just add another `Foo: Bax<T>` clause and we'd be in the same place.

I'll add a comment to that to the test code, though. That's useful information that we didn't include!

tests/test/mod.rs
daboross: Right - I think we wanted to be able to use the internals from `test!` in `logging_db_output_sufficient!` defined `test_util/logging_db.rs`, and I think there were problems using macros from `tests/test/mod.rs` in `tests/test_util/logging_db.rs` with `test/` already using macros from `test_util.rs`.

This change is a separate commit, so making a separate PR wouldn't be hard at all. We could also revert it and work around this some less invasive way - like putting the macro for `logging_db` in `tests/logging_db/mod.rs` and just letting the macro dependency chain be `logging_db -> test -> test_util`.

tests/logging_db/mod.rs
jackh726: Nah, don't worry about it.

daboross:
By the way, thanks for the review! Didn't say that earlier. I realize this has gotten to be quite a large PR in terms of lines of code.

We made some progress today, but we spent more time thinking than fixing the code. We rebased and added support for additional constructs, but didn't implement the fixes for the two big issues mentioned above: `LoggingRustIrDatabase` not including everything referenced, and missing support for `const` values.

With the design guidance from the last zulip meeting, I think we now have everything we need to fix those two issues. We'll focus directly on implementing them next time we work on this, and that will either be next Saturday or the Saturday after that.

tests/test/mod.rs
daboross: We've removed this change, and can make another PR to do the misc. fixing later.

daboross:
We've finally gotten time to work on this this weekend, and have started implementing the fix for `logging_db` missing items! Mainly, finding IDs we're missing, and outputting stub items.

We also removed the big unnecessary `tests/test/mod.rs -> tests/test_util/test.rs` refactor from this PR.

I think we discussed merging this without fully supporting everything - if that's alright, then I think the only remaining blockers are:
- one known bug in the stub items code: It doesn't pick up trait associated item bounds
- a general lack of tests for the logging db

  The display code is well tested, but the logging_db gathering framework is not. We barely have 2 working tests... It should at least have one for each top level item it can record, and some covering the intricacies of stub items.

The PR is also now re-rebased onto master.

jackh726:
Indeed I do think this is worth merging, even for partial functionality.  @daboross how would you feel if I just merged this and opened issue(s) for followup? I have to go through the code again, but last time I looked, it was fairly complete.

I'm mostly just worried about this bitrotting. Especially given that Chalk is very much a moving target right now in terms of `RustIrDatabase` stability.

daboross:
Sorry I didn't see and respond to this earlier! Had a busy week. I'll try to be on more this coming work week.

Since we didn't realize, we've got another round of fixes from today.

If you've got time to review it (or start to review it, at least) we're good with that. Next week, we could also limit our changes to rebase + minimal fixes for that if you want to start a review & it isn't completely compiling when the review is done.

chalk-solve/src/display/bounds.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn fmt(&self, s: &WriterState<'_, I>, f: &'_ mut Formatter<'_>) -> Result {
+        match self {
+            WhereClause::Implemented(trait_ref) => trait_ref.fmt(s, f),
+            WhereClause::AliasEq(alias_eq) => alias_eq.fmt(s, f),
+            WhereClause::LifetimeOutlives(lifetime) => lifetime.display(s).fmt(f),
+        }
+    }
+}
+
+/// This implementation correct inside where clauses.
```
jackh726, line 99: This doesn't make sense?

chalk-solve/src/display/bounds.rs
```diff
@@ -0,0 +1,158 @@
...
+            display_type_with_generics(
+                s,
+                self.trait_id,
+                &self.substitution.parameters(interner)[1..]
+            )
+        )
+    }
+}
+
+/// This implementation correct inside where clauses.
```
jackh726, line 116: Same here

chalk-solve/src/display/items.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        let (impl_params_in_assoc_ty_value_env, _assoc_ty_value_params) =
+            s.db.split_associated_ty_value_parameters(&param_names_in_assoc_ty_value_env, self);
+
+        let s = &s.add_parameter_mapping(
+            impl_params_in_assoc_ty_value_env.iter().cloned(),
+            impl_param_names_in_impl_env,
+        );
+
+        // let params = s
```
jackh726, line 394: This comment can be removed

tests/display/assoc_ty.rs
```diff
@@ -0,0 +1,305 @@
+#[test]
```
jackh726, line 1: Just my general thoughts on these tests: great coverage, but I think I would like these to be combined when possible. For example, the test below could just merge the two programs.

Addtionally, comments for *what* each test/program is for would be nice.

PR #432, opened 2020-05-04
Replace lalrpop-intern with string_cache

### LINKED ISSUE #431: Remove dependency on lalrpop-intern
This is a continuation of #287

Basically, `lalrpop-intern` is no longer maintained (it's actually missing in lalrpop master). In #336, we made it only be used in tests and `chalk-integration`. However, we should use a separate crate. (In #295, I converted to string_cache. I think we can probably just use most of this PR.)

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/431#issuecomment-623183917).

Mcat12:
Closes #431 

These changes are similar to #295, but in addition `ChalkIr` and the `tls` module from chalk_ir have been moved to `chalk_integration` (requested by Niko in https://github.com/rust-lang/chalk/pull/336#issuecomment-593493526).

I also added IntelliJ's `.idea` folder to gitignore, but it can be removed if that is requested.


Mcat12:
The latest commit fixed some tests which weren't caught by `cargo test`, but requires `chalk_solve` to depend on `chalk_integration` during testing. Just wanted to bring it up and make sure that this is fine.

chalk-ir/src/macros.rs
```diff
@@ -5,41 +5,41 @@ macro_rules! ty {
     (apply $n:tt $($arg:tt)*) => {
         $crate::TyData::Apply(ApplicationTy {
             name: ty_name!($n),
-            substitution: $crate::Substitution::from(&chalk_ir::interner::ChalkIr, vec![$(arg!($arg)),*] as Vec<$crate::Parameter<_>>),
-        }).intern(&chalk_ir::interner::ChalkIr)
+            substitution: $crate::Substitution::from(&chalk_integration::interner::ChalkIr, vec![$(arg!($arg)),*] as Vec<$crate::Parameter<_>>),
```
jackh726, line 8: Wait, I'm confused. `chalk-integration` isn't a dependency (dev or otherwise) of `chalk-ir`. How does this build?

jackh726: Oh, these aren't used in `chalk-ir` at all. Only `chalk-solve`. Can you just move this file there too?

Mcat12: Done, but would they fit into `chalk-macros` better?

chalk-ir/src/lib.rs
```diff
@@ -642,7 +637,7 @@ pub struct Fn<I: Interner> {
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, HasInterner)]
 pub struct Lifetime<I: Interner> {
-    interned: I::InternedLifetime,
+    pub interned: I::InternedLifetime,
```
jackh726, line 640: Why did you make this pub? You can use `interned()`

chalk-ir/src/macros.rs
nikomatsakis: No, those macros are really just the logging stuff. Should probably rename it. Or, maybe better yet, remove and adopt some more standard logging system.. :)

jackh726: Yeah, the macros in `chalk-macros` can just be moved to `chalk-engine`. Or they can just be replaced with tracing or whatever. Separate PR though :)

jackh726:
@nikomatsakis rust-analyzer does not: https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/ra_hir_ty/src/traits/chalk.rs#L804

And primarily, the advantage is just that the code doesn't exist on master anymore and thus isn't maintained. (I re-discovered this when I was going to see if we can upgrade some of our dependencies to the latest version).

nikomatsakis:
oh, right, I forgot we removed it even from LALRPOP

nikomatsakis:
ok, let's mergr!

nikomatsakis:
Well, let's fix that one nit you raised, and *then* mergE?

flodiebold:
> * @flodiebold -- is rust-analyzer using `ChalkIr`? Seems like it should define its own IR longer-term.

We've been using [our own interner](https://github.com/rust-analyzer/rust-analyzer/blob/15de338703fd9c789c4cf8041a59d8730f12bc78/crates/ra_hir_ty/src/traits/chalk.rs#L28) for a while now.

chalk-ir/src/lib.rs
Mcat12: Some code in `ChalkIr` referenced it directly. I'll change it to the method call.

Mcat12:
The nit has been fixed!

PR #433, opened 2020-05-05
Include more guidance in ambiguous results

### LINKED ISSUE #429: Better guidance for projections
I think the following test, or something similar, should pass:
```rust
#[test]
fn projection_guidance() {
    test! {
        program {
            trait Iterator { type Item; }
            #[non_enumerable]
            trait Step {}

            struct Range<T> {}

            impl<T> Iterator for Range<T> where T: Step {
                type Item = T;
            }
        }

        goal {
            exists<T> {
                exists<U> {
                    <Range<T> as Iterator>::Item = U
                }
            }
        } yields[SolverChoice::recursive()] {
            "Ambiguous; definite substitution { [?0 := ?0.0, ?1 = ?0.0] }"
        }
    }
}
```
I.e. the recursive solver should be able to tell that there is only one impl that could possibly apply, so we know the `Item` type. This is derived from the following real-world case (from rust-analyzer/rust-analyzer#4072:
```rust
enum Option<T> {}
trait Iterator {
    type Item;

    fn next(&mut self) -> Option<Self::Item>;
}

trait Step { }
impl Step for i32 {}
impl Step for u32 {}

struct Range<T>(T, T);

impl<A: Step> Iterator for Range<A> {
    type Item = A;
}

fn main() {
    let x = Range(0, 10).next();
}
```
which rustc manages to [figure out](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=16e099b6ca27a76998ab1eac3f0ab280). (This isn't related to integer type variables / fallback, it's just an easy way to get a type variable.)

The SLG solver gives `Ambiguous; definite substitution for<?U0> { [?0 := ^0.0, ?1 := (Iterator::Item)<Range<^0.0>>] }`, which is also not very helpful, but that's related to [other problems](https://github.com/rust-lang/chalk/issues/234), I think.

I haven't really looked into why this doesn't work in the recursive solver or how to fix it; I think it may be that the solver just gives up too early when one subgoal flounders (in this case the `?0: Step` goal). It could be a good first issue for getting into the recursive solver. It may be more complicated though.

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/429#issuecomment-630303770).

adamrk:
There are situations where chalk a query is ambiguous and chalk returns `Ambig(Guidance::Unknown)` even though it has deduced some non-trivial substitutions. For example, with this program:
```rust
#[non_enumerable]
trait Constraint {}
trait Trait<T> {}
struct A<T> {}
impl<T> Trait<T> for A<T> where T: Constraint {}
```
the query `exists<T,U> { A<T>: Trait<U> }` flounders because we can't determine if there is a solution to `T: Constraint` and the result is `Ambig(Guidance::Unknown)`. But even with an ambiguous result we should still be able to return the guidance that `T = U`.

The ambiguous guidance in situations like this will be useful in rust-analyzer where some queries flounder because all traits are treated as `non_enumerable` (see e.g. https://github.com/rust-analyzer/rust-analyzer/issues/4072).

The changes here check if any substitutions have been deduced when a goal flounders and returning an ambiguous guidance if the substitutions deduced so far are non-trivial.

Closes #331
Closes #429

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -985,14 +985,6 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
                 }
             }
             None => {
-                if floundered {
-                    // The strand floundered when trying to select a subgoal.
-                    // This will always return a `RootSearchFail`, either because the
-                    // root table floundered or we yield with `QuantumExceeded`.
-                    return NoRemainingSubgoalsResult::RootSearchFail(
-                        self.on_subgoal_selection_flounder(),
-                    );
-                }
                 debug!("answer is not available (or not new)");
 
                 // This table ned nowhere of interest
@@ -1054,47 +1046,6 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
         };
 
         Some(Forest::canonicalize_strand(self.context, strand))
-    }
-
-    fn on_subgoal_selection_flounder(&mut self) -> RootSearchFail {
-        debug!("all subgoals floundered");
-
-        // We were unable to select a subgoal for this strand
-        // because all of them had floundered or because any one
-        // that we dependended on negatively floundered
-
-        loop {
-            // This table is marked as floundered
-            let table = self.stack.top().table;
-            debug!(
-                "Marking table {:?} as floundered! (all subgoals floundered)",
-                table
-            );
-            self.forest.tables[table].mark_floundered();
-
-            let mut strand = match self.stack.pop_and_take_caller_strand() {
-                Some(s) => s,
-                None => {
-                    // That was the root table, so we are done.
-                    return RootSearchFail::QuantumExceeded;
-                }
-            };
-
-            if self.propagate_floundered_subgoal(&mut strand) {
-                // This strand will never lead anywhere of interest.
-                // Drop it and continue around the loop.
-                drop(strand);
-            } else {
-                // We want to maybe pursue this strand later
-                let table = self.stack.top().table;
-                let canonical_strand = Forest::canonicalize_strand(self.context, strand);
-                self.forest.tables[table].enqueue_strand(canonical_strand);
-
-                // Now we yield with `QuantumExceeded`
-                self.unwind_stack();
-                return RootSearchFail::QuantumExceeded;
-            }
-        }
     }
 
     fn on_no_strands_left(&mut self) -> Result<(), RootSearchFail> {
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -22,6 +22,7 @@ type RootSearchResult<T> = Result<T, RootSearchFail>;
 /// many strands) can fail. A root search is one that begins with an
 /// empty stack.
 #[derive(Debug)]
+#[must_use]
 pub(super) enum RootSearchFail {
     /// The table we were trying to solve cannot succeed.
     NoMoreSolutions,
@@ -277,7 +278,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                         }
                     }
                     Err(Floundered) => {
-                        debug!("Marking table {:?} as floundered!", table_idx);
+                        debug!("Marking table {:?} as floundered! (failed to create program clauses)", table_idx);
                         table.mark_floundered();
                     }
                 }
@@ -594,6 +595,7 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
                         // the SLG FACTOR operation, though NFTD just makes it
                         // part of computing the SLG resolvent.
                         if self.forest.answer(subgoal_table, answer_index).ambiguous {
+                            debug!("Marking Strand as ambiguous because answer to (positive) subgoal was ambiguous");
                             ex_clause.ambiguous = true;
                         }
 
@@ -660,6 +662,7 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
                 // We want to disproval the subgoal, but we
                 // have an unconditional answer for the subgoal,
                 // therefore we have failed to disprove it.
+                debug!("Marking Strand as ambiguous because answer to (negative) subgoal was ambiguous");
                 strand.ex_clause.ambiguous = true;
 
                 // Strand is ambigious.
@@ -890,14 +893,12 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
 
     fn on_no_remaining_subgoals(&mut self, strand: Strand<I, C>) -> NoRemainingSubgoalsResult {
         let floundered = strand.ex_clause.floundered_subgoals.len() > 0;
-        let possible_answer = if !floundered {
-            debug!("no remaining subgoals for the table");
-            self.pursue_answer(strand)
-        } else {
+        if floundered {
             debug!("all remaining subgoals floundered for the table");
-            self.pursue_answer_from_floundered(strand)
+        } else {
+            debug!("no remaining subgoals for the table");
         };
-        match possible_answer {
+        match self.pursue_answer(strand) {
             Some(answer_index) => {
                 debug!("answer is available");
 
@@ -940,6 +941,14 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
                 }
             }
             None => {
+                if floundered {
+                    // The strand floundered when trying to select a subgoal.
+                    // This will always return a `RootSearchFail`, either because the
+                    // root table floundered or we yield with `QuantumExceeded`.
+                    return NoRemainingSubgoalsResult::RootSearchFail(
+                        self.on_subgoal_selection_flounder(),
+                    );
+                }
                 debug!("answer is not available (or not new)");
 
                 // This table ned nowhere of interest
@@ -1013,7 +1022,7 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
         loop {
             // This table is marked as floundered
             let table = self.stack.top().table;
-            debug!("Marking table {:?} as floundered!", table);
+            debug!("Marking table {:?} as floundered! (all subgoals floundered)", table);
             self.forest.tables[table].mark_floundered();
 
             let mut strand = match self.stack.pop_and_take_caller_strand() {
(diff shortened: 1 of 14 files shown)


chalk-solve/src/recursive/fulfill.rs
```diff
@@ -422,7 +422,13 @@ impl<'s, 'db, I: Interner> Fulfill<'s, 'db, I> {
 
         let interner = self.solver.program.interner();
 
-        if self.infer.is_trivial_substitution(interner, &subst) {
+        if self
+            .infer
+            .canonicalize(interner, &subst)
+            .quantified
+            .value
+            .is_identity_subst(interner)
```
flodiebold, line 430: hmm what's the difference here? I think the intention was that `is_trivial_substitution` is supposed to do the same thing, maybe it needs to be fixed?

adamrk: Yeah I think it was supposed to do the same thing, but I'm not sure `is_trivial_substitution` can be made to work in this case. The clause we are trying to apply is 
```
clause=for<type> Implemented(A<^0.0>: Trait<^0.0>) :- ForAll<> { Implemented(^1.0: Constraint) })
```
so when determining if the clause can be applied we unify `$0` and `$1` (this is done in https://github.com/rust-lang/chalk/blob/e1234f618a9fbd28cd0080e25798bb21b790ca95/chalk-solve/src/recursive/fulfill.rs#L155). But that doesn't seem to actually update the `Substitution`, it just determines what new goal should be added and updates the `InferenceTable`. We don't end up solving that goal (`Implemented(^0.0: Constraint)`) so when `is_trivial_substitution` is called, neither `$0` or `$1` have been bound to anything. It's only after canonicalizing that you can detect the substitutions are non-trivial because of the unification needed to apply the clause.

flodiebold: Hmm right, actually we canonicalize anyway directly after this. So could you just put the canonicalized substitution into a variable and use that instead of canonicalizing again in the else branch?

flodiebold: Also then I think you could just remove `is_trivial_substitution`.

adamrk:
Made some changes that seem to work for the slg solver, but I still need to clean it up into something presentable.

chalk-solve/src/recursive/fulfill.rs
adamrk: done

test.chalk
```diff
@@ -0,0 +1,6 @@
+#[non_enumerable]
```
flodiebold, line 1: is this file supposed to be here?

chalk-solve/src/recursive/fulfill.rs
```diff
@@ -421,8 +421,9 @@ impl<'s, 'db, I: Interner> Fulfill<'s, 'db, I> {
         // inference as an ambiguous solution.
 
         let interner = self.solver.program.interner();
+        let canonical_subst = self.infer.canonicalize(interner, &subst);
```
Mcat12, line 424: This can be used in the `else` case below instead of recomputing the substitution.

chalk-solve/src/recursive/fulfill.rs
```diff
@@ -421,8 +421,9 @@ impl<'s, 'db, I: Interner> Fulfill<'s, 'db, I> {
...
 
-        if self.infer.is_trivial_substitution(interner, &subst) {
+        if canonical_subst.quantified.value.is_identity_subst(interner) {
```
Mcat12, line 426: This fixes #429, so can you add the test from that issue to `projection.rs`?

chalk-engine/src/logic.rs
```diff
@@ -1288,6 +1301,60 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
         }
     }
 
+    /// This strand has no subgoals left, but some have floundered.
+    /// It's still possible that we can get an ambiguous answer from it though
+    fn pursue_answer_from_floundered(&mut self, strand: Strand<C>) -> Option<AnswerIndex> {
```
jackh726, line 1306: I *think* we can merge this into `pursue_answer`. A good portion of these two are the same.

In `pursue_answer`, we know that if `!floundered_subgoals.is_empty()` then we should essentially have the logic of this function.

chalk-engine/src/logic.rs
```diff
@@ -1288,6 +1301,60 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
...
+        let answer = Answer {
+            subst,
+            ambiguous: true,
+        };
+
+        // If our answer gives trivial information on the canonicalized goal then we have nothing interesting
+        // to return.
+        let is_trivial_answer = self
+            .context
+            .is_trivial_substitution(&self.forest.tables[table].table_goal, &answer.subst);
```
jackh726, line 1344: `pursue_answer` has the `empty_constraints` condition too. We *probably* want to keep that here too.

chalk-engine/src/logic.rs
```diff
@@ -1288,6 +1301,60 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
...
+
+        if is_trivial_answer {
+            None
```
jackh726, line 1347: This is...strange. Maybe. An answer is an answer, even if it's ambiguous and trivial.

chalk-engine/src/logic.rs
```diff
@@ -504,20 +505,18 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
                             self.on_subgoal_selected(strand)?;
                             continue;
                         }
-                        SubGoalSelection::NoRemainingSubgoals => {
-                            match self.on_no_remaining_subgoals(strand) {
+                        selection @ SubGoalSelection::NoRemainingSubgoals
+                        | selection @ SubGoalSelection::Floundered => {
```
jackh726, line 509: This can actually just be simplified to `NotSelected` and `Selected`. And we know if it floundered based on if there are any floundered subgoals.

chalk-engine/src/logic.rs
```diff
@@ -946,6 +954,14 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
                 }
             }
             None => {
+                if floundered {
+                    // The strand floundered when trying to select a subgoal.
+                    // This will always return a `RootSearchFail`, either because the
+                    // root table floundered or we yield with `QuantumExceeded`.
+                    return NoRemainingSubgoalsResult::RootSearchFail(
```
jackh726, line 961: I wonder if we actually need this. I think if we mark the table as floundered in `pursue_answer` and just return `QuantumExceeded`, the correct logic will be followed the next loop around to actually handle that.

At least, that would be cleaner.

jackh726: I think the logic that would handle this is here: https://github.com/rust-lang/chalk/blob/master/chalk-engine/src/logic.rs#L1274

jackh726:
I'm also curious if this has any effect on the issue here: https://github.com/rust-lang/chalk/pull/331

test.chalk
adamrk: Nope, that was a mistake. I removed it.

chalk-solve/src/recursive/fulfill.rs
adamrk: done!

chalk-solve/src/recursive/fulfill.rs
adamrk: done!

chalk-engine/src/logic.rs
adamrk: done

chalk-engine/src/logic.rs
adamrk: done

chalk-engine/src/logic.rs
adamrk: I think I handled this (moved the call to `on_subgoal_selection_flounder` into `pursue_answer_from_floundered`), but I didn't totally follow your second comment - the link is just to the doc before `pursue_answer`. Could you explain it a bit more?

jackh726: Oops, master has changed since I made that comment, so line number is different. Here is a permalink version: https://github.com/rust-lang/chalk/blob/1f9c529f771083794a536f6fb2b1325119f8177a/chalk-engine/src/logic.rs#L1250

jackh726: As far as your change, I think all you really need to add is:
```
// This table is marked as floundered
let table = self.stack.top().table;
debug!("Marking table {:?} as floundered!", table);
self.forest.tables[table].mark_floundered();
```
instead of calling `on_subgoal_selection_flounder`

However, it's interesting that you didn't add it to the "non trivial but new answer" case. And that still works. Makes me wonder if there's a hole in our testing.

adamrk: Hmm... I tried this out and it seems to cause a bunch of other test failures. It seems like a bunch are because `select_subgoal` may now return `SubgoalSelection::NotSelected` (previously `Floundered`) even when there are remaining subgoals.

chalk-engine/src/logic.rs
adamrk: Yeah that makes sense. I tried fixing it and I think it revealed a couple other issues (which I think are unrelated):
1. It's now possible that we will push an ambiguous answer and later push the same answer is unambiguous. This causes a panic at https://github.com/rust-lang/chalk/blob/80c17b8dfadde1440a2e4ddf5236bea650cef4f2/chalk-engine/src/table.rs#L138 where we assert that exactly that situation can't occur. There's a comment there saying that callers of `push_answer` assume a response of `true` doesn't mean an existing answer was updated, but removing the panic doesn't seem to cause any problems and the test that caused the panic now passes (`test::projection::iterator_flatten`).

1. In the test `test::misc::non_enumerable_traits_reorder` a subgoal flounders, then another one succeeds and that allows the first floundered subgoal to succeed. The test is now failing with the error:
```
expected:
substitution [?0 := Foo], lifetime constraints []
actual:
Ambiguous(substitution [?0 := Foo], lifetime constraints [])
```
So it seems like somehow we're not properly updating the existing answer from ambiguous to unambiguous (or maybe not even revisiting the floundered first subgoal?). I'll look into this one some more.

jackh726: This is actually great :) (In a totally "this is a bug that's been there that is getting uncovered" kind of a way).

So that first test `iterator_flatten`: I'm looking into it a bit more, but just removing the panic is *not* okay. This is a panic because we really need to invalidate *all* answers that depends on the ambiguous answer. This somewhat leads to the second point.

For the second, I think what we need to do is fix `make_solution` to be a little bit better at the guidance it creates. The problem is really that we create an initial ambiguous answer, then later create a non-ambiguous answer. We need to check that the non-ambiguous answer "covers" the ambiguous answer. And, if so, ignore that it's ambiguous. I'm not sure that we can just ignore all ambiguous answers all together, since they perhaps represent a different set of answers. It relates to the first problem somewhat in that *maybe* we just want to "restart" the aggregation.

jackh726:
Okay, so after looking around a bit, I think I'm convinced that this isn't *quite* correct on the SLG side. Particularly, I think that adding answers when there are still floundered subgoals is not possible to do correctly *yet*. This seems to work if you *don't* push an ambiguous trivial substitution. But I think that's because we don't have any test cases that have a non-ambiguous non-trivial substitution for a strand with all floundered subgoals *and* will eventually get solved later.

jackh726:
I'm not quite sure the best way to approach this though. Part of me wonders if, instead of adding incomplete (meaning there are still floundered subgoals) answers always, we instead only do this if we can no longer make progress *anywhere*. I think we can do this by essentially doing two passes for subgoals: once where subgoals *can't* create ambiguous or incomplete answers, then a second where they *can*. (I wonder if it's even worth splitting this into three: one "complete and umabiguous", one "complete and ambiguous", and one "incomplete and ambiguous"). I guess I have to think if we have to do this *globally* or if *per-Strand* is enough. (i.e. if another Strand is followed, will subgoals of the current Strand ever become non-ambiguous. I think the answer is no, but have to look.)

jackh726:
I have a partial fix for this locally. It fixed `iterator_flatten`, but not `non_enumerable_traits_reorder` yet

jackh726:
Okay I completed the fix. (You can see them here: https://github.com/rust-lang/chalk/compare/master...jackh726:add-more-guidance)

I essentially split answer generation into "complete" and "ambiguous" phases. Basically, we always be sure to process non-ambiguous answers first. For the purposes of answers, any answer we get from strands with all floundered subgoals is ambiguous (which like in this PR we actually use).

In general, it needs a bit more cleaning up before it can be used/merged. A couple things to start:
- general cleanup
- I think `on_subgoal_selection_flounder` is never called. Or at least, it doesn't really make sense anymore
- (Not really from this PR, but I came across this) I think we can remove the call to `merge_answer_into_strand` in `on_no_remaining_subgoals` and just let the loop merge in the tabled answer in `on_subgoal_selected` (the same strand will stay active)

adamrk:
Cool, thanks! Either way sounds good to me. I could start working on some of the cleanup of your branch?

jackh726:
@adamrk I think I can finish the cleanup pretty soon. Did you enable the option for me to be able to push to your branch?

Also, these changes fix the tests in https://github.com/rust-lang/chalk/pull/331. Which is fantastic because this highlights that the changes here are a little bit more fundamental to how we handle ambiguous answers than just including more guidance :)

adamrk:
Sounds good! Yeah I turned on edits by maintainers.

jackh726:
Okay, done with cleanups.

@nikomatsakis care to take a look at this? A quick summary/tldr of this thread and PR:
In chalk-engine, we now *always* generate an answer for a strand, even if there are floundered subgoals. To correctly handle this, we split solving a goal into two stages (`AnswerMode`) either `Complete`, where we *don't* allow ambiguous answers (answers from strands with floundered subgoals count as ambiguous), and `Ambiguous` where we *do* allow ambiguous answers.

This was basically done to fix this case: https://github.com/rust-lang/chalk/blob/80c17b8dfadde1440a2e4ddf5236bea650cef4f2/chalk-engine/src/table.rs#L138. We've talked about this previously, but never had a test that actually hit this. Unfortunately, I don't think there's a better way to actually allow ambiguous answers before non-ambiguous answers.

This also has the side-effect of fixing some other tests as well, including those from #331 and #429, so these changes definitely are a bit more fundamental to how we handle ambiguous answers than just giving some extra guidance.

PR #441, opened 2020-05-06
Add the str type

### LINKED ISSUE #437: extend `TypeName` with `str` type
As part of https://github.com/rust-lang/chalk/issues/368, we need to extend chalk's `TypeName` to model the built-in `str` type.

You can follow https://github.com/rust-lang/chalk/pull/394 which provides a rough model of what needs to be extended. `str` should be roughly like `u32`.

You would need to

* add `str` to `TypeName` 
* extend parser to accept `str` as a type name, and add a suitable node to the AST
* Update the [current state](http://rust-lang.github.io/chalk/book/clauses/well_known_traits.html#current-state) table to include str. I don't think it requires any special handling, its impls live in the standard library (and it does not implement `Sized`).

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/437#issuecomment-624348784).

Mcat12:
Closes #437


tests/test/coherence.rs
```diff
@@ -264,9 +264,9 @@ fn fundamental_traits() {
     lowering_error! {
         program {
             #[upstream] trait Sized { }
-            #[upstream] struct str { }
+            #[upstream] struct str2 { } // str is already known to be !Sized, so make a new str type
```
Mcat12, line 267: This test was broken after adding `str` (and removing the struct declaration here), and I couldn't find a suitable `!Sized` replacement in std that would fit the explanation. Or, maybe I'm misinterpreting the reason the test failed (can't apply `#[upstream]` to `str` because it's predefined?).

jackh726:
I skimmed the PR, but not enough for a "review" yet. But I did want to bring up https://github.com/rust-lang/rust/pull/70911. Which *might* mean that we don't need to special type `str` in the future.

tests/test/coherence.rs
jackh726: This is fine (probably, this should have never been called `str`. Can you instead make this be named `Foo`?

Mcat12: The reason for using `str` is noted in the comment above the test:
> We want to enable negative reasoning about some traits. For example, consider the str type.
We know that str is never going to be Sized and we have made a decision to allow people to
depend on that. The following two impls are rejected as overlapping despite the fact that we
know that str will never be Sized.

I can remove the explanation around `str`, since it no longer fits with `Foo`.

Mcat12:
(BTW, I assume you want PRs to fix conflicts with master by rebasing instead of merging, like in rust-lang/rust, but let me know if this isn't the case)

tests/test/coherence.rs
jackh726: Yeah, this is just a coherence check. Nothing special about `Sized`/`str`. You're changes are good :)

jackh726:
> (BTW, I assume you want PRs to fix conflicts with master by rebasing instead of merging, like in rust-lang/rust, but let me know if this isn't the case)

Yes please :) You'll have to rebase again unfortunately

Mcat12:
The rebase is no problem :slightly_smiling_face:

PR #442, opened 2020-05-07
Add raw pointers to `TypeName`

### LINKED ISSUE #439: create typename for raw pointers
As part of https://github.com/rust-lang/chalk/issues/368, we need to extend chalk's `TypeName` to model the built-in `*const` and `*mut` types. We can add a variant `Raw(Mutability)`, where `Mutability` is the same type described in https://github.com/rust-lang/chalk/issues/438.

You can follow https://github.com/rust-lang/chalk/pull/394 which provides a rough model of what needs to be extended. You probably want to model this on how tuples are handled, a N-ary tuple like `(_, _)` has two generic type parameters, and raw pointers have one generic type parameter.

You would need to

* add `Raw(Mutability)` to `TypeName`
* extend parser to accept `*const`, `*mut` types

This issue has been assigned to @nathanwhit via [this comment](https://github.com/rust-lang/chalk/issues/439#issuecomment-625407394).

nathanwhit:
Closes #439. 

I'm a bit unsure about a few changes here, namely the lowering and the duplication of the `Mutability` enum in both the `ast` module and in `chalk-ir`, which feels a bit messy.


chalk-parse/src/ast.rs
```diff
@@ -230,6 +234,12 @@ pub enum ScalarType {
     Float(FloatTy),
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum Mutability {
```
jackh726, line 238: This is fine :) Other types are duplicated between `chalk`ir` and `chalk-parse` (like the scalars above)

book/src/clauses/well_known_traits.md
```diff
@@ -35,6 +35,7 @@ Some common examples of auto traits are `Send` and `Sync`.
 | scalar types    | 📚    | 📚    | ✅     | ⚬     | ⚬   |  ⚬  |  ⚬     |  ⚬       |    ❌        |
 | trait objects   | ⚬    | ⚬    | ⚬     |  ✅    | ⚬    | ⚬   | ⚬      |  ⚬       |    ⚬        |
 | functions ptrs  | ✅    | ✅    | ✅     | ⚬     | ⚬    | ❌   | ⚬      |  ⚬       |    ❌         |
+| raw ptrs        | ✅   |  ✅  |   ✅   |  o    |   o   |  o  |   o    |   o      |      ❌      |
```
jackh726, line 38: I think these `⚬` are different

nathanwhit: Good catch! Should be fixed now.

PR #443, opened 2020-05-07
Add slice type to `TypeName`

### LINKED ISSUE #440: extend `TypeName` with slice type
As part of https://github.com/rust-lang/chalk/issues/368, we need to extend chalk's `TypeName` to model the built-in `[T]` type. We can add a variant `Slice`.

You can follow https://github.com/rust-lang/chalk/pull/394 which provides a rough model of what needs to be extended. You probably want to model this on how tuples are handled, a N-ary tuple like `(_, _)` has two generic type parameters, and slices have one generic type parameter.

You would need to

* add `Slice` to `TypeName`
* extend parser to accept `[T]` type
* Update the [current state](http://rust-lang.github.io/chalk/book/clauses/well_known_traits.html#current-state) table to include `[T]`, though slices don't I think require any special handling (they are not `Sized`, and their other impls live in the standard library).

This issue has been assigned to @nathanwhit via [this comment](https://github.com/rust-lang/chalk/issues/440#issuecomment-625544785).

nathanwhit:
Closes #440.


nathanwhit:
> Can you add explicit tests that slices are _not_ copy or clone?

Done!

jackh726:
@nathanwhit can you rebase. And then also update the well known types table for refs since I didn't think about it before I merged your other PR 🤦‍♀️

nathanwhit:
> @nathanwhit can you rebase? And then also update the well known types table for refs since I didn't think about it before I merged your other PR 🤦‍♀️

Rebased and updated the table! Should be good to go now.

PR #444, opened 2020-05-08
Add reference types to `TypeName`

### LINKED ISSUE #438: create `TypeName` for `&` and `&mut`
As part of https://github.com/rust-lang/chalk/issues/368, we need to extend chalk's `TypeName` to model the built-in `&` and `&mut` types. We can add a variant `Ref(Mutability)`, where `Mutability` is a new type to add to chalk-ir based on [rustc's Mutability enum](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.Mutability.html). 

You can follow https://github.com/rust-lang/chalk/pull/394 which provides a rough model of what needs to be extended. You probably want to model this on how tuples are handled, a N-ary tuple like `(_, _)` has two generic type parameters, and `&` and `&mut` have two generic parameters too (a lifetime and a type).

You would need to

* add `Ref(Mutability)` to `TypeName`
* extend parser to accept `&`, `&mut` types

This issue has been assigned to @nathanwhit via [this comment](https://github.com/rust-lang/chalk/issues/438#issuecomment-625616425).

nathanwhit:
Closes #438.

Just a couple of notes: 
- I'm not totally confident in the lowering
- For parsing, I've made the assumption that references will always have an explicit lifetime specifier

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -1221,6 +1221,24 @@ impl LowerTy for Ty {
                 )?,
             })
             .intern(interner)),
+
+            Ty::Ref {
+                mutability,
+                lifetime,
+                ty,
+            } => Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
+                name: chalk_ir::TypeName::Ref(ast_mutability_to_chalk_mutability(
+                    mutability.clone(),
+                )),
+                substitution: chalk_ir::Substitution::from(
+                    interner,
+                    &[
+                        lifetime.lower(env)?.cast(interner),
+                        ty.lower(env)?.cast(interner),
+                    ],
+                ),
+            })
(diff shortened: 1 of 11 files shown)


jackh726:
@nathanwhit I think I merged this too early. `&mut` references shouldn't be copy/clone. Think you can do a followup PR?

nathanwhit:
> @nathanwhit I think I merged this too early. `&mut` references shouldn't be copy/clone. Think you can do a followup PR?

Ah my bad, I overlooked that detail. I'll have a PR up in a few minutes.

jackh726:
> > @nathanwhit I think I merged this too early. `&mut` references shouldn't be copy/clone. Think you can do a followup PR?
> 
> Ah my bad, I overlooked that detail. I'll have a PR up in a few minutes.

No problem, I didn't catch it either :) Thanks

PR #445, opened 2020-05-09
Add ObjectSafe goal and flag

### LINKED ISSUE #434: model `ObjectSafe` goals
We need to model object safety. I think a decent starting point is:

* add a `ObjectSafe(TraitId)` domain goal
* add a `RustIrDatabase` callback `is_object_safe(TraitId)`
* when we have to solve a `ObjectSafe(TraitId)` goal, we invoke the callback, and generate a fact or no fact depending on the result
* in rustc-integration, add a `#[object_safe]` flag attached to traits, and use that to test

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/434#issuecomment-625874455).

Mcat12:
Closes #434

### DIFF
--- a/chalk-integration/src/program.rs
+++ b/chalk-integration/src/program.rs
@@ -13,7 +13,7 @@ use chalk_rust_ir::{
 };
 use chalk_solve::split::Split;
 use chalk_solve::RustIrDatabase;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 use std::fmt;
 use std::sync::Arc;
 
@@ -61,6 +61,9 @@ pub struct Program {
 
     /// For each user-specified clause
     pub custom_clauses: Vec<ProgramClause<ChalkIr>>,
+
+    /// Store the traits marked with `#[object_safe]`
+    pub object_safe_traits: HashSet<TraitId<ChalkIr>>,
 }
 
 impl Program {
@@ -394,6 +397,6 @@ impl RustIrDatabase<ChalkIr> for Program {
     }
 
     fn is_object_safe(&self, trait_id: TraitId<ChalkIr>) -> bool {
-        self.trait_datum(trait_id).is_object_safe_trait()
+        self.object_safe_traits.contains(&trait_id)
     }
 }
--- a/chalk-integration/src/program.rs
+++ b/chalk-integration/src/program.rs
@@ -392,4 +392,8 @@ impl RustIrDatabase<ChalkIr> for Program {
     fn interner(&self) -> &ChalkIr {
         &ChalkIr
     }
+
+    fn is_object_safe(&self, trait_id: TraitId<ChalkIr>) -> bool {
+        self.trait_datum(trait_id).is_object_safe_trait()
+    }
 }
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -168,10 +168,6 @@ impl<I: Interner> TraitDatum<I> {
         self.flags.coinductive
     }
 
-    pub fn is_object_safe_trait(&self) -> bool {
-        self.flags.object_safe
-    }
-
     /// Gives access to the where clauses of the trait, quantified over the type parameters of the trait:
     ///
     /// ```ignore
@@ -225,9 +221,6 @@ pub struct TraitFlags {
     pub non_enumerable: bool,
 
     pub coinductive: bool,
-
-    /// Indicates that a trait is object safe
-    pub object_safe: bool,
 }
 
 /// An inline bound, e.g. `: Foo<K>` in `impl<K, T: Foo<K>> SomeType<T>`.
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -168,6 +168,10 @@ impl<I: Interner> TraitDatum<I> {
         self.flags.coinductive
     }
 
+    pub fn is_object_safe_trait(&self) -> bool {
+        self.flags.object_safe
+    }
+
     /// Gives access to the where clauses of the trait, quantified over the type parameters of the trait:
     ///
     /// ```ignore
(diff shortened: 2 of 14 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -244,16 +244,17 @@ fn program_clauses_that_could_match<I: Interner>(
...
                 .to_program_clauses(builder),
         },
-        DomainGoal::WellFormed(WellFormed::Trait(trait_predicate)) => {
-            db.trait_datum(trait_predicate.trait_id)
+        DomainGoal::WellFormed(WellFormed::Trait(trait_ref))
+        | DomainGoal::LocalImplAllowed(trait_ref) => {
+            db.trait_datum(trait_ref.trait_id)
                 .to_program_clauses(builder);
         }
+        DomainGoal::ObjectSafe(trait_id) => db.trait_datum(*trait_id).to_program_clauses(builder),
```
Mcat12, line 252: According to the function documentation, we should be more precise in the clauses we generate in order to make solving more efficient. However, most of the branches go through general clause-generating functions which generate all possible clauses. Should this ObjectSafe goal follow along and go through the general function (current implementation) or simply check the database to see if it's object safe and add the fact if so?

tests/lowering/mod.rs
```diff
@@ -505,7 +505,7 @@ fn raw_pointers() {
             impl Foo for *i32 { }
         }
         error_msg {
-            "parse error: UnrecognizedToken { token: (30, Token(51, \"i32\"), 33), expected: [\"\\\"const\\\"\", \"\\\"mut\\\"\"] }"
+            "parse error: UnrecognizedToken { token: (30, Token(52, \"i32\"), 33), expected: [\"\\\"const\\\"\", \"\\\"mut\\\"\"] }"
```
Mcat12, line 508: These tests seem pretty flaky. They seem to break when the parser adds a token. Can the test macro ignore parts of the error (regex?) so that we can keep the `expected` error data while ignoring the specific token IDs?

tests/test/object_safe.rs
```diff
@@ -0,0 +1,29 @@
+use super::*;
+
+/// The current implementation of RustIrDatabase::is_object_safe only checks for
+/// the #[object_safe] flag. This test will start failing once it performs a
+/// proper test because Foo should be object safe.
```
Mcat12, line 5: Is this explanation correct? That is, will chalk eventually check trait definitions to determine object safety, or will rustc just give us that information?

chalk-rust-ir/src/lib.rs
```diff
@@ -221,6 +225,9 @@ pub struct TraitFlags {
     pub non_enumerable: bool,
 
     pub coinductive: bool,
+
+    /// Indicates that a trait is object safe
+    pub object_safe: bool,
```
jackh726, line 230: I think we probably want a function `is_object_safe(TraitId)` instead of a flag, since we actually have to do some computation for this and we don't have to have to do that unless we need to.

chalk-solve/src/clauses.rs
jackh726: So, we don't *really* have to be all that precise with the number of clauses we generate, since all of them go through filtering before getting turned into `Strand`s. (See `program_clauses_for_goal`). However, here we probably *do* only want to generate just the fact.

chalk-rust-ir/src/lib.rs
jackh726: Okay, coming back to this. I realized that you *did* add that. This should just be removed.

chalk-integration/src/program.rs
```diff
@@ -392,4 +392,8 @@ impl RustIrDatabase<ChalkIr> for Program {
     fn interner(&self) -> &ChalkIr {
         &ChalkIr
     }
+
+    fn is_object_safe(&self, trait_id: TraitId<ChalkIr>) -> bool {
+        self.trait_datum(trait_id).is_object_safe_trait()
```
jackh726, line 397: This should just be stored in  `Program`.

tests/lowering/mod.rs
jackh726: Hmm, I wonder how useful it is to have an error msg here anyways.

tests/test/object_safe.rs
jackh726: If we *do* get to a point where Chalk is checking for object safety itself, it won't be for a bit (probably). But, I'm not sure how helpful this comment is. But that's also somewhat in part because I'm not sure what tests we *do* want in this file.

tests/test/object_safe.rs
```diff
@@ -0,0 +1,29 @@
...
+#[test]
+fn by_default_not_object_safe() {
```
jackh726, line 7: I don't think this is super helpful. For testing the flag, let's just add a `trait Bar {}` to the test below and a `goal { ObjectSafe(Bar) } yields { "No possible solution" }`

tests/test/object_safe.rs
```diff
@@ -0,0 +1,29 @@
...
+#[test]
+fn object_safe_flag() {
+    test! {
+        program {
+            #[object_safe]
+            trait Foo {}
+        }
+
+        goal { ObjectSafe(Foo) } yields { "Unique" }
+        goal { not { ObjectSafe(Foo) } } yields { "No possible solution" }
```
jackh726, line 27: We don't really need both a goal and a `not` goal. Negation and such is tested elsewhere and it so here it doesn't really *test* anything other than just adding another thing to have to think about when looking at the test.

tests/lowering/mod.rs
Mcat12: I can remove everything after `UnrecognizedToken`, or leave that for a later PR?

jackh726: I think we can leave that for a later PR :)

chalk-integration/src/program.rs
Mcat12: Should I use `HashSet`, or is there a specific kind of set that should be used (`BTreeMap` of booleans)?

PR #447, opened 2020-05-09
Mutable references should not be Copy/Clone

### LINKED ISSUE #438: create `TypeName` for `&` and `&mut`

nathanwhit:
Closes #438.


nathanwhit:
Done!

PR #448, opened 2020-05-10
Make lowering failure tests less fragile

nathanwhit:
The current lowering failure tests check that the error message resulting from the lowering attempt matches a given string exactly. This is fine in most cases, but tests which expect parsing errors are incredibly fragile in that the contents of their error messages change whenever the lexer or parser is modified.

This PR makes a small change so that lowering failure tests which produce a parsing error only have to provide the expected error kind (such as "UnrecognizedToken", etc.) instead of a full error message, which makes the tests less fragile to parser/lexer changes.

It's not exactly a beautiful fix, but it at least alleviates the immediate annoyance of tests frequently breaking.

### DIFF
--- a/tests/test_util.rs
+++ b/tests/test_util.rs
@@ -27,8 +27,18 @@ macro_rules! lowering_error {
             chalk_solve::SolverChoice::default(),
         )
         .checked_program()
-        .unwrap_err();
-        let expected = $expected;
-        assert_eq!(error.to_string(), expected.to_string());
+        .unwrap_err()
+        .to_string();
+        let expected = $expected.to_string();
+        crate::test_util::assert_same(&error, &expected);
     };
+}
+
+pub fn assert_same(result: &str, expected: &str) {
+    println!("expected:\n{}", expected);
+    println!("actual:\n{}", result);
+
+    let expected1: String = expected.chars().filter(|w| !w.is_whitespace()).collect();
+    let result1: String = result.chars().filter(|w| !w.is_whitespace()).collect();
+    assert!(!expected1.is_empty() && result1.starts_with(&expected1));
 }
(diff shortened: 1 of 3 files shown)


Mcat12:
The `test` macro allows you to only give the prefix of the output you expect (ex. "Unique") by using `assert_same`: https://github.com/rust-lang/chalk/blob/26242c00ee877b2d79eb0472bc14c70a119da02e/tests/test/mod.rs#L35

We could do the same thing for lowering tests, and then we wouldn't need the special `parse_error` syntax and it would work for non-parse errors (assuming they exist). It should also be backwards-compatible with the existing tests.
```rust
lowering_error! {
    // ...
    error_msg {
        "parse error: UnrecognizedToken"
    }
}
```

jackh726:
I agree with @Mcat12 here: we should just check whether the message starts with the expected text. That `assert_same` function can just be moved into `test_util`

nathanwhit:
Sounds reasonable to me. I've updated the PR to make that change.

tests/test_util.rs
```diff
@@ -27,8 +27,18 @@ macro_rules! lowering_error {
             chalk_solve::SolverChoice::default(),
         )
         .checked_program()
-        .unwrap_err();
-        let expected = $expected;
-        assert_eq!(error.to_string(), expected.to_string());
+        .unwrap_err()
+        .to_string();
+        let expected = $expected.to_string();
+        crate::test_util::assert_same(&error, &expected);
```
jackh726, line 33: Shouldn't need the full path here, right?

nathanwhit: Due to macro hygiene the expanded code only has access to the scope of the caller, so we can either use a full path or bring `assert_same` into scope in the calling code. I tend towards the former for portability, but I can change it if that's preferable.

jackh726: Ah yeah, macros. This is better. Thanks :)

PR #449, opened 2020-05-10
Add `FnDef` to `TypeName`

nathanwhit:
Addresses the `FnDef` part of #368.

Lowering and program clause generation are probably the areas that need the most review. I'm also a bit unsure about if/how the builtin traits apply to `FnDef`, so some guidance on that would be helpful. For parsing I've chosen to only handle function declarations for simplicity's sake (plus I'm not sure why we would need to inspect the function bodies).

In its current state the PR could use some refactoring/polish, but I'll defer that until the more substantive aspects of the PR are in good shape.

### DIFF
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -107,7 +107,6 @@ pub struct AdtFlags {
 pub struct FnDefDatum<I: Interner> {
     pub id: FnDefId<I>,
     pub binders: Binders<FnDefDatumBound<I>>,
-    pub flags: FnDefFlags,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
@@ -116,9 +115,6 @@ pub struct FnDefDatumBound<I: Interner> {
     pub return_type: Ty<I>,
     pub where_clauses: Vec<QuantifiedWhereClause<I>>,
 }
-
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub struct FnDefFlags {}
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 /// A rust intermediate representation (rust_ir) of a Trait Definition. For
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -9,9 +9,9 @@ use chalk_ir::cast::Cast;
 use chalk_ir::fold::shift::Shift;
 use chalk_ir::interner::{Interner, TargetInterner};
 use chalk_ir::{
-    AdtId, AliasEq, AliasTy, AssocTypeId, Binders, DebruijnIndex, GenericArg, ImplId, OpaqueTyId,
-    ProjectionTy, QuantifiedWhereClause, Substitution, ToGenericArg, TraitId, TraitRef, Ty, TyData,
-    TypeName, VariableKind, WhereClause, WithKind,
+    AdtId, AliasEq, AliasTy, AssocTypeId, Binders, DebruijnIndex, FnDefId, GenericArg, ImplId,
+    OpaqueTyId, ProjectionTy, QuantifiedWhereClause, Substitution, ToGenericArg, TraitId, TraitRef,
+    Ty, TyData, TypeName, VariableKind, WhereClause, WithKind,
 };
 use std::iter;
 
@@ -102,6 +102,23 @@ pub struct AdtFlags {
     pub upstream: bool,
     pub fundamental: bool,
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct FnDefDatum<I: Interner> {
+    pub id: FnDefId<I>,
+    pub binders: Binders<FnDefDatumBound<I>>,
+    pub flags: FnDefFlags,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
+pub struct FnDefDatumBound<I: Interner> {
+    pub argument_types: Vec<Ty<I>>,
+    pub return_type: Ty<I>,
+    pub where_clauses: Vec<QuantifiedWhereClause<I>>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct FnDefFlags {}
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 /// A rust intermediate representation (rust_ir) of a Trait Definition. For
--- a/chalk-rust-ir/src/lib.rs
+++ b/chalk-rust-ir/src/lib.rs
@@ -104,15 +104,45 @@ pub struct AdtFlags {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+/// A rust intermediate represention (rust_ir) of a function definition/declaration.
+/// For example, in the following rust code:
+///
+/// ```ignore
+/// fn foo<T>() -> i32 where T: Eq;
+/// ```
+///
+/// This would represent the declaration of `foo`.
+///
+/// Note this is distinct from a function pointer, which points to
+/// a function with a given type signature, whereas this represents
+/// a specific function definition.
 pub struct FnDefDatum<I: Interner> {
     pub id: FnDefId<I>,
     pub binders: Binders<FnDefDatumBound<I>>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
+/// Represents the bounds on a `FnDefDatum`, including
+/// the function definition's type signature and where clauses.
 pub struct FnDefDatumBound<I: Interner> {
+    /// Types of the function's arguments
+    /// ```ignore
+    /// fn foo<T>(bar: i32, baz: T);
+    ///                ^^^       ^
+    /// ```
+    ///
     pub argument_types: Vec<Ty<I>>,
+    /// Return type of the function
+    /// ```ignore
+    /// fn foo<T>() -> i32;
+    ///                ^^^
+    /// ```
     pub return_type: Ty<I>,
+    /// Where clauses defined on the function
+    /// ```ignore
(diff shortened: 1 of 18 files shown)


chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -337,6 +337,63 @@ impl<I: Interner> ToProgramClauses<I> for StructDatum<I> {
...
+        let binders = self.binders.map_ref(|b| &b.where_clauses);
+        builder.push_binders(&binders, |builder, where_clauses| {
+            let self_appl_ty = &ApplicationTy {
+                name: self.id.cast(interner),
+                substitution: builder.substitution_in_scope(),
+            };
+            let self_ty = self_appl_ty.clone().intern(interner);
+
+            // forall<T> {
+            //     WF(foo<T>) :- WF(T: Eq).
```
nikomatsakis, line 354: This comment kind of makes no sense without a bit more context :)

nikomatsakis: It's intended to refer into an example that would be at the top of the function. For example

```rust
fn foo<T: Eq>() 
```

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -337,6 +337,63 @@ impl<I: Interner> ToProgramClauses<I> for StructDatum<I> {
...
+                // too -- e.g., if we had `fn foo<T: for<'a> Bar<&'a i32>>()`, we would
+                // create a reverse rule like:
+                //
+                // ```notrust
+                // forall<T, 'a> { FromEnv(T: Bar<&'a i32>) :- FromEnv(foo<T>) }
+                // ```
+                //
+                // In other words, you can assume `T: Bar<&'a i32>`
+                // for any `'a` *if* you are assuming that `foo<T>` is
+                // well formed.
```
nikomatsakis, line 385: Hmm so I guess this code is copy-and-pasted from elsewhere, right? Makes me wonder if we should extract some helper functions here. Seems like it.

tests/lowering/mod.rs
```diff
@@ -559,3 +559,28 @@ fn slices() {
...
     }
 }
+
+#[test]
+fn fn_defs() {
+    lowering_success! {
+        program {
+            trait Quux { }
+
+            fn foo<'a, T>(bar: T, baz: &'a mut T) -> u32
```
nikomatsakis, line 569: woah. I guess we landed PRs for `&mut T` while I wasn't looking. Exciting. =)

tests/test/functions.rs
```diff
@@ -23,3 +23,42 @@ fn functions_are_sized() {
...
+            WellFormed(baz<Xyzzy>)
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            WellFormed(garply)
+        } yields {
+            "Unique"
+        }
```
nikomatsakis, line 60: Let's add an implied bounds test. Consider:

```
goal {
    FromEnv(baz<Bar>) => Implemented(Bar: Foo)
}
```

which I think should be provable

nathanwhit:
Thanks for the helpful feedback!

I've added the implied bounds test and cleaned up the program clause generation code quite a bit, so it should be ready for re-review.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -180,6 +180,95 @@ impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
...
+        DomainGoal::IsFullyVisible(ty.clone()),
+        appl_ty
+            .type_parameters(interner)
+            .map(|typ| DomainGoal::IsFullyVisible(typ).cast::<Goal<_>>(interner)),
+    );
+}
+
+fn implied_bounds_program_clauses<'a, I, Wc>(
+    builder: &'a mut ClauseBuilder<'_, I>,
+    id: impl CastTo<TypeName<I>>,
```
nikomatsakis, line 244: maybe rename to `type_name`?

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -180,6 +180,95 @@ impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
...
+    let ty = interned_application_ty(interner, &appl_ty);
+    builder.push_clause(
+        DomainGoal::IsFullyVisible(ty.clone()),
+        appl_ty
+            .type_parameters(interner)
+            .map(|typ| DomainGoal::IsFullyVisible(typ).cast::<Goal<_>>(interner)),
+    );
+}
+
+fn implied_bounds_program_clauses<'a, I, Wc>(
```
nikomatsakis, line 242: This could use a comment. Something like:

/// Generates the "implied bounds" clauses for an applicative 
/// type with the name `type_name`. As an example, if `type_name`
/// represents a struct `S` that is declared like:
///
/// ```rust
/// struct S<T> where T: Eq { .. }
/// ```
///
/// then we would generate the rule
///
/// ```notrust
/// FromEnv(T: Eq) :- FromEnv(S<T>)
/// ```
///
/// # Parameters
///
/// - `builder` -- the clause builder. We assume all the generic types from `S` are in scope.
/// - `type_name` -- in our example above, the name `S`
/// - `where_clauses` -- the list of where clauses declared on the type (`T: Eq`, in our example).

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -180,6 +180,95 @@ impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
...
+    let ty = appl_ty.clone().intern(interner);
+    builder.push_clause(
+        WellFormed::Ty(ty.clone()),
+        where_clauses
+            .cloned()
+            .map(|qwc| qwc.into_well_formed_goal(interner)),
+    );
+}
+
+fn fully_visible_program_clauses<'a, I>(
```
nikomatsakis, line 221: Move the comment from inside the function to a doc-comment

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -180,6 +180,95 @@ impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
...
+        name: id.cast(interner),
+        substitution: builder.substitution_in_scope(),
+    }
+}
+
+fn interned_application_ty<I: Interner>(interner: &I, appl_ty: &ApplicationTy<I>) -> Ty<I> {
+    appl_ty.clone().intern(interner)
+}
+
+fn well_formed_program_clauses<'a, I, Wc>(
```
nikomatsakis, line 198: Move comment from inside the function to a doc comment

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -180,6 +180,95 @@ impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
...
+    id: impl CastTo<TypeName<I>>,
+) -> ApplicationTy<I> {
+    let interner = builder.interner();
+    ApplicationTy {
+        name: id.cast(interner),
+        substitution: builder.substitution_in_scope(),
+    }
+}
+
+fn interned_application_ty<I: Interner>(interner: &I, appl_ty: &ApplicationTy<I>) -> Ty<I> {
```
nikomatsakis, line 194: debatable if this helper is worth it, but .. seems fine :)

chalk-rust-ir/src/lib.rs
```diff
@@ -103,6 +103,23 @@ pub struct StructFlags {
     pub fundamental: bool,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
nikomatsakis, line 106: Can we add a doc comment explaining what a `FnDef` is, giving some examples

chalk-rust-ir/src/lib.rs
```diff
@@ -103,6 +103,23 @@ pub struct StructFlags {
...
+pub struct FnDefDatum<I: Interner> {
+    pub id: FnDefId<I>,
+    pub binders: Binders<FnDefDatumBound<I>>,
+    pub flags: FnDefFlags,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
```
nikomatsakis, line 113: Add doc comment referencing `FnDefDatum`, probably, or maybe giving some more details from an example

chalk-rust-ir/src/lib.rs
```diff
@@ -103,6 +103,23 @@ pub struct StructFlags {
...
+pub struct FnDefDatumBound<I: Interner> {
+    pub argument_types: Vec<Ty<I>>,
+    pub return_type: Ty<I>,
+    pub where_clauses: Vec<QuantifiedWhereClause<I>>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct FnDefFlags {}
```
nikomatsakis, line 121: Hmm an empty set of flags may not be worth including unless we expect to put things in here...

chalk-rust-ir/src/lib.rs
```diff
@@ -103,6 +103,23 @@ pub struct StructFlags {
...
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct FnDefDatum<I: Interner> {
+    pub id: FnDefId<I>,
+    pub binders: Binders<FnDefDatumBound<I>>,
+    pub flags: FnDefFlags,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
+pub struct FnDefDatumBound<I: Interner> {
```
nikomatsakis, line 114: I feel like we are going to need .. well, eventually we'll have to add ABI and other details, but maybe we can add that together with adding the same details to the function pointer types

nikomatsakis:
r=me but this needs to be rebased.

nikomatsakis:
And we should probably file a followup issue regarding ABI details (which apply equally to function pointers)

nathanwhit:
Done and done!

Issue #435, opened 2020-05-05
extend chalk to include outlives goals


nikomatsakis:
As discussed in our meeting today, [starting around here](https://zulip-archive.rust-lang.org/144729wgtraits/54350designmeeting20200505.html#196364729), we want to extend chalk to handle region/type outlives goals and constraints to enable rustc integration. The steps are probably:

* add region outlives goals that generate region outlives constraints
* use region outlives goals in unification to simplify things
* add type outlives goals and constraints (and don't address the for for<'a> T: 'a thing)-
* connect to rustc

jackh726:
I'm going to work a bit on the region outlives.

rvolgers:
Just out of curiosity, would this end up solving https://github.com/rust-lang/rust/issues/25860 ?

nikomatsakis:
Not directly. It's part of a larger plan that will address that problem, though.

nathanwhit:
I've been doing some work on the type outlives goals/constraints.

jackh726:
This is done now

PR #451, opened 2020-05-11
Add lifetime outlives goal

jackh726:
c.c. #435 

I'm not sure if this is *exactly* the approach we want to take here, but I figured I would open this for comments at least

### DIFF
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -5,7 +5,8 @@ use crate::{ExClause, Literal, TimeStamp};
 use chalk_ir::debug;
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    Environment, Fallible, Goal, GoalData, InEnvironment, QuantifierKind, Substitution,
+    Constraint, DomainGoal, Environment, Fallible, Goal, GoalData, InEnvironment, QuantifierKind,
+    Substitution, WhereClause,
 };
 
 impl<I: Interner, C: Context<I>> Forest<I, C> {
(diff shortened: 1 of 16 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -244,6 +244,15 @@ fn program_clauses_that_could_match<I: Interner>(
                 .opaque_ty_data(opaque_ty.opaque_ty_id)
                 .to_program_clauses(builder),
         },
+        DomainGoal::Holds(WhereClause::LifetimeOutlives(a, b)) => {
+            builder.push_clause(
+                DomainGoal::Holds(WhereClause::LifetimeOutlives(a.clone(), b.clone())),
+                Some(DomainGoal::Holds(WhereClause::LifetimeOutlives(
```
nikomatsakis, line 250: This...doesn't seem right. This says

```
Holds('a: 'b) :- Holds('a: 'b)
```

which is kind of .. tautological.

jackh726:
@nikomatsakis Okay so I rebased this to master. But I'm a little unsure how to continue here.

1) The `AddRegionConstraint` goal isn't really possible to do now that `chalk-engine` uses `Goal` directly
2) I looked at your branch in regards to having unification add goals instead of constraints. This doesn't quite fit into the current scheme since there's no translation there of `Outlives -> AddRegionConstraint`.
3) Can you elaborate on your last point? (applying `AddRegionConstraint` for things in the root universe, and higher-ranked things) Maybe a test case?

chalk-engine/src/simplify.rs
```diff
@@ -68,14 +69,22 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                     &goal.b,
                     &mut ex_clause,
                 )?,
-                GoalData::DomainGoal(domain_goal) => {
-                    ex_clause
-                        .subgoals
-                        .push(Literal::Positive(InEnvironment::new(
+                GoalData::DomainGoal(domain_goal) => match domain_goal {
```
nikomatsakis, line 72: it seems like this logic is missing from the recursive solver?

tests/test/misc.rs
```diff
@@ -612,3 +612,33 @@ fn normalize_ambiguous() {
...
+            struct Bar {}
+
+            impl<'a, 'b> Foo<'a, 'b> for Bar where 'a: 'b {}
+        }
+
+        goal {
+            exists<'a, 'b> {
+                Bar: Foo<'a, 'b>
+            }
+        } yields[SolverChoice::slg(10, None)] {
```
nikomatsakis, line 630: ah I guess that's why you have `slg` here

chalk-engine/src/simplify.rs
nikomatsakis: I think you would add it here

https://github.com/rust-lang/chalk/blob/ffa2e515a4fac85f053dcbe6307d73ebc0e82043/chalk-solve/src/recursive/fulfill.rs#L174-L180

but it'd be nice to refactor the two solvers to share this logic.

nikomatsakis:
@jackh726 

So, I think that your current branch is effectively "hard-coding" the `AddRegionConstraint` logic in some sense, by baking it into the "simplification" logic for handling domain goals (rather than adding a special sort of goal that is baked in, and then "converting" to that via a standard program clause). 

I don't quite understand your point about having refactored the `resolvent` special handling, and I don't quite understand what you mean about not being able to modify unification, but I can maybe try to do the thing I was talking about and see what happens.

I *think* I still prefer to have a special goal like `AddRegionConstraint` (or maybe a built-in notion of `Outlives`) so that the solver never tries to "bake in" what a domain-goal means, but it's not a really big difference.

As for the interaction with higher-ranked, let's shelve it for now, I don't think it's relevant to this PR yet.

PR #454, opened 2020-05-13
Rename struct types to ADT types and introduce new interned ADT ID

### LINKED ISSUE #453: generalize the builtin struct type to adt to match rustc
Chalk currently has a notion of [struct types](http://rust-lang.github.io/chalk/chalk_ir/enum.TypeName.html#variant.Struct) but rustc has the more general [adt](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.TyKind.html#variant.Adt). Also, chalk expects a `DefId` but in rustc we use an `AdtDef` (a pointer to an interned structure).

We should align with rustc here:

* Rename `TypeName::Struct` to `TypeName::Adt`
* Rename `StructId<I>` to `AdtId<I>`
* Extend `Interner` trait to have `I::InternedAdtId` and adapt `AdtId<I>` to be a wrapper around `InternedAdtId` (instead of `I::DefId`, [as today](http://rust-lang.github.io/chalk/chalk_ir/struct.StructId.html))
* We can map `InternedAdtId` to the same type as `DefId` in our integration, rustc can use `&'tcx AdtDef`
* Rename [`StructDatum`](http://rust-lang.github.io/chalk/chalk_rust_ir/struct.StructDatum.html) and [`StructDatumBound`](http://rust-lang.github.io/chalk/chalk_rust_ir/struct.StructDatumBound.html) to use the `Adt` terminology -- we don't actually have to change the list of fields for now, though eventually we may want to do so to allow rustc to more cheaply give access from the `AdtDef`

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/453#issuecomment-628011487).

Mcat12:
Closes #453

The only functional changes are due to the introduction of the ADT ID, and they are minimal.

### DIFF
--- a/chalk-ir/src/interner.rs
+++ b/chalk-ir/src/interner.rs
@@ -1,3 +1,4 @@
+use crate::AdtId;
 use crate::AliasTy;
 use crate::ApplicationTy;
 use crate::AssocTypeId;
@@ -20,7 +21,6 @@ use crate::ProjectionTy;
 use crate::QuantifiedWhereClause;
 use crate::QuantifiedWhereClauses;
 use crate::SeparatorTraitRef;
-use crate::StructId;
 use crate::Substitution;
 use crate::TraitId;
 use crate::Ty;
@@ -127,7 +127,7 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
     /// and can be converted back to its underlying data via `quantified_where_clauses_data`.
     type InternedQuantifiedWhereClauses: Debug + Clone + Eq + Hash;
 
-    /// "Interned" representation of a list of variable kinds.  
+    /// "Interned" representation of a list of variable kinds.
     /// In normal user code, `Self::InternedVariableKinds` is not referenced.
     /// Instead, we refer to `VariableKinds<Self>`, which wraps this type.
     ///
@@ -135,7 +135,7 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
     /// and can be converted back to its underlying data via `variable_kinds_data`.
     type InternedVariableKinds: Debug + Clone + Eq + Hash;
 
-    /// "Interned" representation of a list of variable kinds with universe index.  
+    /// "Interned" representation of a list of variable kinds with universe index.
     /// In normal user code, `Self::InternedCanonicalVarKinds` is not referenced.
     /// Instead, we refer to `CanonicalVarKinds<Self>`, which wraps this type.
     ///
@@ -144,8 +144,11 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
     /// to its underlying data via `canonical_var_kinds_data`.
     type InternedCanonicalVarKinds: Debug + Clone + Eq + Hash;
 
-    /// The core "id" type used for struct-ids and the like.
+    /// The core "id" type used for trait-ids and the like.
     type DefId: Debug + Copy + Eq + Ord + Hash;
+
+    /// The ID type for ADTs
+    type InternedAdtId: Debug + Copy + Eq + Ord + Hash;
 
     type Identifier: Debug + Clone + Eq + Hash;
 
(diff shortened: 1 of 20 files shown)


chalk-solve/src/wf.rs
```diff
@@ -144,7 +144,8 @@ where
         Self { db, solver_choice }
     }
 
-    pub fn verify_struct_decl(&self, struct_id: StructId<I>) -> Result<(), WfError<I>> {
+    /// TODO: Currently only handles structs, may need more work for enums & unions
+    pub fn verify_adt_decl(&self, adt_id: AdtId<I>) -> Result<(), WfError<I>> {
```
Mcat12, line 148: I'm not sure if this function generalizes to all ADTs in its current form, so I flagged it with this TODO.

chalk-ir/src/interner.rs
```diff
@@ -144,9 +144,12 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
     /// to its underlying data via `canonical_var_kinds_data`.
     type InternedCanonicalVarKinds: Debug + Clone + Eq + Hash;
 
-    /// The core "id" type used for struct-ids and the like.
+    /// The core "id" type used for trait-ids and the like.
     type DefId: Debug + Copy + Eq + Ord + Hash;
 
+    /// The ID type for ADTs
+    type InternedAdtId: Debug + Copy + Eq + Ord + Hash;
```
jackh726, line 151: So, I'm not actually sure why we need this. @nikomatsakis on your issue, you said that for rustc this would be `&'tcx AtdDef`. But we can access that with `tcx.adt_def(DefId)`. So...I'm not really sure why `Adt`s need their own special id.

nikomatsakis: The reason is that, eventually, we want rustc's types to be equal to chalk's types. We could potentially convert rustc to use a `DefId` and not a `&'tcx AdtDef`, and I considered that, but it would mean doing a query every time we want to access field information, so it might imply extra overhead. 

To be honest, I suspect that we may want to take this same approach for *all* ids, to give `Interner` implementations more flexibility in how they manage their ids, but it seems ok to start here.

We could always remove it later if converting rustc to use `DefId` is better, or we could go further, or leave it as is.

jackh726: Hmm. I can understand the sentiment towards wanting to not have to query every time we want to access field information. But I don't think we should do that *here*, for a couple reasons:
1) It's different than what we do for everything else (right now). I would rather things be consistent than trying to optimize for less overhead. (This is, in part, because we don't exactly know how much overhead there would be. And everything else isn't optimized at all.)
2) I think this is possibly just *wrong*. If we want to store the `AdtDef` (or any other info for other types), then I think it should be stored in `InternedType`, *not* in the equivalent of `DefId`.

nikomatsakis: I can see the argument for uniformity, but I don't understand the argument for it being wrong. From the perspective of chalk, it **is** an id -- an id is just something opaque that we give to the `RustIrDatabase` in exchange for information. 

The only difference here is that we've given the `Interner` (and `RustIrDatabase`) more license to decide how to represent ids (as I noted above, I suspect we ought to do this for all ids).

nikomatsakis: As for performance, it's true that we haven't (in rustc) measured the impact of replacing `AdtDef` with a `DefId`. It's also true that when we *added* `AdtDef` (instead of `def_id`), it was a noticeable win. But the compiler was a very different beast back then.

jackh726: Ok, I'll concede.

nikomatsakis: Well, to be clear, I do think you have a point about performance and uniformity. And I resisted suggesting we introduce this indirection for some time because I wanted to try converting rustc to use a `DefId` instead for precisely this reason. And I still sort of do. But I also think we can do that work independently by adding the indirection, which is why I wound up suggesting it.

PR #455, opened 2020-05-14
refactor Parameter into GenericArg

### LINKED ISSUE #452: refactor `ParameterKind` to match rustc, accomodate constants
In order to resolve some of the problems we encountered in integrating const generics, we want to do the following refactorings:

* replaces `type Parameter = ParameterKind<...>` with a `GenericArg` (copying Rust's name) -- for bonus points, it should be an interned struct as well with a `data` method that gets a `GenericArgData`
* and refactors `ParameterKind<()>` into `BoundVariableKind<I>` or something like (which for now is just type/lifetime and I guess a phantomdata for the `I` parameter)
* and refactors `ParameterKind<EnaVariable>` into a `(BoundVariableKind, EnaVariable)` pair

Then we can, in the const generics PR:

* extend `BoundVariableKind` with `Const(Ty<I>)`
* extend `GenericArg` with a `Const(Const<I>)` variable, and `Const<I>` can carry the `Ty<I>` (as rustc handles it)

Areredify:
Some notes: 
1. I decided to leave most of the entries for `*parameters*` functions and variables, because it flows better, while `parameter` was refactored into `generic_arg` where applicable. 
2. I am not sure about `ParameterEnaVariable` ( = `WithKind<I, EnaVariable<I>>`), but `GenericArgEnaVariable` is too much. All other names are open to bikeshedding; speak now, or forever hold your peace.
3. there is a suspicious place that I marked with a `FIXME(areredify)`, which I will deal with in the const generic pr  
closes #452

### DIFF
--- a/book/src/types.md
+++ b/book/src/types.md
@@ -57,8 +57,6 @@ not been thoroughly discussed by the Rust compiler team as a whole.
 Here is a (partial) list of some things that have to be adapted in
 Chalk as of today to match this document:
 
-* `Parameter<I>` needs to be renamed to `GenericArgument`
-* `Vec<Parameter<I>>` needs to be replaced with `GenericArguments`
 * Extract `TypeName` into something opaque to chalk-ir.
 * Dyn type equality should probably be driven by entailment.
 * Projections need to be renamed to aliases.
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -103,7 +103,7 @@ pub trait Context: Clone + Debug {
 
     /// A term that can be quantified over and unified -- in current
     /// Chalk, either a type or lifetime.
-    type Parameter: Debug;
+    type GenericArg: Debug;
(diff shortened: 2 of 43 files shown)


Areredify:
@jackh726 we don't yet, but we need `I: Interner` bound to add a `Const(Ty<I>)` variant to `VariableKind`, so it's  just a placeholder

jackh726:
Ok, then if it's just a placeholder for Const, then let's just change it to Const and change unreachables to unimplementeds

Areredify:
What for? I don't see how it's different from `LifetimeData` having phantom data, and semantically it's not `Const`, it's just a way for `VariableKind` to have the interner parameter. Also, all `unreachable`s are in fact unreachable, because `Variable::Phantom` contains `Void`

jackh726:
Ok, you've convinced me. And @nikomatsakis already approved.

Issue #399, opened 2020-04-15
coinduction semantics in the recursive solver


nikomatsakis:
The recursive solver introduced by @flodiebold  in https://github.com/rust-lang/chalk/issues/351 isn't handling coinduction correctly. A number of the tests have "FIXME" results. ([See comment.](https://github.com/rust-lang/chalk/pull/372#discussion_r408708780))

nikomatsakis:
So I [described in high-level terms](https://zulip-archive.rust-lang.org/144729wgtraits/21846coinductionintherecursivesolverchalk399.html#199203174) the way I think we should implement this. It builds on the [way that the recursive solver handles cycles now](http://rust-lang.github.io/chalk/book/recursive/search_graph.html).

### How inductive solver works

The basic structure for handling cycles in the recursive solver is "iterate until we reach a fixed point". For inductive cycles, I think the mental model is that we are "discovering more and more answers". Therefore we have a kind of lattice that works like this (values begin at the first entry and move towards the final one):

* Error (zero answers)
* Some unique answer with a substitution; I'll write this as `Unique(T)` but the idea is really that there is `Unique(S)` where `S` is some substitution like `[V0 = T0, ..., Vn = Tn]` where `V0..Vn` are the input variables from the goal
* Ambiguous (multiple answers, or answers that may or may not be true because of floundering)

We start out with the result of error (i.e., no answers) and then we try to solve the goal via all the "routes" available to us (i.e., applying each possible program clause). Each clause might have back an answer in the form above. If for example we have two clauses, and we get back Error from one and `Unique(Vec<u32>)` from the other, then the answer would be `Unique(Vec<u32>)`. Similarly, if any clause gives back `Ambiguous`, then the result is always ambiguous.

If multiple clauses give back answers, though, things can be a bit more complicated. If the answers are "incompatible", then the result is just ambiguous: so if our two clauses gave back `u32` and `i32`, then we can't find a single unique answer that encompasses both of them (but which doesn't also encompass other things). That is, we couldn't summarize as `?T`, because that would be overly general.

There are also cases where neither is more general than the other. For example, consider two answers like `(?T, u32)` and `(u32, ?U)`. Both are instantiations of the other, but neither is strictly more general. We have to return ambiguous for this, because `(?T, ?U)` would admit too many answers, and `(u32, u32)` would be too narrow.

[...]

PR #457, opened 2020-05-15
Improve coinduction handling in recursive solver

zaharidichev:
This change is an attempt to handle coinduction semantics better when solving with the recursive solver. The main idea is that we introduce the concept of solutions with delayed goals that can be refined later. The main concepts behind the approach are borrowed from: https://rust-lang.github.io/chalk/book/engine/logic/coinduction.html Would love to know if I missed something obvious. 

Fix: #399

### DIFF
--- a/chalk-solve/src/recursive.rs
+++ b/chalk-solve/src/recursive.rs
@@ -16,10 +16,37 @@ use rustc_hash::FxHashMap;
 
 type UCanonicalGoal<I> = UCanonical<InEnvironment<Goal<I>>>;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Answer<I: Interner> {
+    pub solution: Solution<I>,
+    pub delayed_goals: Vec<UCanonicalGoal<I>>,
+}
+
+impl<I: Interner> Answer<I> {
+    pub fn new(solution: Solution<I>, delayed_goals: Vec<UCanonicalGoal<I>>) -> Self {
+        Self {
+            solution,
+            delayed_goals,
+        }
+    }
+
+    pub fn needs_refinement(&self) -> bool {
+        !self.delayed_goals.is_empty()
+    }
+
+    pub fn combine(self, other: Answer<I>, interner: &I) -> Self {
+        let mut goals = vec![];
+        let combined_solution = self.solution.combine(other.solution, interner);
+        goals.extend(self.delayed_goals);
+        goals.extend(other.delayed_goals);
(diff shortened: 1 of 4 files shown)


shepmaster:
> reecursive

typo

jackh726:
@flodiebold do you think you could take a look at this?

flodiebold:
It seems fine code-wise, but I don't really know a lot about coinduction or how it works, so probably it would be better if @nikomatsakis can take a look as well.

chalk-solve/src/recursive.rs
```diff
@@ -138,11 +199,15 @@ impl<'me, I: Interner> Solver<'me, I> {
...
-                        value,
-                        binders: goal.canonical.binders,
-                    }));
+                    return Ok(Answer::new(
+                        Solution::Unique(Canonical {
+                            value,
+                            binders: goal.clone().canonical.binders,
+                        }),
+                        vec![goal],
+                    ));
```
nikomatsakis, line 210: So I was to think a bit about this -- in particular, I think the recursive solver can handle this case differently than the SLG solver. Unlike the SLG solver, it's ok for intermediate queries in the recursive solver to not get cached (and/or get cached at the end, along with everything else). I have to review how we handle caching and minimums, but I suspect that we want to work towards a setup where the cache itself **never** has delayed goals, they are always ephemeral things that exist on the stack during solving.

zaharidichev:
@nikomatsakis Thanks a bunch for the feedback. I will address the comments and revise my logic here. If you happen to arrive at any more insights, feel free to share :)

zaharidichev:
@nikomatsakis I think I am a bit stuck on that. I understand the overall intended approach but have a few questions: 

- In .fulfill() we effectivelly want to check whether the delayed goals that we get from solving a subgoal are actually trivial cycles. I wonder what is the most practical way to do that since the goal that fullfill is instantiated with could be of type ` UCanonical<InEnvironment<Goal<I>>>` or ` UCanonical<InEnvironment<DomainGoal<I>>>` but what we get as delayed goals in the `Answer` is  `UCanonical<InEnvironment<Goal<I>>>` Could there be a delayed DomainGoal actually? 

- Is it correct to assume that we want to store all of these `partially solved` goals on the stack and attempt to solve all of their delayed subgoals once we enchounter them again and only if that succeeds put the solution in the cache. 

For some reason all of this makes my head hurt at times :)

jackh726:
Closing this. I think we want to go with the alternative approach @nikomatsakis described.

PR #458, opened 2020-05-16
Bump most dependency crate version.

crlf0710:
This bumps all dependencies expect `salsa` to newest version.

Upgrading salsa is not straight forward because of the removal of `volatile` attribute, so i'm leaving it out.

This also reduced the total num of crates in the dependency tree.


jackh726:
Thanks! I looked into doing this before, but didn't finish. I'm curious what it's going to take to get salsa upgraded. Maybe @nikomatsakis knows

PR #459, opened 2020-05-17
Cleanup crate structure and add features for SLG/recursive solvers

### LINKED ISSUE #456: Crate structure cleanup
There are a few different crate structure cleanups we want to do.

* [ ] Move recursive solver out of `chalk-solve`
  - A lot of the functionality here overlaps with `chalk-engine`, so it would be nice try to set up the crate structure in a way that we only need to compile one or the other.
* [ ] Merge chalk-rust-ir into chalk-solve
  - This is a pretty small crate and there's not really a benefit to keeping it split.
* [ ] Move chalk-macros into chalk-engine
  - chalk-macros is also pretty small. This might be a little different, depending on the how the first bullet point goes. But basically, this doesn't need to be all by itself.

This issue has been assigned to @Mcat12 via [this comment](https://github.com/rust-lang/chalk/issues/456#issuecomment-629698312).

Mcat12:
Closes #456 

- The `chalk-rust-ir` crate is moved into `chalk-solve` as the `rust_ir` module.
- `chalk-macros` is renamed to `chalk-base` and some types are moved there from `chalk-engine` so that the recursive solver does not depend on `chalk-engine`.
- The SLG and recursive solvers have their own feature flags in `chalk-solve` so that only one solver needs to be compiled. By default, both features are enabled.


chalk-solve/Cargo.toml
```diff
@@ -15,11 +15,16 @@ itertools = "0.9.0"
...
+chalk-ir = { version = "0.10.1-dev", path = "../chalk-ir", default-features = false }
 
 [dev-dependencies]
 chalk-integration = { version = "0.10.1-dev", path = "../chalk-integration" }
+
+[features]
+default = ["slg-solver", "recursive-solver"]
+
+slg-solver = ["chalk-engine", "chalk-ir/slg-solver"]
+recursive-solver = []
```
Mcat12, line 30: I ran into https://github.com/rust-lang/cargo/issues/7916 while testing the feature flags, but I don't think it affects downstream users (like RA). The issue is that a `cargo check --no-default-features --features recursive-solver` in `chalk-solve` would still compile `chalk-engine` because the dev-dependency `chalk-integration` implicitly enables `chalk-ir`'s `slg-solver` feature. If the dev-dependency features are not unified, this issue goes away (tested on nightly with `-Z features=dev_dep`).

chalk-engine-base/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
-name = "chalk-macros"
+name = "chalk-engine-base"
```
jackh726, line 2: I don't particularly like the name `chalk-engine-base`. Maybe just `chalk-base`?

jackh726: Or, since `chalk-ir` is getting a feature flag. Just have this be part of `chalk-engine` with all SLG stuff under a feature flag.

Not sure how I feel about this one.

chalk-solve/Cargo.toml
jackh726: Hmm. So what are the implications of this? We just always have to compile with both features?

chalk-engine-base/Cargo.toml
Mcat12: If it's part of `chalk-engine`, 90% of the crate will be under one feature flag.

jackh726: Yeah, exactly. Which is probably not a great tradeoff to just get rid of a small crate.

chalk-solve/Cargo.toml
Mcat12: I think it works as expected for downstream crates (with `recursive-solver`, `chalk-engine` is not compiled and vice versa). However, when running commands in this repo like `cargo test`/`cargo check`, both solvers will get built.

nikomatsakis:
So -- I tend to think that the engine boundary is useful conceptually, but what I don't think is all that useful is the fact that the engine is totally generic over all the types. I feel like one of the following two options makes sense:

* move chalk-engine into chalk-solve, but try to have a clear conceptual boundary between the modules, ideally still using a trait to guide "callbacks"
* make chalk-engine depend on chalk-ir and move the recursive solver into it

If we did the latter, there would still need to be a kind of "context" trait, but it would be much thinner, I imagine?

Conceptually, we could go further, and move the unification code and some other things from chalk-solve into chalk-ir as well, which would probably narrow the interface further. However, I *might* not want to do that, or at least I might want to wait until the sem-vs-syn split proceeds a bit, because I think I'd like it if the "semantic unification" remained in chalk-solve, as it is kind of "Rust-specific semantics" and feels like a better fit for a callback.

jackh726:
I think either of your ideas - moving chalk-engine into chalk-solve, or making chalk-engine depend on chalk-ir - would be good.

> ideally still using a trait to guide "callbacks"

I think this would be the same trait that we would get by having it in a separate crate

Moving unification and such code into chalk-ir I'm less inclined to do. I think we should strive to make chalk-ir as "type library"-like as possible, and less "trait solver library"-like. I *guess* unification can fall under types, but I'm not sure. 

I'm more or less thinking long-term in regards to rustc integration, what we might want things to look like. Specifically, right now chalk-ir is a dependency in rustc_middle. Later on, the shared type library will probably also be a dependency of rustc_middle.

Mcat12:
> I think I'd also like to see a directed graph (in a hackmd or in a book) that shows all the crates we intend to wind up with and summarizes their purpose. I think it'd be helpful.

I'm not sure if I have the context to do this, but maybe it wasn't directed at me :)

Mcat12:
The book lint check is failing because the new `chalk_solve::rust_ir` module documentation is not built yet. I'm not sure if this can be solved without changing how the book references the documentation (reference current branch's docs instead of published master branch docs in PRs?).

nikomatsakis:
@jackh726 

> I think this would be the same trait that we would get by having it in a separate crate

Yes -- and ideally it'd be the same trait for both solvers, too.

> I'm more or less thinking long-term in regards to rustc integration, what we might want things to look like. Specifically, right now chalk-ir is a dependency in rustc_middle. Later on, the shared type library will probably also be a dependency of rustc_middle.

I'm happy to wait, though I was thinking of the same thing. My main motivation was to try and make the trait smaller. But I think that if we have the solvers able to reference chalk-ir, the size of the trait will already go down considerably.

jackh726:
@Mcat12 how much do you want to try to experiment here? Would like to try to turn this into a bit of a refactor to make chalk-engine depend on chalk-ir?

Mcat12:
I'm ok with experimenting but I'm afraid it might make this PR too large (there are other nice changes in this PR like moving `chalk-rust-ir` and `chalk-macros`). Could we merge this and then experiment on further simplifying the SLG interface in a new PR?

jackh726:
I think the changes here are fine. @nikomatsakis do we want to merge this? Or wait for docs? Or wait for refactor?

Mcat12:
I think the docs are done, please feel free to point out if they are not completely updated.

jackh726:
I'm referring to Niko's suggestion of adding the directed graph

nikomatsakis:
I think we can wait on the directed graph until we know what we want

nikomatsakis:
I would be happy to edit in a hackmd to start

chalk-solve/Cargo.toml
pksunkara: Yup, had some experience with it. This is okay.

PR #461, opened 2020-05-20
Added scheduled release workflow

pksunkara:
As per the discussion on Zulip, here's the action that schedules and releases chalk every Sunday.

I used [cargo-workspaces](https://github.com/pksunkara/cargo-workspaces) for this which means if there are no changes in a week, a version will not be released. It also intelligently releases only the changes instead of everything.

You need to add 2 secrets in github for this to work:
* PERSONAL_ACCESS_TOKEN: Github personal access token that has rights to push to the master branch of this repo
* CARGO_REGISTRY_TOKEN: Crates.io token of the user who has access to publishing all the chalk crates

You can see an example release commit [here](https://github.com/pksunkara/chalk/commit/d8fdc2348d2a3abe1a987cf031a98e01a1442d8d)

cc @jackh726 @flodiebold @nikomatsakis 

I saw your message about asking me to also integrate bors-ng. While I did do that for https://github.com/clap-rs/clap, I am not sure if I can do that here without having access to settings. I could just create/edit the related config for it but it would be hard for me to debug.

**EDIT**: I realized that I would need to update `Cargo.lock` as well. Can someone confirm? While I would like to just run `cargo build` to update the `Cargo.lock`, but I think it would also update other deps, right? Is there an option to update only the local crates? I am not sure how to proceed on this.

Does running `cargo update -p {crate_name}` for each crate in workspace fix it?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,12 +16,16 @@ jobs:
         rust: [stable, nightly]
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
 
-      - name: Setup Rust ${{ matrix.rust }}
-        run: rustup set profile minimal && rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
 
       - name: Build chalk-engine without default features
         run: cd chalk-engine && cargo build --no-default-features
@@ -69,27 +73,28 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
 
-      - name: Setup Rust stable with rustfmt
-        run: |
-          rustup set profile minimal
-          rustup update stable --no-self-update
-          rustup default stable
-          rustup component add rustfmt
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          components: rustfmt
 
       - name: Check formatting of all crates in the workspace
         run: cargo fmt --all -- --check
-  
+
   mdbook-linkcheck:
     name: Book link check
     runs-on: ubuntu-latest
     if: github.ref != 'refs/heads/master'
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,43 @@
+name: Publish
+on:
+  schedule:
+    - cron: "0 0 * * 0" # midnight UTC on Sunday
+
+jobs:
+  publish:
+    name: Publish
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ssh-key: ${{ secrets.PUBLISH_DEPLOY_KEY }}
+          fetch-depth: 0
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Install cargo-workspaces
+        uses: actions-rs/install@v0.1
+        with:
+          crate: cargo-workspaces
+
+      - name: Release
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        shell: bash
+        run: |
+          git config --global user.name "Github Action"
+          cargo workspaces version -ay --force '*' --include-merged-tags --no-git-commit --exact patch
+          export VERSION=$(cargo pkgid | sed -E 's/.*#(.*)/\1/g')
+          sed -E -i '' $'s/(# Unreleased)/\\1\\\n\\\n# Release '"$VERSION/" RELEASES.md
+          git commit -am "Release $VERSION"
+          git tag "v$VERSION"
+          git push --tags
+          cargo publish --from-git
+          cargo workspaces version -ay --force '*' --include-merged-tags --no-git-tag --pre-id dev preminor
--- a/chalk-solve/Cargo.toml
+++ b/chalk-solve/Cargo.toml
@@ -21,7 +21,7 @@ chalk-engine = { version = "0.10.1-dev", path = "../chalk-engine", optional = tr
 chalk-ir = { version = "0.10.1-dev", path = "../chalk-ir", default-features = false }
 
 [dev-dependencies]
-chalk-integration = { version = "0.10.1-dev", path = "../chalk-integration" }
+chalk-integration = { path = "../chalk-integration" }
 
 [features]
 default = ["slg-solver", "recursive-solver"]


Mcat12:
Can this tool also bump the versions to a dev version after publishing?

Mcat12:
For updating `Cargo.lock`, you could do `cargo update -p chalk chalk-solve chalk-engine ...`. `cargo update -p chalk*` might work if the shell expands the glob.

pksunkara:
> Can this tool also bump the versions to a dev version after publishing?

I didn't think that is necessary. Is there any reason for doing this?

Mcat12:
I'm not sure if it's necessary, but I've observed it happening here.

pksunkara:
The tool now updates the `Cargo.lock`. [sample commit](https://github.com/pksunkara/chalk/commit/bca6246df5dddd94dcefef6e8ba9d41e7de627ec)

As you can see in that commit, only the changed packages got released. The next run would update to `0.13.0` if changes are added to a package that wasn't updated in the current run

jackh726:
This is really nice :) So, I know it doesn't update the version to add a `-dev` tag. I really like doing this, since it allows anyone to keep track of whether or not the code on master is released. (I've so often looked at a codebase and had to track down if the code I'm seeing is the a released version). If we can make it do this, I think I'm completely on board. (I might be able to submit a PR to cargo-workspaces if that what it takes).

Other than that, does it do everything on the [checklist](http://rust-lang.github.io/chalk/book/publishing.html)? If so, then I could maybe be persuaded that not adding a `-dev` tag is okay.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
jackh726, line 15: We'll have to figure out *who's* token this will be, and get it added by someone who has access to the repo settings. @nikomatsakis not sure if you do

chalk-solve/Cargo.toml
```diff
@@ -22,4 +22,4 @@ chalk-ir = { version = "0.10.1-dev", path = "../chalk-ir" }
 chalk-rust-ir = { version = "0.10.1-dev", path = "../chalk-rust-ir" }
 
 [dev-dependencies]
-chalk-integration = { version = "0.10.1-dev", path = "../chalk-integration" }
+chalk-integration = { path = "../chalk-integration" }
```
jackh726, line 25: Is that not needed because it's a dev dependency?

.github/workflows/ci.yml
```diff
@@ -16,12 +16,16 @@ jobs:
         rust: [stable, nightly]
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
pksunkara: That is what the github docs recommend. We should not try to live on the edge.

chalk-solve/Cargo.toml
pksunkara: Yes. And also the tool does not update this version because of that reason.

pksunkara:
> does it do everything on the checklist? 

It does everything except for the following things:

* add a `-dev` suffix
* editing RELEASES.md

> I really like doing this, since it allows anyone to keep track of whether or not the code on master is released. I've so often looked at a codebase and had to track down if the code I'm seeing is the a released version

Just doing `git log` should show you all the info you need since every release is tagged. Also, doing this suffix not only complicates the release process but also complicates the usage of these libraries using `git` in others by a little bit.

The work on cargo-workspaces needed for this is not much and I can do it, but i just don't understand the reason for it.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jackh726: Oh, this is the name of the action. Not the branch we're checking out. Whoops

chalk-solve/Cargo.toml
jackh726: Hmm. So I *was* versioning the non-published crates for consistency. But I guess it's okay if they don't get updated. What is going to be there "permanent version"?

(I don't know if we ever *want* to publish these. They're really just a fancy testing infrastructure)

jackh726:
So, the reasoning for a `RELEASES.md` is for users to have a general overview of the changes that went into a release. But we discussed it in a meeting, and I don't think anyone was opposed to *not* having one. And I don't think *anyone* wanted e.g. just a paste in of `git log`. So I'm mostly okay with not keeping it updated.

How does adding the suffix complicate the usage of the crates? If someone is using the git repo as a dependency, then they don't require a version. If they publish, then there *has* to be a crates.io release of the version they need. Which is *not* the same as the git master.

chalk-solve/Cargo.toml
pksunkara: The private crates do get versioned with the action here. It's just that version key in dev-dependencies for crates in workspace don't get updated. The reason behind this is that the version key is meaningless for dev deps if path exists.

Mcat12:
Why not update `RELEASES.md` in the PRs? That's a common and easy way to keep it up to date with the code changes.

pksunkara:
> If they publish, then there has to be a crates.io release of the version they need. Which is not the same as the git master.

Yeah, but imagine if the releases are non-breaking. For example, I would use crate `A` in `B` like the following:

```
A = { git = "git://github.com/rust/A", branch = "master", version = "1.2" }
```

I would want to develop my crate against the git master but since `A` follows semver I can safely use the `version` keyword which is used when I am publishing `B`. But if `A` starts adding `-dev` tags`, most of the time it's not an issue, but when the suffix is added to an incompatible version, that's when things start to get complicated.

But I understand that this basically comes down to the general preference of the maintainer and if you want it I will add it.

jackh726:
Well, I don't think cargo looks at version when pulling a git dependency. So it doesn't matter what the version is on master. (-dev or not). Also, while the *releases* follow semver, the master branch does not. So using master as a dependency but falling back to a version could theoretically break. (In fact, the only time we are guaranteed to not break by semver is if master would only be bumping to a patch version. E.g. if we are relying on new features in master)

If it's not a lot of work, it would be awesome if you could add it.

Thanks :)

pksunkara:
Did it. You can see the same run [here](https://github.com/pksunkara/chalk/runs/699797204?check_suite_focus=true) without the publishing ofcourse.

[This](https://github.com/pksunkara/chalk/commit/ef83cf37a0f84309b634c0d5bdc20b0e63c1ee4f) is the release commit generated by that run.
[This](https://github.com/pksunkara/chalk/commit/d8a54f8b7d0a27b5560f9e6a93709d38b4f0944e) is the release commit that adds the `dev` prerelease tag.

chalk-solve/Cargo.toml
jackh726: Oh cool

Mcat12:
> Why not update `RELEASES.md` in the PRs? That's a common and easy way to keep it up to date with the code changes.

https://github.com/rust-lang/chalk/pull/461#issuecomment-632176964
Not sure if this was seen, but I'm confused why we would stop updating `RELEASES.md`. For example, in the const-generics PR we could add a line saying "added const generics support" under the next minor version, simple as that.

jackh726:
@Mcat12 I did miss your comment, sorry.

We did discuss this in the meeting too. I don't think having people update the `RELEASES.md` in commits is a *bad* idea, but if this workflow doesn't update the version there too, then it's not quite as useful. The biggest argument against this, I think, is it just creates extra things to keep track of for  a (potentially) small gain.

pksunkara:
It's not a big deal to update it. Just a `sed` command should take care of it. But we haven't been using it thus I didn't do it.

jackh726:
Ooh then that would be ideal. (I do think it is a nice thing to have. Even if it's a little bit more work)

Mcat12:
I have previously looked into [`cargo-release`] for my own crates, and it supports running `sed`-like commands during the update: https://github.com/sunng87/cargo-release/blob/master/docs/faq.md#how-do-i-update-my-readme-or-other-files

Maybe something similar can be added with `cargo-workspaces`.

[`cargo-release`]: https://github.com/sunng87/cargo-release

pksunkara:
> Just a sed command should take care of it

I was wrong.

If we want to use this generally, we wouldn't know the version until `cargo-workspaces` would be run. But once it's run, we can't update the `RELEASES.md`. (Well, we could, but the updated RELEASES.md will not be present in the tagged commit). To correctly handle this, I would have to implement search and replace functionality in the tool like how `cargo-release` did which unfortunately is more work than I expected.

Mcat12:
We could stay one version ahead in the changelog. Then we just add a section for the next version when bumping post-release. Ex: after releasing 0.10, add a 0.11 section to the changelog.

pksunkara:
So, I had to add some manual steps with `git commit` & `git tag`, etc.. but I did it with `sed` now. You can see it [here](https://github.com/pksunkara/chalk/commit/b5d5157d898f65840b3ced3a8dfea130fd89754b#diff-5cafcf46bfb947f8e58fa5c16d39018c)

.github/workflows/publish.yml
nikomatsakis: I have access, and I'm not averse to adding a token, but I do wonder if we can use a deploy key.

pksunkara:
Master is failing due to recent merge of @Mcat12's PR

Mcat12:
It looks like the PR wasn't fully up to date with master and the panic tests which were in master reference the old `chalk-rust-ir` crate. I'm on mobile for the next few hours, otherwise I would make a PR to fix the panic test imports.

pksunkara:
We should definitely have bors

jackh726:
I fixed master and restarted the checks

.github/workflows/publish.yml
pietroalbini: Hmm, is there a reason why this is not using the builtin `${{ github.token }}`? Looking at the things you do in this job it seems to me that the builtin token has all the permissions required.

pksunkara: Pushes from builtin token doesn't trigger workflows or webhooks. Is that okay with us?

pietroalbini: That's right, I forgot about that limitation.

I'd still prefer not to use custom tokens, as there is no way to limit which repos they have access unless we create a dedicate account for this. We have a base64-encoded deploy key in the `GITHUB_DEPLOY_KEY` secret, maybe you could use it to push the commits and tags?

pksunkara: Can you describe a bit more about this `GITHUB_DEPLOY_KEY`? How is it created? I need to duplicate the process on my own repo so that I can test it.

pietroalbini: So, the contents of that variable is the base64-encoded copy of a private SSH key. You can generate a key with the same settings as the one configured in the repo with:

```
ssh-keygen -t ed25519 -a 100 -f key -q -N ""
```

Then you can get the contents of the variable with:

```
cat key | base64 -w 0
```

Finally you add the contents of `key.pub` in the "deploy keys" settings page of your repository.

pksunkara: Github doesn't allow us to add a secret with `GITHUB_` prefix. Can you recheck and verify it is the name this repo has?

Secondly, is there any reason for it to be base64 encoded? I would need to decode it first but then I am unable to pass it to `checkout` action's `ssh-key` field since it will contain newlines.

You can check the attempt [here](https://github.com/pksunkara/chalk/runs/706720851?check_suite_focus=true)

pietroalbini: > Github doesn't allow us to add a secret with GITHUB_ prefix. Can you recheck and verify it is the name this repo has?

It is, the restriction was added after CI in this repository was configured.

> Secondly, is there any reason for it to be base64 encoded? I would need to decode it first but then I am unable to pass it to checkout action's ssh-key field since it will contain newlines.

The key was base64-encoded because the tooling in the rust-lang/simpleinfra repo expects it that way. I created another key in the `PUBLISH_DEPLOY_KEY` variable, which is not base64-encoded.

By the way, setting an environment variable with `::set-env::` makes its content public, so your workflow would've leaked the key if it worked.

pksunkara: I can confirm that worked. Updated the PR.

> By the way, setting an environment variable with ::set-env:: makes its content public, so your workflow would've leaked the key if it worked.

Yeah, which was why I asking for the non base64 key. 😄

pksunkara:
@pietroalbini @nikomatsakis Only the [cargo token](https://github.com/rust-lang/chalk/pull/461/files#diff-963196a76268651f2135d3bbebba60faR32) is left to be added and this thing can start publishing.

nikomatsakis:
OK. I'm debating if I should make a "dummy user" to `nikomatsakis-chalk` to *just* host the publishing credentials here.

nikomatsakis:
Since crates.io tokens are kind of coarse-grained :)

pksunkara:
I have been trying to find some standardized info on the pre-release tags but am unable to. AFAIK, the pre-release tags should always be tagged as `word.num` where `num` starts with `0` and all of this can be followed by a build tag.

This has been my experience and the other release tools which support prereleases. But as I said, I can't find this written down anywhere.

jackh726:
@pksunkara See the "Backus–Naur Form Grammar for Valid SemVer Versions" section at https://semver.org/

Specifically, the `<version core> "-" <pre-release>`

pksunkara:
Let me rephrase, yes it is possible to write anything for the pre-release, but there is a convention majority people follow and I was describing it. I built the cargo-workspaces prerelease functionality around this convention. I don't think I can easily fix this.

jackh726:
I have always seen it as just `-dev`, or if it's alpha/beta maybe `alpha.0`. But we're only doing `-dev`. Why does `cargo-workspaces` auto-append a `.0` to a prerelease tag? Does that preclude have, say, multiple alphas?

pksunkara:
Yeah, the assumption behind it was that pre releases will always be numbered since they can be multiple with the same tag.

Mcat12:
The difference here seems to be that we never intend on publishing the `-dev` versions.

pietroalbini:
> OK. I'm debating if I should make a "dummy user" to nikomatsakis-chalk to just host the publishing credentials here.

That'd be the best way, even though I'd lean torwards a bot account managed by the infra team (I can set it up).

nikomatsakis:
@pietroalbini a bot sounds better, yes

nikomatsakis:
fyi, @pietroalbini was looking into creating an infra bot to host the token, but he ran into some problem that he's trying to resolve.

pietroalbini:
Added the `CARGO_REGISTRY_TOKEN` secret, but it does not work yet. I'll post here once it's ready to be used.

pksunkara:
What's the issue? Just curious.

pietroalbini:
The issue we had before was that the bot account we created got flagged, so we talked with GitHub to get it fixed. Now I need to coordinate with Niko to add ownership of the crates to the bot.

PR #462, opened 2020-05-20
Issue260

### LINKED ISSUE #260: Make chalk_solve::Solver UnwindSafe

hashedone:
The test is fixed, so it passes when panic is disabled, and then implementation is fixed - so table is not inserted, if it is not initialized correctly.

I tried to leave initial style, modifying only what is needed to be modified for this to work.

Also @jackh726 mentioned, that maybe after panic there should be two tests - one for goal which might be cached before, and the new one. I would say, that going this direction there could be even three test - one for goal cached before, one for goal on which there was a panic, and one for completely new goal. However I am not very much sure how much benefit would it give, as the fix is actually not to cache thing on panic, but I am ok to implement more tests - I am just worried that `MockDatabase` might grow to size, when it would be unreadable.

Fixes #260


chalk-engine/src/logic.rs
```diff
@@ -222,9 +222,9 @@ impl<C: Context> Forest<C> {
             goal
         );
         let coinductive_goal = context.is_coinductive(&goal);
-        let table = self.tables.insert(goal, coinductive_goal);
-        self.push_initial_strands(context, table);
-        table
+        let mut table = Table::new(goal.clone(), coinductive_goal);
+        Self::push_initial_strands(context, self.tables.next_index(), &mut table);
```
jackh726, line 226: The other way to do this would be to have `push_initial_strands` just be `get_initial_strands` and make `Table::new` take a `Vec` of `Strand`s?

Hmm, maybe not, since it also needs to be able to mark as floundered.

jackh726: `push_intial_strands_instantiated` could just return a `Table`? Then the `mut` is only in that function.

tests/integration/panic.rs
```diff
@@ -0,0 +1,196 @@
...
+
+#[derive(Debug, Default)]
+struct MockDatabase {
+    panic: bool,
+}
+
+#[allow(unused_variables)]
+impl RustIrDatabase<ChalkIr> for MockDatabase {
+    fn custom_clauses(&self) -> Vec<ProgramClause<ChalkIr>> {
+        if self.panic {
```
jackh726, line 15: So, I think ideally we want this impl to panic in each of the functions to *really* test for panic safety more thoroughly. Do you want to try to do this?

I imagine you could make `panic` be an `Option<enum>` with variants `CustomClauses`, `TraitDatum`, etc. Where if `panic` is that variant when the function is called, then that function panics. I think if the solver panic-safe enough to solve the test goal given here, that's a great start.

tests/integration/panic.rs
```diff
@@ -0,0 +1,196 @@
...
+fn unwind_safety() {
+    use self::MockDatabase;
+    use chalk_integration::interner::{self, ChalkIr};
+    use chalk_ir::*;
+    use std::panic;
+
+    // lower program
+    /*
+    let mut db = lower_program_with_db! {
+        program {
```
jackh726, line 149: Let's just make this program a comment by itself :) It's helpful to follow what we're actually writing.

chalk-engine/src/logic.rs
hashedone: I prefer to rename the function to somthing like `build_table` (I find `build_table_with_initial_strands` too long, but I would doc-comment it, that it just pushes table with initial strands). Obviously function would return finalized table - I just assumed that you have some reason I am not aware to go with it this way (like being more conformant with some book you are basing on, so it is easier to follow code reading documentation), that's way I was cautious with bigger changes around.

tests/integration/panic.rs
hashedone: I little more work, but I think I can do that.

The problem that I see is that we actually basing on internal implementation anyway - because most of functions are `unimplemented!()` (so they just panic). The problem is that to really test all the functions I would need some program and goal which actually call *every* single DB function, which I think can take me something around eternity with my current Chalk knowledge (ok, maybe slightly overestimated).

tests/integration/panic.rs
hashedone: I don't extly understand what do you mean - we should leave this program comment, or remove?

I am opting for leaving it, because I find it very helpful understanding what is the goal and things returned from DB about. I think I would move it the module documentation (as DB functions are actually related to it) and maybe describe goal we are proving here.

hashedone:
> Good work :) A couple comments.
> 
> I would be good with merging this mostly as-is now. And filing an issue for followup work to check more functions. (Maybe you want to work on it, maybe not). Or if you want to work a bit more on this (see above comment), we can leave this open as a WIP.

I moved it to draft so I will correct some simple things. I am not sure what about testing all DB functions for panic (I mean the functions which are unimplemented - see relevant comment). I can try it, but if it is this difficult I assume it is, then I would like some comment if it is worth - maybe it is better to help with "constants or maybe assoc constants?" mentioned by nikomatsakis on design meeting? TBH - anything is ok for me.

chalk-engine/src/logic.rs
jackh726: `build_table` works :)

In general, some *ideas* of SLG are based on the paper. But the code itself it's a "new" implementation. (In fact, since I've started with Chalk, I've refactored quite a bit of this code)

tests/integration/panic.rs
jackh726: Haha yeah.

I'm thinking. It might be worth going through and figuring out *when* in the `chalk-engine` lifecycle the db methods are called. I imagine *most* of them are called when building the table. (In which case it probably isn't helpful to add extra tests for each method)

tests/integration/panic.rs
jackh726: Just add a comment:
```
// goal we are trying to solve
// ...
```

So basically, just remove the extra bits around the program

chalk-engine/src/logic.rs
```diff
@@ -238,23 +236,35 @@ impl<C: Context> Forest<C> {
...
+            goal,
+        )
     }
 
     fn push_initial_strands_instantiated(
-        &mut self,
         context: &impl ContextOps<C>,
-        table: TableIndex,
+        table_idx: TableIndex,
+        mut table: Table<C>,
```
hashedone, line 262: I am not sure, if it wont be performance regression to take initialized table and then return it in result instead of in place mutate it (I find it is not trivial to implement such an optimization, and I don't know how rustc handles this), so I am debating on leaving it as `table: &mut Table<C>`, on the other hand I like it this way better... So I will just do what I would be told to do in this case, but here is my proposal.

tests/integration/panic.rs
hashedone: I was thinking on this a little. As you said - those methods are called mostly while building a table. This is however possible, that some method might be called somewhere else (maybe is now, or maybe will be in week, month, year, whatever). The problem is, that we would like to test the context on which method panicked, not the panicking method itself. So if method `impls_for_trait` is called for specific program 4 times, we would actually need to panic on it only on first call, only on second call, and so on. This would literally take an eternity, and I can do it for any timeframe you will tell me to do this (I mean I am able too, not willing to - I would like to help, not to die out of boredom).

What it comes to, that to really test all cases, we need to test all flows, and this is like at least big problem. My opinion is, that proper way to approach this is maybe create some fuzz tests, which are creating bigger and bigger programs and goals, and then try to solve the goal (unwinding panic), and then try to solve another goal on this program (also complicated, but such we already tested does not panic). This way we may find some actual flow which is a candidate to be improved.

However proposed solution is not easy, and I am not raising my hand to do this - I don't think how to actually approach such kind of testing.

tests/integration/panic.rs
hashedone: Yea, I reorganized comments a bit so they are more reasonable I think.

chalk-engine/src/logic.rs
```diff
@@ -238,23 +236,35 @@ impl<C: Context> Forest<C> {
...
     /// as possible.
-    fn push_initial_strands(&mut self, context: &impl ContextOps<C>, table: TableIndex) {
+    fn build_table(
+        context: &impl ContextOps<C>,
+        table_idx: TableIndex,
+        goal: C::UCanonicalGoalInEnvironment,
+    ) -> Table<C> {
         // Instantiate the table goal with fresh inference variables.
-        let table_goal = self.tables[table].table_goal.clone();
+        let table = Table::<C>::new(goal.clone(), context.is_coinductive(&goal));
```
jackh726, line 245: You can just move this to `push_intial_strands_instantiated` (or better, just move that function *into* this one. We don't really need the distinction when this function is 4 lines long.

hashedone:
I don't exactly understand why I still see "Changes requested" as I think I changed everything which we talked about (except eternity-estimated tests ;) ), and resolved all conversations - TBH this is my second OpenSource contribution, and preview one was accepted without any comments, so there may be something I don't understand about GitHub review tool...

jackh726:
I think github requests a re-review of approve before changes requested is changed

tests/integration/panic.rs
```diff
@@ -0,0 +1,341 @@
...
+        Self::NoPanic
+    }
+}
+
+#[derive(Debug, Default)]
+struct MockDatabase {
+    panicking_method: PanickingMethod,
+}
+
+/// This DB is representint lowered program:
```
jackh726, line 29: "This DB represents the following lowered program:"

tests/integration/panic.rs
```diff
@@ -0,0 +1,341 @@
...
+///
+/// struct Foo { }
+/// trait Bar { }
+/// impl Bar for Foo { }
+// FIXME: Check if items returned from functions of this struct can be simplified
```
jackh726, line 34: Apparently this didn't get saved before I submitted the review:

I would move this to the top of the file above `PanickingMethod` and save "FIXME: some of these are probably redundant, so we should figure out which panic in the same place in `chalk-engine`"

hashedone:
Worst things that recurring review happened to me on my daily job before, so don't worry ;)

PR #463, opened 2020-05-20
Handle bound vars in builtin trait impls

flodiebold:
The builtin impls have the same problem that the dyn trait impl did, which is
that they reuse the trait ref as given. The problem with that is that the
recursive solver calls `program_clauses` with trait refs that contain bound vars
from the original goal. So the builtin impls returned clauses with free bound
vars, which is a problem. The solution is the same as for the dyn trait impls:
Use the `Generalize` folder to collect those free bound vars and basically put a
`forall` around them. To give an example:

Looking for

```
Implemented((^0.0, ^0.1): Sized)
```

previously returned something like

```
if (Implemented(^0.1: Sized)) { Implemented((^0.0, ^0.1): Sized) }
```

as a clause; now it returns

```
forall<type, type> if (Implemented(^0.1: Sized)) { Implemented((^0.0, ^0.1): Sized) }
```

(The SLG solver instantiates goals before asking for program clauses, so it has
inference variables instead of bound vars in those places.)

### DIFF
--- a/tests/test/tuples.rs
+++ b/tests/test/tuples.rs
@@ -82,6 +82,30 @@ fn tuples_are_sized() {
         } yields {
             "Unique; substitution [], lifetime constraints []"
         }
+
+        goal {
+            exists<T> { (T, u8): Sized }
+        } yields {
+            "Unique; for<?U0> { substitution [?0 := ^0.0], lifetime constraints [] }"
+        }
+
+        goal {
+            forall<T> { (T, u8): Sized }
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
+        }
+
+        goal {
+            forall<T> { (u8, T): Sized }
+        } yields {
+            "No possible solution"
+        }
+
+        goal {
+            forall<T> { if (T: Sized) { (u8, T): Sized } }
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
+        }
     }
 }
 
@@ -89,6 +113,10 @@ fn tuples_are_sized() {
 fn tuples_are_copy() {
     test! {
         program {
+            // FIXME: If we don't declare Copy non-enumerable, `exists<T> { T:
+            // Copy }` gives wrong results, because it doesn't consider the
+            // built-in impls.
+            #[non_enumerable]
             #[lang(copy)]
             trait Copy { }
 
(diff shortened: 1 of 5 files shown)


tests/test/tuples.rs
```diff
@@ -82,13 +82,41 @@ fn tuples_are_sized() {
...
     }
 }
 
 #[test]
 fn tuples_are_copy() {
     test! {
         program {
+            // FIXME: If we don't declare Copy non-enumerable, `exists<T> { T:
+            // Copy }` gives wrong results, because it doesn't consider the
+            // built-in impls.
```
jackh726, line 118: Is this a previous issue? Can we add a test for this? What exactly do you mean by "wrong results"? (`No solution`?)

tests/test/tuples.rs
```diff
@@ -82,13 +82,41 @@ fn tuples_are_sized() {
...
+        }
+
+        goal {
+            forall<T> { (T, u8): Sized }
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
+        }
+
+        goal {
+            forall<T> { (u8, T): Sized }
```
jackh726, line 99: If we instead made `Sized` `non_enumerable`, this would give `Ambiguous`, right?

tests/test/tuples.rs
flodiebold: In this case, it returns `u8` as unique solution because that's the only explicit impl for `Copy`. The builtin code simply returns no clauses if the self type is a bound var, instead of returning all clauses or floundering.

tests/test/tuples.rs
flodiebold: No, it's still no possible solution (since this is a `forall`, I think).

flodiebold:
It might be slightly more work, but it's just a no-op fold instead of a clone, which I would think should not be much slower.

tests/test/tuples.rs
flodiebold: I've added a test.

flodiebold:
I'm not a fan of doing this for all program clauses because most program clauses don't actually need it. It's only when we take full types from the goal that it becomes necessary. It would actually be possible to write the builtin clauses code in a way that doesn't require this either (instead of a clause for e.g. the specific tuple type always generate a clause for all tuples of the given cardinality), this is just a bit easier.

jackh726:
I'm convinced.

PR #464, opened 2020-05-20
Opaque Type Generics

nathanwhit:
This PR picks up where #402 left off, and the failing test case from that PR now passes.

### DIFF
--- a/tests/test/opaque_types.rs
+++ b/tests/test/opaque_types.rs
@@ -45,3 +45,56 @@ fn opaque_reveal() {
         }
     }
 }
+
+#[test]
+fn opaque_generics_simple() {
+    test! {
+        program {
+            trait Iterator { type Item; }
+
+            struct Vec<T> { }
+            struct Bar { }
+            impl<T> Iterator for Vec<T> {
+                type Item = u32;
+            }
+
+            opaque type Foo<X>: Iterator = Vec<X>;
+        }
+
+        goal {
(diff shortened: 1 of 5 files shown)


tests/test/opaque_types.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        goal {
+            Foo<Bar>: Iterator<Item = Bar>
+        } yields {
+            "Unique; substitution []"
+        }
+
+        goal {
+            forall<T> {
+                if (T: Iterator) {
```
flodiebold, line 96: why the `T: Iterator` bound here?

detrumi: Right, this bound can be assumed, because we're assuming here that `Foo<X>` is well-formed.
Actually checking that `T` implements `Iterator` should only be done when you actually use `Foo`, at that point the WF-check (which isn't implemented yet for impl trait in chalk) checks the `T: Iterator` bound.

flodiebold: Well, not even that. `T` is the type parameter, there's no `Iterator` bound or reason why it would need to be an `Iterator` anywhere (`Foo<T>` is the `Iterator`).

nathanwhit: Whoops you're right, that bound is totally unnecessary because `T` is the `Item`, not the `Iterator`.

PR #465, opened 2020-05-21
Fix broken link

Y-Nak:
Fix some broken internal links.

### DIFF
--- a/book/src/what_is_chalk.md
+++ b/book/src/what_is_chalk.md
@@ -41,8 +41,8 @@ and impls, into *logical predicates*. This process is called "lowering", and you
 can learn more about in the [*Lowering to Logic*][lowering-to-logic] and
 [*Lowering Rules*][lowering-rules]) sections.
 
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
I had a look into this and don't believe `mdbook-linkcheck` is missing broken links here, it seems to be how browsers handle the `file://` protocol and absolute links.

When you open a document from the filesystem  (i.e. `mdbook build --open`) and click on a link to `/clauses.html` the browser will dutifully navigate to `file:///clauses.html` instead of the intended `file:///home/michael/Documents/forks/chalk/book/book/clauses.html` (or wherever the page happens to be). When served from a web server (i.e. `mdbook serve` or https://rust-lang.github.io/chalk/book/) this is all handled correctly.

Would it be worth adding some sort of "warn-on-absolute-links" check to `mdbook-linkcheck` to help people avoid this issue in the future?

jackh726:
A warning might be nice :)

PR #466, opened 2020-05-21
Add never type to `TypeName`

nathanwhit:
Addresses part of #368


nathanwhit:
> Copy/clone are in libcore: https://github.com/rust-lang/rust/blob/master/src/librustc_trait_selection/traits/select.rs#L2255
> 
> Also, looking over the table, looks like raw ptrs and immutable refs also have their impls in libcore. Think you can fix that here too? (It's small enough I don't think we _need_ to split into a separate PR)

Whoops you're right, should be fixed now!

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -74,9 +74,7 @@ pub fn add_sized_program_clauses<I: Interner>(
             TypeName::Tuple(arity) => {
                 push_tuple_sized_conditions(db, builder, trait_ref, *arity, substitution)
             }
-            TypeName::Scalar(_) | TypeName::Raw(_) | TypeName::Ref(_) => {
```
jackh726, line 77: These should stay. Only copy/clone is provided.

book/src/clauses/well_known_traits.md
```diff
@@ -34,6 +34,7 @@ Some common examples of auto traits are `Send` and `Sync`.
 | structs         | ⚬    | ⚬    |  ✅    | ✅     | ⚬    | ⚬  |  ⚬      |  ⚬       |   ✅         |
 | scalar types    | 📚    | 📚    | ✅     | ⚬     | ⚬   |  ⚬  |  ⚬     |  ⚬       |    ❌        |
 | str             | 📚    | 📚    | ⚬     | ⚬     | ⚬   |  ⚬  |  ⚬     |  ⚬       |    ❌        |
+| never type      | ✅   |  ✅   |  ✅   |  ⚬    |  ⚬    | ⚬   |   ⚬    |  ⚬       |   ❌       |
 | trait objects   | ⚬    | ⚬    | ⚬     |  ✅    | ⚬    | ⚬   | ⚬      |  ⚬       |    ⚬        |
 | functions ptrs  | ✅    | ✅    | ✅     | ⚬     | ⚬    | ❌   | ⚬      |  ⚬       |    ❌         |
 | raw ptrs        | ✅   |  ✅  |   ✅   |  ⚬    |   ⚬   |  ⚬  |   ⚬    |   ⚬      |      ❌      |
```
jackh726, line 40: Can you update this table for copy/clone of raw ptrs and immutable refs?

tests/test/refs.rs
```diff
@@ -51,71 +51,7 @@ fn mut_refs_are_sized() {
...
-fn mut_refs_are_not_clone() {
-    test! {
-        program {
-            #[lang(clone)]
-            trait Clone { }
-        }
-
-        goal {
-            forall<'a, T> { not { &'a mut T: Clone } }
+            forall<'a, T> { &'a T: Sized }
```
jackh726, line 54: This should be `&'a mut T`

PR #467, opened 2020-05-21
Add array types to `TypeName`

nathanwhit:
Addresses part of #368.


chalk-parse/src/ast.rs
```diff
@@ -212,6 +212,10 @@ pub enum Ty {
     Slice {
         ty: Box<Ty>,
     },
+    Array {
+        ty: Box<Ty>,
+        len: Box<GenericArg>,
```
jackh726, line 217: Why is one here a `Ty` and the other ` GenericArg`?

chalk-solve/src/clauses/builtin_traits/copy.rs
```diff
@@ -41,9 +42,18 @@ pub fn add_copy_program_clauses<I: Interner>(
             TypeName::Tuple(arity) => {
                 push_tuple_copy_conditions(db, builder, trait_ref, *arity, substitution)
             }
-            TypeName::Raw(_) | TypeName::Ref(Mutability::Not) => {
+            TypeName::Scalar(_) | TypeName::Raw(_) | TypeName::Ref(Mutability::Not) => {
```
jackh726, line 45: `Scalar` shouldn't have been added?

chalk-parse/src/ast.rs
nathanwhit: Yeah I'm not super pleased with it as is.
Basically the idea was that the first parameter has to be a type, hence it's a `Ty`, but for the array length you probably want to be able to use either a literal const value *or* a const parameter (i.e. either a `GenericArg::Id` or `GenericArg::ConstValue` in the AST). It seemed like the easiest way to abstract over those two possibilities was just to make the length a `GenericArg`.

Any ideas for a clean solution?

jackh726: Hmm let's see what @Areredify thinks

chalk-solve/src/clauses/builtin_traits/copy.rs
nathanwhit: Did not mean to commit that, my bad.

book/src/clauses/well_known_traits.md
```diff
@@ -40,7 +40,7 @@ Some common examples of auto traits are `Send` and `Sync`.
 | immutable refs  | ✅   |  ✅  |   ✅   |  ⚬    |   ⚬   |  ⚬  |   ⚬    |   ⚬      |      ❌      |
 | mutable refs    | ⚬    |  ⚬   |   ✅   |  ⚬    |   ⚬   |  ⚬  |   ⚬    |   ⚬      |      ❌      |
 | slices          | ⚬     | ⚬    | ⚬     |   ✅    | ⚬   | ⚬   | ⚬      |  ⚬       |    ❌       |
-| arrays❌         | ❌     | ❌    | ❌     | ❌      | ⚬   | ⚬   | ⚬      |  ⚬       |    ❌        |
```
Areredify, line 43: `Unsize` pr is still pending, ✅ is not correct here

chalk-parse/src/ast.rs
Areredify: So, I fell like the "idiomatic" solution in the spirit of the current implementation would be a) adding `ConstWithoutId` which is (for now) just `ConstValue` b) adding `Const` = `ConstWithoutId` + `Id` c) change `GenericArg::ConstValue` to `GenericArg::ConstWithoutId`. Then array would be `struct Array { ty: Box<Ty>, len: Box<Const> }`.
I can't really think of a better way to handle this without a handwritten parser

book/src/clauses/well_known_traits.md
nathanwhit: Ah yes, that's a mistake. Then the other ✅'s in the column are also incorrect, no?

Areredify: I think it would be easier to just extend this pr so `Sized`, `Copy` and `Clone` cover arrays, it's fairly trivial

Areredify: (should've probably mentioned it in my comment :upside_down_face: )

nathanwhit: I might be misunderstanding, but the PR already handles the `Sized`, `Copy` and `Clone` clauses for arrays.

Areredify: Uhh, you are correct, sorry, I completely missed it. What other ✅ were you talking about then?

nathanwhit: No worries. I was talking about the existing ✅ in the "Unsize" column for the other types, for instance for tuples and structs.

Areredify: They are implemented, but not yet merged, so I took the liberty to add them prematurely when writing the book page :)

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -74,6 +74,15 @@ pub fn add_sized_program_clauses<I: Interner>(
...
             }
+            TypeName::Array => {
+                let interner = db.interner();
+                needs_impl_for_tys(
+                    db,
+                    builder,
+                    trait_ref,
+                    iter::once(substitution.at(interner, 0).assert_ty_ref(interner).clone()),
+                );
+            }
```
Areredify, line 85: Arrays are always `Sized`, proving `Sized` for the element type is a part of WF checks (which are not yet implemented (or designed, for that matter)), so here you should just `push_fact(trait_ref)`.

chalk-parse/src/ast.rs
nathanwhit: Yeah that's along the lines of what I was thinking. I added a commit with those changes.

chalk-integration/src/lowering.rs
```diff
@@ -1454,6 +1469,22 @@ impl LowerGenericArg for GenericArg {
     }
 }
```
Areredify, line 1470: I don't think ast needs `ConstWithoutId` struct. Parsing `ConstWithoutId` into `Const` (just like `TyWithoutId` is parsed into `Ty`) seems cleaner implementation-wise.

chalk-integration/src/lowering.rs
```diff
@@ -1431,18 +1443,21 @@ impl LowerTy for Ty {
...
-trait LowerGenericArg {
-    fn lower(&self, env: &Env) -> LowerResult<chalk_ir::GenericArg<ChalkIr>>;
+impl LowerGenericArg for Const {
+    fn lower(&self, env: &Env) -> LowerResult<chalk_ir::GenericArg<ChalkIr>> {
+        let interner = env.interner();
+        match self {
+            Const::Id(name) => env.lookup_generic_arg(name),
+            Const::WithoutId(const_without_id) => Ok(const_without_id.lower(env)?.cast(interner)),
+        }
+    }
```
Areredify, line 1453: Consts should be lowered to consts, not to generic args

chalk-parse/src/parser.lalrpop
```diff
@@ -279,11 +280,20 @@ Lifetime: Lifetime = {
...
+    ConstWithoutId => Const::WithoutId(<>),
+};
+
 GenericArg: GenericArg = {
     TyWithoutId => GenericArg::Ty(<>),
     Lifetime => GenericArg::Lifetime(<>),
     Id => GenericArg::Id(<>),
-    ConstValue => GenericArg::ConstValue(<>),
+    ConstWithoutId => GenericArg::ConstWithoutId(<>),
 };
```
Areredify, line 297: This should be something like
```
ConstWithoutId: Const = {
    ConstValue => Const::ConstValue(<>),
};

Const : Const = {
    Id => Const::Id(<>),
    ConstWithoutId,
};

GenericArg: GenericArg = {
    TyWithoutId => GenericArg::Ty(<>),
    Lifetime => GenericArg::Lifetime(<>),
    Id => GenericArg::Id(<>),
    ConstWithoutId => GenericArg::Const(<>),
};
```

chalk-parse/src/ast.rs
```diff
@@ -109,9 +109,20 @@ pub enum GenericArg {
     Ty(Ty),
     Lifetime(Lifetime),
     Id(Identifier),
+    ConstWithoutId(ConstWithoutId),
```
Areredify, line 112: => `Const(Const)`

chalk-integration/src/lowering.rs
Areredify: Oops, I missed the right spot, I meant to comment on `impl LowerGenericArg for ConstWithoutId `

nathanwhit:
Updated to apply your suggestions @Areredify, and I also added some more lowering tests as suggested by @jackh726 for good measure.

chalk-integration/src/lowering.rs
```diff
@@ -1448,6 +1460,36 @@ impl LowerTy for Ty {
     }
 }
 
+trait LowerConst {
+    fn lower(&self, env: &Env) -> LowerResult<chalk_ir::Const<ChalkIr>>;
+}
+
+impl LowerConst for Const {
+    fn lower(&self, env: &Env) -> LowerResult<chalk_ir::Const<ChalkIr>> {
```
nikomatsakis, line 1468: I'm surprised we didn't already have something like this code?

chalk-integration/src/lowering.rs
```diff
@@ -1459,14 +1501,7 @@ impl LowerGenericArg for GenericArg {
             GenericArg::Ty(ref t) => Ok(t.lower(env)?.cast(interner)),
             GenericArg::Lifetime(ref l) => Ok(l.lower(env)?.cast(interner)),
             GenericArg::Id(name) => env.lookup_generic_arg(&name),
-            GenericArg::ConstValue(value) => Ok(chalk_ir::ConstData {
```
nikomatsakis, line 1462: Oh, hmm, I guess this is what we had before..? OK.

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -74,9 +74,11 @@ pub fn add_sized_program_clauses<I: Interner>(
             TypeName::Tuple(arity) => {
                 push_tuple_sized_conditions(db, builder, trait_ref, *arity, substitution)
             }
-            TypeName::Never | TypeName::Scalar(_) | TypeName::Raw(_) | TypeName::Ref(_) => {
-                builder.push_fact(trait_ref.clone())
-            }
+            TypeName::Array
```
nikomatsakis, line 77: Do we have "WF" condition code somewhere? We'll have to check that the `[T;N]` requires `T: Sized`

tests/test/arrays.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[test]
+fn arrays_are_well_formed() {
+    test! {
+        program { }
+
+        goal {
+            forall<const N, T> {
+                WellFormed([T; N])
```
nikomatsakis, line 111: This actually shouldn't hold, I suspect, we don't know that `T: Sized`...

chalk-integration/src/lowering.rs
Areredify: It wasn't needed before, because we only had consts in generic arg positions and parser didn't have a notion of `Const`; it was flattened into `GenericArg`.

chalk-solve/src/clauses/builtin_traits/sized.rs
Areredify: We push WF clauses in `match_type_name`, and I don't think we had a big discussion about how we want to handle well-formedness of tuples and arrays. IIRC the takeaway was that we rely on rustc to provide well-formed tuples (and, in this case, arrays), but I might be wrong

nikomatsakis: I guess I was thinking of [`WfSolver`](http://rust-lang.github.io/chalk/chalk_solve/wf/struct.WfSolver.html) -- but yeah I think we probably shouldn't block on this right now, it's something we'll definitely want to think about.

nikomatsakis: We should just file an issue I guess.

PR #468, opened 2020-05-22
Generate auto trait clauses for opaque type goals

detrumi:
When you have `opaque type Foo: Trait = Bar`,
solving the goal `Foo: Send`
will generate `Foo: Send :- Bar: Send`

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -104,6 +104,69 @@ pub fn push_auto_trait_impls<I: Interner>(
     });
 }
 
+/// Leak auto traits for opaque types, just like `push_auto_trait_impls` does for structs.
+///
+/// For example, given the following program:
+///
+/// ```notrust
+/// #[auto] trait Send { }
+/// trait Trait { }
+/// struct Bar { }
+/// opaque type Foo: Trait = Bar
+/// ```
+/// Checking the goal `Foo: Send` would generate the following:
+///
+/// ```notrust
+/// Foo: Send :- Bar: Send
+/// ```
+pub fn push_auto_trait_impls_opaque<I: Interner>(
+    builder: &mut ClauseBuilder<'_, I>,
+    auto_trait_id: TraitId<I>,
+    opaque_id: OpaqueTyId<I>,
+) {
+    debug_heading!(
+        "push_auto_trait_impls_opaque({:?}, {:?})",
+        auto_trait_id,
+        opaque_id
+    );
+
+    let opaque_ty_datum = &builder.db.opaque_ty_data(opaque_id);
+    let interner = builder.interner();
+
+    // Must be an auto trait.
+    assert!(builder.db.trait_datum(auto_trait_id).is_auto_trait());
+
+    // Auto traits never have generic parameters of their own (apart from `Self`).
+    assert_eq!(
+        builder.db.trait_datum(auto_trait_id).binders.len(interner),
+        1
+    );
+
+    let binders = opaque_ty_datum.bound.map_ref(|b| &b.hidden_ty);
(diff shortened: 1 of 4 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -161,7 +224,10 @@ fn program_clauses_that_could_match<I: Interner>(
 
             if trait_datum.is_non_enumerable_trait() || trait_datum.is_auto_trait() {
                 let self_ty = trait_ref.self_type_parameter(interner);
-                if self_ty.bound_var(interner).is_some()
+
+                if let TyData::Alias(AliasTy::Opaque(opaque_ty)) = self_ty.data(interner) {
+                    push_auto_trait_impls_opaque(builder, trait_id, opaque_ty.opaque_ty_id)
```
detrumi, line 229: I was a bit surprised that this needed to be here instead of next to the [`push_auto_trait_impls` call site](https://github.com/rust-lang/chalk/blob/8a3e89f5bc5aff02bb8b037cfab71d60dd9b0840/chalk-solve/src/clauses.rs#L257), so maybe we need to add `!Foo` syntax for goals in the ast? But that sounds wrong, since we have `ApplyTypeLookup::Opaque` for that.

chalk-solve/src/clauses.rs
```diff
@@ -104,6 +104,69 @@ pub fn push_auto_trait_impls<I: Interner>(
...
+    auto_trait_id: TraitId<I>,
+    opaque_id: OpaqueTyId<I>,
+) {
+    debug_heading!(
+        "push_auto_trait_impls_opaque({:?}, {:?})",
+        auto_trait_id,
+        opaque_id
+    );
+
+    let opaque_ty_datum = &builder.db.opaque_ty_data(opaque_id);
```
nikomatsakis, line 133: Reminder that we should factor the "hidden type" out from this struct and make it something that we lazilly request

chalk-solve/src/clauses.rs
nikomatsakis: Are you saying that you expected to be looking for a `TyData::Apply` with type name `TypeName::OpaqueType`? If so, I think I agree, but that's probably blocked on https://github.com/rust-lang/chalk/pull/401. Still, you could probably add this logic in various places that would be better. For one thing, this code runs also for non-auto-traits here, which will lead to assertion failures.

You might add it here:

https://github.com/rust-lang/chalk/blob/8a3e89f5bc5aff02bb8b037cfab71d60dd9b0840/chalk-solve/src/clauses.rs#L288

e.g., `if trait_datum.is_auto_trait() { push_auto_trait_impls_opaque(...); }`

chalk-solve/src/clauses.rs
detrumi: How would that work, actually? A method on the database to request it, or something similar?

flodiebold: Yeah, just a separate method on the db that only gets invoked when auto trait impls are actually required (that's important) would be fine, I think

chalk-solve/src/clauses.rs
```diff
@@ -161,7 +224,12 @@ fn program_clauses_that_could_match<I: Interner>(
 
             if trait_datum.is_non_enumerable_trait() || trait_datum.is_auto_trait() {
                 let self_ty = trait_ref.self_type_parameter(interner);
-                if self_ty.bound_var(interner).is_some()
+
+                if let TyData::Alias(AliasTy::Opaque(opaque_ty)) = self_ty.data(interner) {
```
nikomatsakis, line 228: I *think* this probably wants to be the placeholder, and not the alias type, but we can leave it as "future work" for now

PR #469, opened 2020-05-22
Check for trivial cycles in persue_answer

### LINKED ISSUE #313: improve detection of trivial self-cycles for coinductive goals
In #272, @jackh726 [commented](https://github.com/rust-lang/chalk/pull/272#discussion_r360658179):

Ok, so there's a couple things going on here, and I'll try to explain it succinctly since I'm traveling today (so on mobile at least for the minute).

So first, it's maybe not obvious from the PR (and maybe not even *wrong*), but something I found when looking through `CHALK_DEBUG` output. But in `apply_answer_subst`, we can't *properly* check if the delayed subgoals are equal to the table goal, because the delayed subgoals get unique inference variables. (Perhaps this would be easier/quicker to check using the variant3 test and adding dbg statements there.)

Second, I would REALLY prefer to have these be checked *before* they are added to the answer. But, because of 1,  we have to wait until they are both canonicalized, which happens in pursue_answer. At that point it is *too late* because that doesn't know about the subgoal. That was mostly the point of this FIXME: we can be smarter here I think, but it requires a bit of refactoring.

Third, there are essentially two places that do this check. First is in `apply_answer_subst` and second is here. They aren't the same: removing this causes the variant3 test to fail. Removing the other causes...many of the conductive tests to fail (and a never-ending test for one that I don't remember off the top of my head). I'm not convinced that there couldn't be just one check in the right place.

So, overall. I'm a bit skeptical. Between both checks, everything works. But as I said, I don't know if it's perfectly correct and I definitely think there's room for improvement. Whether that needs to be done in this PR or can be done separately idk. 

Nonetheless, I should have some time soon to, at the very least, cleanup this FIXME and a couple other things. (Like the extra apply_binders function that was added but isn't used anymore). If we decide to land the PR as is, I'll file a followup issue with more detailed info.

_Originally posted by @jackh726 in https://github.com/rust-lang/chalk/pull/272_

This issue has been assigned to @zaharidichev via [this comment](https://github.com/rust-lang/chalk/issues/313#issuecomment-629695303).

zaharidichev:
This change ensures that we do not add trivial coinductive goals to the answer at all and filter them in `pursue_answer`. This allows us to have this check performed in a single place as @jackh726 was suggesting. 

Fix #313

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -909,14 +909,9 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
                         // we might have to build on it (see the
                         // Delayed Trivial Self Cycle, Variant 3
                         // example).
-                        let (_, _, _, table_goal) = self
-                            .context
-                            .instantiate_ucanonical_goal(&self.forest.tables[table].table_goal);
 
                         let answer = self.forest.answer(table, answer_index);
-                        if let Some(strand) =
-                            self.create_refinement_strand(table, answer, table_goal)
-                        {
+                        if let Some(strand) = self.create_refinement_strand(table, answer) {
                             self.forest.tables[table].enqueue_strand(strand);
                         }
 
@@ -964,7 +959,6 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
         &self,
         table: TableIndex,
         answer: &Answer<C>,
-        table_goal: C::Goal,
     ) -> Option<CanonicalStrand<C>> {
         // If there are no delayed subgoals, then there is no need for
         // a refinement strand.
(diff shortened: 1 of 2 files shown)


chalk-engine/src/logic.rs
```diff
@@ -1328,11 +1314,23 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
             return None;
         }
 
+        let table = self.stack.top().table;
```
jackh726, line 1317: This is already in scope at the beginning of the function.

chalk-engine/src/logic.rs
```diff
@@ -1328,11 +1314,23 @@ impl<'forest, C: Context + 'forest, CO: ContextOps<C> + 'forest> SolveState<'for
...
+        let table_goal = &self.forest.tables[table].table_goal;
+
+        let filtered_delayed_subgoals = delayed_subgoals
+            .into_iter()
+            .filter(|delayed_subgoal| {
+                let (canonicalized, _) =
+                    infer.fully_canonicalize_goal(self.context.interner(), delayed_subgoal);
```
jackh726, line 1324: Hmm, so each of these is getting canonicalized twice (here and in `canonicalize_answer_subst` below. I wonder if it's worth creating a new `AnswerSubst` below by filtering the delayed subgoals. Though it might be a little less clean. Maybe leave a FIXME here for a missed optimization.

zaharidichev: Do you think you can give an example. We want to create the `C::CanonicalAnswerSubst` without actually calling  `canonicalize_answer_subst` or ?

jackh726: No, we would just do the filtering using the canonicalized answer subst created below. But it would be less clean right now, since we would essentially have to have a way to extract out those delayed subgoals, compare, and add back. But that would have to be through the `Context` trait, since `chalk-engine` can't access them directly. We *might* have all the functions around already (I don't remember), but this is fine for now.

jackh726:
Potentially spurious error in CI, just gonna retry.

PR #470, opened 2020-05-23
Integer and float variable kinds

### LINKED ISSUE #327: Handling integer/float variables
Integer literals of unspecified type like e.g. `1` get assigned an 'integer type' variable in Rust, which is for most uses like a normal type variable, but we know it can only be one of the integer types. The same goes for float literals. Chalk mostly doesn't care about this, but it can lead us to a non-ambiguous solution because in some cases, only one solution actually is an integer type. The most notable example (I think) happens when indexing into an array or slice:
```rust
let x: &[u32];
let i = 1;
x[i]
```
`i` gets inferred to `usize` here because that's the only integer type for which `Index` is implemented in this case, as far as I understand. Chalk currently can't do this.

There was already some discussion about this in [a Zulip stream](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/Inference.20with.20integer.20variables) a while ago; I thought I'd open a ticket to keep track of the problem.

Mcat12:
Closes #327 

This adds support for integer and float variable kinds. Example usage:
```rust
test! {
    program {
        trait Foo {}
        struct Bar {}

        impl Foo for usize {}
        impl Foo for Bar {}
    }

    goal {
        exists<int N> {
            N: Foo
        }
    } yields {
        "Unique; substitution [?0 := Uint(Usize)]"
    }
}
```

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -179,7 +179,8 @@ impl<I: Interner> Debug for TyData<I> {
         match self {
             TyData::BoundVar(db) => write!(fmt, "{:?}", db),
             TyData::Dyn(clauses) => write!(fmt, "{:?}", clauses),
-            TyData::InferenceVar(var) => write!(fmt, "{:?}", var),
+            TyData::InferenceVar(var, TyKind::General) => write!(fmt, "{:?}", var),
+            TyData::InferenceVar(var, TyKind::Integer) => write!(fmt, "{:?}i", var),
             TyData::Apply(apply) => write!(fmt, "{:?}", apply),
             TyData::Alias(alias) => write!(fmt, "{:?}", alias),
             TyData::Placeholder(index) => write!(fmt, "{:?}", index),
@@ -276,7 +277,8 @@ impl<'a, I: Interner> Debug for VariableKindsInnerDebug<'a, I> {
                 write!(fmt, ", ")?;
             }
             match binder {
-                VariableKind::Ty => write!(fmt, "type")?,
+                VariableKind::Ty(TyKind::General) => write!(fmt, "type")?,
+                VariableKind::Ty(TyKind::Integer) => write!(fmt, "integer type")?,
                 VariableKind::Lifetime => write!(fmt, "lifetime")?,
                 VariableKind::Const(ty) => write!(fmt, "const: {:?}", ty)?,
             }
@@ -765,7 +767,8 @@ impl<I: Interner> Debug for GenericArgData<I> {
 impl<I: Interner> Debug for VariableKind<I> {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         match self {
-            VariableKind::Ty => write!(fmt, "type"),
+            VariableKind::Ty(TyKind::General) => write!(fmt, "type"),
+            VariableKind::Ty(TyKind::Integer) => write!(fmt, "integer type"),
             VariableKind::Lifetime => write!(fmt, "lifetime"),
             VariableKind::Const(ty) => write!(fmt, "const: {:?}", ty),
         }
@@ -776,7 +779,8 @@ impl<I: Interner, T: Debug> Debug for WithKind<I, T> {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         let value = self.skip_kind();
         match &self.kind {
-            VariableKind::Ty => write!(fmt, "{:?} with kind type", value),
+            VariableKind::Ty(TyKind::General) => write!(fmt, "{:?} with kind type", value),
+            VariableKind::Ty(TyKind::Integer) => write!(fmt, "{:?} with kind integer type", value),
             VariableKind::Lifetime => write!(fmt, "{:?} with kind lifetime", value),
             VariableKind::Const(ty) => write!(fmt, "{:?} with kind {:?}", value, ty),
         }
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -181,6 +181,7 @@ impl<I: Interner> Debug for TyData<I> {
             TyData::Dyn(clauses) => write!(fmt, "{:?}", clauses),
             TyData::InferenceVar(var, TyKind::General) => write!(fmt, "{:?}", var),
             TyData::InferenceVar(var, TyKind::Integer) => write!(fmt, "{:?}i", var),
+            TyData::InferenceVar(var, TyKind::Float) => write!(fmt, "{:?}f", var),
             TyData::Apply(apply) => write!(fmt, "{:?}", apply),
             TyData::Alias(alias) => write!(fmt, "{:?}", alias),
             TyData::Placeholder(index) => write!(fmt, "{:?}", index),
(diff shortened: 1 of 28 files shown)


chalk-ir/src/debug.rs
```diff
@@ -178,7 +178,9 @@ impl<I: Interner> Debug for TyData<I> {
         match self {
             TyData::BoundVar(db) => write!(fmt, "{:?}", db),
             TyData::Dyn(clauses) => write!(fmt, "{:?}", clauses),
-            TyData::InferenceVar(var) => write!(fmt, "{:?}", var),
+            TyData::InferenceVar(var, TyKind::General) => write!(fmt, "{:?}", var),
+            TyData::InferenceVar(var, TyKind::Integer) => write!(fmt, "{:?}i", var),
+            TyData::InferenceVar(var, TyKind::Float) => write!(fmt, "{:?}f", var),
```
Mcat12, line 183: Some new notation is added to distinguish integer/float inference variables from normal ones. Does this make sense, or do we not care about adding more notation for this?

nikomatsakis: I think rustc uses a very similar notation, except maybe we write `?0t` for the general case.

nikomatsakis: Well, `_#22t`, something like that :)

chalk-solve/src/infer/invert.rs
```diff
@@ -150,7 +150,7 @@ where
             .inverted_ty
             .entry(universe)
             .or_insert_with(|| table.new_variable(universe.ui))
-            .to_ty(self.interner())
+            .to_ty(self.interner(), TyKind::General)
```
nikomatsakis, line 153: maybe make `to_ty` assume `TyKind::General`

nikomatsakis: and add `to_ty_with_kind` or something?

chalk-solve/src/infer/unify.rs
```diff
@@ -112,16 +115,41 @@ impl<'t, I: Interner> Unifier<'t, I> {
                     .expect("unification of two unbound variables cannot fail"))
             }
 
-            // Unifying an inference variables with a non-inference variable.
-            (&TyData::InferenceVar(var), &TyData::Apply(_))
-            | (&TyData::InferenceVar(var), &TyData::Placeholder(_))
-            | (&TyData::InferenceVar(var), &TyData::Dyn(_))
-            | (&TyData::InferenceVar(var), &TyData::Function(_)) => self.unify_var_ty(var, b),
+            // Tried to unify mis-matching inference variables (not caught by prior match arm)
```
nikomatsakis, line 118: Unifying a "general" type variable with an "int/float" type variable should be ok. In rustc, I believe we just "bind" the general one to the int/float one.

tests/test/numerics.rs
```diff
@@ -0,0 +1,79 @@
...
+/// A more straightforward version of the `integer_index` test where the
+/// variable is on the impl side of the trait ref.
+#[test]
+fn integer_kind_trait() {
+    test! {
+        program {
+            trait Foo {}
+            struct Bar {}
+
+            impl Foo for usize {}
```
nikomatsakis, line 45: Can we add a test with `impl Foo for u32` and `impl Foo for i32` and show that we still get ambiguity?

nikomatsakis:
I guess I didn't expect to add an enum to `Ty` but rather to have 5 variants:

* Ty
* IntTy
* FloatTy
* Lifetime
* Const(Ty)

but I'm not sure which setup is nicer. This one may be more elegant.

Mcat12:
> I guess I didn't expect to add an enum to Ty but rather to have 5 variants

Adding the enum to `VariableKind::Ty` is effectively the same thing (now you have `Ty(General)`, `Ty(Integer)`, and `Ty(Float)`), but it's easier to group the "Ty-like" variants together (`Ty(_)`).

chalk-solve/src/infer/unify.rs
Mcat12: Without this change, the SLG solver passes the tests and the recursive solver fails the `T = N, N = usize` test (and possibly more). Perhaps this change isn't needed (just a fix for the recursive solver), or is the SLG solver just lucky with the order it chooses to unify the types?

After adding the change the recursive solver fails on `T = N, T = char` because individually `?0` can unify with both `?1i` and `char`. That is, the SLG solver unifies `?0` with `char` first, then `?0=char` with `?1i` (fails here). The recursive solver unifies `?0` with `?1i` and then `?0` with `char` (succeeds, because the `?1i` kind data was lost or not used).

I'm unsure of how to "bind" the general inference variable to the more specific integer/float variable, since `unify_var_var` is in `ena` and only knows about `InferenceVar`. I suspect this would fix the problem because then the recursive solver would try to unify `?0=?1i` with `char` and fail.

chalk-solve/src/infer/invert.rs
Mcat12: More often than not (except for the tests), the kind is available because `to_ty` is called as part of a transformation on the inference variable. Maybe instead I can add a `to_new_ty(interner)` method to signify that this will return a new, general ty variable?

chalk-solve/src/infer/unify.rs
Mcat12: I think I figured it out. I didn't think that `unify_var_value` with a `Ty::InferenceVar` as the value would work, but it does.

chalk-solve/src/infer/invert.rs
Mcat12: Hm, looking at it again, the kind is only available once. Maybe some refactoring removed the other usages. I'll make `to_ty` assume `TyKind::General`.

chalk-solve/src/infer/unify.rs
```diff
@@ -101,7 +101,10 @@ impl<'t, I: Interner> Unifier<'t, I> {
         match (a.data(interner), b.data(interner)) {
             // Unifying two inference variables: unify them in the underlying
             // ena table.
-            (&TyData::InferenceVar(var1), &TyData::InferenceVar(var2)) => {
+            (
+                &TyData::InferenceVar(var1, kind1),
+                &TyData::InferenceVar(var2, kind2),
+            ) if kind1 == kind2 => {
```
nikomatsakis, line 107: Nit: Maybe this would be nicer if we combined the three cases and moved the `if` into the arm?

I'm imagining

```rust
(
                &TyData::InferenceVar(var1, kind1),
                &TyData::InferenceVar(var2, kind2),
            ) => {
    if kind1 == kind2 {
        ...
    } else if kind1.is_general() || kind2.is_general() {
        ...
    } else {
        Err(...)
    }
}
```

tests/test/numerics.rs
```diff
@@ -0,0 +1,171 @@
...
+                T = N, N = usize
+            }
+        } yields {
+            "Unique; substitution [?0 := Uint(Usize), ?1 := Uint(Usize)], lifetime constraints []"
+        }
+
+        goal {
+            exists<T, float N> {
+                T = N, N = f32
+            }
```
nikomatsakis, line 141: let's add 

```
exists<int N, float F> { N = F }
```

and check it gives an error

PR #472, opened 2020-05-23
Make chalk-engine depend on chalk-ir

jackh726:
This actually cleans things up a lot.

There might be other things that can be cleaned up (for example, here I removed `HhGoal` since it basically mirrors `GoalData`.

c.c. @Mcat12 @nikomatsakis


chalk-engine/src/boring_impls.rs
```diff
@@ -0,0 +1,199 @@
...
+    {
+        match self {
+            Literal::Positive(goal) => goal.visit_with(visitor, outer_binder),
+            Literal::Negative(goal) => goal.visit_with(visitor, outer_binder),
+        }
+    }
+}
+
+// FIXME: I can't figure out macro visibility rules
+macro_rules! copy_fold {
```
jackh726, line 155: I could use some help here. Just doing `chalk_ir::folder::copy_fold!(TableIndex)` gives the error `macro copy_fold is private`

Mcat12: `#[macro_export]` is used to make macros public: https://doc.rust-lang.org/reference/macros-by-example.html#scoping-exporting-and-importing

jackh726: `copy_fold` does have `macro_export`: https://github.com/rust-lang/chalk/blob/master/chalk-ir/src/fold/boring_impls.rs#L220

nathanwhit: Try invoking `copy_fold` using `chalk_ir::copy_fold!` instead of `chalk_ir::folder::copy_fold`.

Basically using the `macro_export` attribute actually exports the macro at the root of the crate, rather than the module where it's defined. I'm not exactly *sure* why it works this way though, because it's rather unintuitive.

chalk-engine/src/simplify.rs
```diff
@@ -1,20 +1,22 @@
...
-        subst: C::Substitution,
-        initial_environment: C::Environment,
-        initial_hh_goal: HhGoal<C>,
-    ) -> Fallible<ExClause<C>> {
+    pub(super) fn simplify_goal(
+        context: &impl ContextOps<I, C>,
+        infer: &mut dyn InferenceTable<I, C>,
+        subst: Substitution<I>,
+        initial_environment: Environment<I>,
+        initial_hh_goal: Goal<I>,
```
nathanwhit, line 18:
```suggestion
        initial_goal: Goal<I>,
```

chalk-engine/src/simplify.rs
```diff
@@ -1,20 +1,22 @@
...
-use crate::hh::HhGoal;
 use crate::{ExClause, Literal, TimeStamp};
 use chalk_base::results::Fallible;
 
-impl<C: Context> Forest<C> {
+use chalk_ir::interner::Interner;
+use chalk_ir::{Environment, Goal, GoalData, InEnvironment, QuantifierKind, Substitution};
+
+impl<I: Interner, C: Context<I>> Forest<I, C> {
     /// Simplifies an HH goal into a series of positive domain goals
```
nathanwhit, line 10: s/HH goal/goal

chalk-engine/src/simplify.rs
```diff
@@ -29,51 +31,50 @@ impl<C: Context> Forest<C> {
         let mut pending_goals = vec![(initial_environment, initial_hh_goal)];
 
         while let Some((environment, hh_goal)) = pending_goals.pop() {
-            match hh_goal {
-                HhGoal::ForAll(subgoal) => {
+            match hh_goal.data(context.interner()) {
```
nathanwhit, line 34:
```suggestion
            match goal.data(context.interner()) {
```

chalk-solve/src/solve.rs
```diff
@@ -269,7 +271,15 @@ pub struct Solver<I: Interner>(SolverImpl<I>);
 
 enum SolverImpl<I: Interner> {
     #[cfg(feature = "slg-solver")]
-    Slg { forest: Box<Forest<SlgContext<I>>> },
+    Slg {
+        forest: Box<Forest<I, SlgContext<I>>>,
+        max_size: usize,
+        /// The expected number of answers for a solution.
+        /// Only really sseful for tests, since `make_solution`
```
nathanwhit, line 278:
```suggestion
        /// Only really useful for tests, since `make_solution`
```

chalk-engine/src/boring_impls.rs
jackh726: Thanks

chalk-engine/src/boring_impls.rs
```diff
@@ -0,0 +1,162 @@
+use crate::{ExClause, FlounderedSubgoal, Literal};
+use chalk_base::results::Fallible;
+use chalk_ir::fold::{Fold, Folder};
+use chalk_ir::interner::{HasInterner, Interner, TargetInterner};
+use chalk_ir::visit::{Visit, VisitResult, Visitor};
+use chalk_ir::{Canonical, ConstrainedSubst, Constraint, DebruijnIndex, InEnvironment};
+use std::fmt::Debug;
+
+impl<I: Interner, TI: TargetInterner<I>> Fold<I, TI> for ExClause<I> {
```
nikomatsakis, line 9: Can we not use the derives here?

jackh726: Fixed :)

jackh726:
@nikomatsakis addressed your comment

PR #475, opened 2020-05-25
Add built-in support for Fn* traits

nathanwhit:
This is an attempt at adding support for `Fn`, `FnMut`, and `FnOnce`. 

There isn't anything too interesting at the moment because closures are not implemented yet, but with this PR function pointers will now implement the `Fn` traits if the function pointer's argument types match the type parameters on the trait-ref.

I'm not very confident about the (lack of) well-formedness constraints, so some confirmation on that would be helpful.

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -1757,6 +1757,9 @@ impl LowerWellKnownTrait for WellKnownTrait {
             Self::CopyTrait => rust_ir::WellKnownTrait::CopyTrait,
             Self::CloneTrait => rust_ir::WellKnownTrait::CloneTrait,
             Self::DropTrait => rust_ir::WellKnownTrait::DropTrait,
+            Self::FnTrait => rust_ir::WellKnownTrait::FnTrait,
+            Self::FnMutTrait => rust_ir::WellKnownTrait::FnMutTrait,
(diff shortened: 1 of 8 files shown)


Mcat12:
Should it also handle return types?

jackh726:
Hmm, so a couple things
1) Definitely feel like there should be an associated `Output` type
2) What exactly is the purpose of these? Is there anything special about the `Fn` types other than that closures will auto impl them?

nathanwhit:
> Should it also handle return types?

It should indeed, I was starting just with handling function pointers which currently only store one parameter and don't (explicitly) model a return type.



> Hmm, so a couple things
> 
>     1. Definitely feel like there should be an associated `Output` type
> 
>     2. What exactly is the purpose of these? Is there anything special about the `Fn` types other than that closures will auto impl them?

Regarding 1, yes there definitely should be. I'm slightly embarrassed that I overlooked this.
Regarding 2, there isn't anything *too* special about the `Fn` traits as far as I know besides the special handling for closures and function pointers. We haven't modeled closures yet, though, so currently these traits aren't very useful.

Thanks to you both for the feedback!

All in all, I feel a bit silly as I now realize I jumped the gun on this PR. Right now I'd like to close this PR and do some refactoring on function pointers and function definitions first, then add support for closures, then revisit these traits in a future PR.

PR #476, opened 2020-05-26
Use chain and filter instead of extending a vec in program_clauses_fo…

zaharidichev:
Remove a minor FIXME in the code and resort to using chain + filter instead of extending a vec. 

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -127,13 +127,12 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
 
     let clauses: Vec<ProgramClause<I>> = custom_clauses
         .chain(clauses_that_could_match)
-        .filter(|c| c.could_match(interner, goal))
         .chain(
             db.program_clauses_for_env(environment)
                 .iter(interner)
-                .filter(|c| (*c).could_match(interner, goal))
                 .cloned(),
         )
+        .filter(|c| c.could_match(interner, goal))
         .collect();
 
     debug!("vec = {:#?}", clauses);
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -121,21 +121,24 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
     );
     let interner = db.interner();
 
-    // FIXME: change this to use `.chain().filter()`
-    let mut vec = vec![];
-    vec.extend(db.custom_clauses());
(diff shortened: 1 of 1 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -121,21 +121,24 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
...
+    let clauses: Vec<ProgramClause<I>> = custom_clauses
+        .chain(clauses_that_could_match)
+        .filter(|c| c.could_match(interner, goal))
+        .chain(
+            db.program_clauses_for_env(environment)
+                .iter(interner)
+                .filter(|c| (*c).could_match(interner, goal))
+                .cloned(),
+        )
+        .collect();
```
jackh726, line 137: This should just have a single `filter`.

chalk-solve/src/clauses.rs
```diff
@@ -121,21 +121,23 @@ pub(crate) fn program_clauses_for_goal<'db, I: Interner>(
...
+    let custom_clauses = db.custom_clauses().into_iter();
+    let clauses_that_could_match =
+        program_clauses_that_could_match(db, environment, goal).map(|cl| cl.into_iter())?;
+
+    let clauses: Vec<ProgramClause<I>> = custom_clauses
+        .chain(clauses_that_could_match)
+        .chain(
+            db.program_clauses_for_env(environment)
+                .iter(interner)
+                .cloned(),
```
Mcat12, line 133: We can avoid the cloning here if we don't abstract away the `Vec` via `InternedProgramClauses`. Why do we do this?

zaharidichev: Hm, not sure really. @jackh726 Any idea why this is the case ?

jackh726: Why do we do what exactly? Return `InternedProgramClauses` from `program_clauses_for_env`? Or just use `InternedProgramClauses` instead of `Vec`? Either way, we don't want `Vec`s, we want the implementor of `Interner` to decide what we use (maybe it's `SmallVec`s, maybe it's an interned slice). Cloning here is fine, since we also say that cloning a `ProgramClause` is fine.

zaharidichev: Got it, thanks !

Mcat12: Hm, maybe we could add an `IntoIterator<Item = InternedProgramClause>` bound on the type?

PR #477, opened 2020-05-27
Add function def clauses

jackh726:
Missed in https://github.com/rust-lang/chalk/pull/449

Also Sized test for `Never`


book/src/clauses/well_known_traits.md
```diff
@@ -36,6 +36,7 @@ Some common examples of auto traits are `Send` and `Sync`.
 | str             | 📚    | 📚    | ⚬     | ⚬     | ⚬   |  ⚬  |  ⚬     |  ⚬       |    ❌        |
 | never type      | 📚   |  📚   |  ✅   |  ⚬    |  ⚬    | ⚬   |   ⚬    |  ⚬       |   ❌       |
 | trait objects   | ⚬    | ⚬    | ⚬     |  ✅    | ⚬    | ⚬   | ⚬      |  ⚬       |    ⚬        |
+| functions defs  | ✅    | ✅    | ✅     | ⚬     | ⚬    |  ⚬  | ⚬      |  ⚬       |    ❌         |
```
jackh726, line 39: I actually don't know, should `FnDef`s impl `Fn`?

flodiebold: [Yes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c91518443d32a200da4f0a299bfe9b44).

tests/test/fn_def.rs
```diff
@@ -0,0 +1,66 @@
...
+            "Unique"
+        }
+    }
+}
+
+#[test]
+fn fn_def_is_copy() {
+    test! {
+        program {
+            #[lang(sized)]
```
bjorn3, line 38: This should be `copy`, not `sized`.

tests/test/fn_def.rs
```diff
@@ -0,0 +1,66 @@
...
+            "Unique"
+        }
+    }
+}
+
+#[test]
+fn fn_def_is_clone() {
+    test! {
+        program {
+            #[lang(sized)]
```
bjorn3, line 55: This should be `clone`, not `sized`.

jackh726:
Well shoot...I'll just include these in the chalk-base PR.

tests/test/fn_def.rs
```diff
@@ -0,0 +1,66 @@
+use super::*;
```
Mcat12, line 1: Another boo-boo, these tests aren't run because the module is not listed in `tests/test/mod.rs`

Mcat12: They fail in the parser because it expects `;` instead of `{ }`.

jackh726: 🙄

PR #479, opened 2020-05-28
Remove chalk-base

jackh726:


chalk-engine/src/lib.rs
```diff
@@ -285,3 +282,50 @@ chalk_ir::copy_fold!(TimeStamp);
 
 chalk_ir::const_visit!(TableIndex);
 chalk_ir::const_visit!(TimeStamp);
+
+#[macro_export]
+macro_rules! index_struct {
```
Mcat12, line 287: Good idea, moving to chalk-engine!

chalk-ir/src/lib.rs
```diff
@@ -1,13 +1,15 @@
 #![deny(rust_2018_idioms)]
 
+#[macro_use]
+extern crate lazy_static;
```
Mcat12, line 4: Can we directly import `lazy_static::lazy_static`, or is it still idiomatic to `macro_use` it (compared to when we directly import the debug macros)?

jackh726: Good catch! I'll change this.

jackh726:
Thanks @Mcat12 for the review. I also snuck in the two test fixes from my boo-boo in https://github.com/rust-lang/chalk/pull/477.

Once CI is green, I'll go ahead and merge.

PR #481, opened 2020-05-28
Model function ABI in the Rust IR

### LINKED ISSUE #460: Track function ABI information
This should (minimally) include:

- Calling convention (e.g. "C", "Rust")
- ABI of each argument
- ABI of the return type

In chalk, we should create a `FnAbi` (or equivalent) struct which holds the relevant information. [Here](https://github.com/rust-lang/rust/blob/9e2a6a29ce82e4fc5decad86dab7911a38582438/src/librustc_target/abi/call/mod.rs#L536) is where rustc models this information, which is a good starting point. Once this is modeled in chalk, the `FnAbi` should be added to both function definitions and function pointers.

Mcat12:
Closes #460

Only tracks function calling convention based on the discussion in the issue (though the interner could put more data in the `FnAbi` type).


chalk-integration/src/lowering.rs
```diff
@@ -1031,6 +1031,7 @@ impl LowerFnDefn for FnDefn {
 
         Ok(rust_ir::FnDefDatum {
             id: fn_def_id,
+            abi: ChalkFnAbi::Rust,
```
Mcat12, line 1034: Locally I have added a `#[fn_abi(Rust/C)]` annotation for functions to control the ABI, but I haven't committed it because I don't know if we can use it in the tests currently (I don't think you can equate function definitions?), or if it's even desired.

chalk-integration/src/interner.rs
```diff
@@ -28,6 +28,12 @@ impl Debug for RawId {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum ChalkFnAbi {
+    Rust,
+    C,
```
bjorn3, line 34: The `system` ABI is also stable. The standard library also uses `rust-intrinsic`, `rust-call`, `platform-intrinsic` and maybe some other ABI's. Maybe use a string instead to be able to represent all ABI's without changes to chalk.

flodiebold: This is just the testing harness, the actual ABI type is already an associated type that can be anything.

chalk-integration/src/lowering.rs
nikomatsakis: hmm, yeah, it seems like we need a way to test this, so I think you should add it, though I'd probably prefer 

```
extern "C" fn() { ... }
```

notation to `#[fn_abi(C)]`. You can handle that by parsing a string literal and comparing it against known values...

chalk-ir/src/lib.rs
```diff
@@ -747,6 +747,7 @@ impl InferenceVar {
 #[derive(Clone, PartialEq, Eq, Hash, HasInterner)]
 pub struct Fn<I: Interner> {
     pub num_binders: usize,
+    pub abi: I::FnAbi,
```
nikomatsakis, line 750: We need to be checking in type equality that the ABIs are the same. I guess that, given how chalk is currently setup, this would take place here

https://github.com/rust-lang/chalk/blob/4dba5a5c13c6a96375a889700c9bfc6a5f47b577/chalk-solve/src/infer/unify.rs#L146-L148

i.e., just before calling `unify_binders`, but I don't love that this slipped through. I think the *best* place for it to be checked would be in the `zip` code, i.e., there should be a `super_zip_tys` function or something that manages some of the grungy details, but I definitely think that's out of scope for this PR, so no need to consider anything like that.

We should add a test for unifying function types with distinct ABIs, basically testing that this goal gets an error

```
extern "C" fn() == extern "Rust" fn()
```

nikomatsakis:
One other thing, apparently a `extern "C" fn()` doesn't implement the `Fn` trait, so we have to fix that too -- this is kind of growing the scope of this PR a lot, I think if you want to back out the extensions to the `Fn` trait, we could move them instead into a follow-up PR (actually they would fit neatly into #507, which @mental32 was going to tackle)

Mcat12:
I kept the Fn ABI commit in the PR so the changes can be reused/referenced for future PRs.

PR #483, opened 2020-06-01
Get ready for the first publish

pksunkara:
Since the current tag is `v0.10.1-dev`, the action is trying to publish `v0.10.1` which we don't want. Luckily the action failed because of other reasons. So, we had the change to correct the first publish to be `v0.11.0`

https://github.com/rust-lang/chalk/releases/tag/v0.10.1 existed because of a failed publish earlier. So, the next publish had issues. Moving the pushing of tags to run after publishing makes sure the action is a bit more fault tolerant.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,7 +1,7 @@
 name: Publish
 on:
   schedule:
-    - cron: "0 0 * * 1" # midnight UTC on Monday
+    - cron: "0 0 * * 2" # midnight UTC on Tuesday
 
 jobs:
   publish:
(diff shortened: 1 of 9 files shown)


jackh726:
Can we just keep the Sunday or Monday release? And I'll just trigger this manually when I merge.

pksunkara:
I don't think you can manually trigger this.

I will update it to Sunday as soon as the first release happens and we know it works.

.github/workflows/publish.yml
```diff
@@ -38,6 +38,6 @@ jobs:
           sed -E -i '' $'s/(# Unreleased)/\\1\\\n\\\n# Release '"$VERSION/" RELEASES.md
           git commit -am "Release $VERSION"
           git tag "v$VERSION"
-          git push --tags
           cargo workspaces publish --from-git
+          git ps --tags
```
Mcat12, line 42: Is this supposed to be `git push --tags`?

pksunkara: @Mcat12 Good catch. https://github.com/rust-lang/chalk/pull/484

PR #486, opened 2020-06-02
Update publish.yml

pksunkara:
I released cargo-workspaces@0.2.4 which should handle the published like we talked yesterday. This PR is to test the first publish.


jackh726:
I'm gonna just change it publish ASAP. Thanks!

PR #487, opened 2020-06-03
Refactor the recursive solver a bit

jackh726:
This is about 95% of the work needed to put the recursive solver in a separate crate.

Everything in `chalk_solve::recursive::*` can get moved. `recursive.rs` is mostly the equivalent of the `slg` submodule.

### DIFF
--- a/chalk-engine/src/forest.rs
+++ b/chalk-engine/src/forest.rs
@@ -3,11 +3,10 @@ use crate::logic::RootSearchFail;
 use crate::table::AnswerIndex;
 use crate::tables::Tables;
 use crate::{TableIndex, TimeStamp};
-use std::fmt::Display;
 
 use chalk_ir::debug;
 use chalk_ir::interner::Interner;
-use chalk_ir::{Canonical, ConstrainedSubst, Goal, InEnvironment, Substitution, UCanonical};
+use chalk_ir::{Goal, InEnvironment, Substitution, UCanonical};
 
 pub struct Forest<I: Interner, C: Context<I>> {
     pub(crate) tables: Tables<I>,
@@ -39,7 +38,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
     /// iterator. Each time you invoke `next`, it will do the work to
     /// extract one more answer. These answers are cached in between
     /// invocations. Invoking `next` fewer times is preferable =)
-    fn iter_answers<'f>(
+    pub fn iter_answers<'f>(
         &'f mut self,
         context: &'f impl ContextOps<I, C>,
         goal: &UCanonical<InEnvironment<Goal<I>>>,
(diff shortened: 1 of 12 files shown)


chalk-engine/src/forest.rs
```diff
@@ -54,89 +53,6 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
             _context: std::marker::PhantomData::<C>,
         }
     }
-
-    /// Solves a given goal, producing the solution. This will do only
-    /// as much work towards `goal` as it has to (and that works is
-    /// cached for future attempts).
-    pub fn solve(
```
jackh726, line 61: As sort of a drive-by change, I moved these two functions `solve` and `solve_multiple` out of `Forest` (and `chalk-engine`) and into `chalk_solve::solve`. For `solve`, it honestly just makes more sense, since this is the *only* place solution is needed. For `solve_multiple`, there's a little work done here, but really I think these are implementation details akin to `aggregate`.

chalk-solve/src/recursive.rs
```diff
@@ -26,6 +33,104 @@ pub(crate) struct RecursiveContext<I: Interner> {
     caching_enabled: bool,
 }
 
+pub(crate) struct RecursiveInferenceTableImpl<I: Interner> {
+    pub(crate) infer: InferenceTable<I>,
+}
+
+impl<I: Interner> RecursiveInferenceTable<I> for RecursiveInferenceTableImpl<I> {
```
jackh726, line 40: I started to wonder if maybe it's worth having this notion of an `InferenceTable` (as a trait) in `chalk_ir`. A lot of these functions are also used for slg. Notable differences include `invert_then_canonicalize` and actually using the `free_vars` from `canonicalize`

chalk-solve/src/recursive/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    pub fn update_from(&mut self, minimums: Minimums) {
+        self.positive = ::std::cmp::min(self.positive, minimums.positive);
+    }
+}
+
+/// A (possible) solution for a proposed goal.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Solution<I: Interner> {
```
jackh726, line 30: This is mostly just copied from `chalk_solve::solve`. But I didn't really think of a good way to handle this. We can make `SearchGraph` generic over `Solution`, but `Fulfill:solve` needs to know about these. I wish it was easier to split out `solve` from `Fulfill` (similar to how we split out `aggregate` from `chalk-engine`, but the `solve` function needs to know a *lot* about the internals of `Fulfill` that we would have to expose publicly.

chalk-solve/src/solve.rs
```diff
@@ -57,100 +60,6 @@ impl<I: Interner> Solution<I> {
         }
     }
 
-    /// There are multiple candidate solutions, which may or may not agree on
```
jackh726, line 60: On the other hand (continued from where these are copied to), the fact that several of these functions are only used in the recursive solver makes me wonder if this is mostly okay anyways.

chalk-solve/src/solve.rs
```diff
@@ -316,10 +225,32 @@ impl<I: Interner> Solver<I> {
                 expected_answers,
             } => {
                 let ops = SlgContextOps::new(program, *max_size, *expected_answers);
-                forest.solve(&ops, goal, || true)
+                ops.make_solution(goal, forest.iter_answers(&ops, goal), || true)
             }
             #[cfg(feature = "recursive-solver")]
-            SolverImpl::Recursive(ctx) => ctx.solver(program).solve_root_goal(goal).ok(),
+            SolverImpl::Recursive(ctx) => {
+                ctx.solver(program)
```
jackh726, line 232: *But* (continued from above still), this is a bit annoying.

zaharidichev: Yes.. that is indeed a bit annoying. Nevertheless if what you suggested happens (chalk engine + chalk_recursive depending on chalk solve), this one solver struct can go away correct?

chalk-solve/src/recursive.rs
nikomatsakis: I think the inference code *itself* should maybe move to chalk-ir, actually, but we should discuss, especially in light of your idea to "invert" the dependency layers.

chalk-solve/src/recursive/lib.rs
nikomatsakis: I feel.. hmm.. a bit confused about this. I guess I think I would expect this type to potentially be in chalk-ir -- is it really *specific* to the recursive solver? Don't we use a similar type for SLG? (I guess as the *output* from aggregation?) 

I think my take is that aggregation probably belongs in with the SLG solver in the end, although it is definitely a distinct "layer".

PR #496, opened 2020-06-06
fix ucanonicalization bugs

Areredify:
fix some small bugs

### DIFF
--- a/chalk-solve/src/infer/ucanonicalize.rs
+++ b/chalk-solve/src/infer/ucanonicalize.rs
@@ -27,6 +27,10 @@ impl<I: Interner> InferenceTable<I> {
             DebruijnIndex::INNERMOST,
         );
 
+        for universe in value0.binders.iter(interner) {
+            universes.add(*universe.skip_kind())
+        }
+
         // Now re-map the universes found in value. We have to do this
         // in a second pass because it is only then that we know the
         // full set of universes found in the original value.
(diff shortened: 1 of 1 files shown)


shepmaster:
Is "ucanonicalization" a typo or a fancy compiler term?

bjorn3:
I think @Areredify means universe canonicalization.

Areredify:
@shepmaster  bjorn3 is right, ucanonicalization stands for universe canonicalization: basically, after goal canonicalization (that replaces bound inference variables with their values) we do universe canonicalization, so that if the same subgoal is encountered with slightly different universes, we can use cached result instead of recomputing it. Consider
```
goal {
   forall<T1> {
      forall<T2> {
          Implemented(S<T2>: Trait)
      }
   }
}


goal {
   forall<T1> {
      Implemented(S<T1>: Trait)
   }
}
```
These goals are almost the same, except when pursuing the `Implemented` subgoal, the first will look like `Implemented<!2_0>` and second will look like `Implemented<!1_0>`. We, of course, want to reuse the first goal when proving the second goal because they are almost the same. After ucanonicalizing they will both look like `Implemented<!1_0>` (or `Implemented<!0_0>`, don't remember which one is it), which allows for caching.

Areredify:
Uhh, sure

Areredify:
Hmm, I am no longer sure that this is a bug

PR #498, opened 2020-06-07
Where clauses on opaque types

detrumi:
I slightly prefer this:
```
opaque type T: Clone where T: Trait = Ty;
```
to this:
```
opaque type T: Clone = Ty where T: Trait;
```
since the where clause is on the bounds, not on the hidden type `Ty`

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -512,11 +512,11 @@ impl LowerProgram for Program {
                                         chalk_ir::VariableKind::Ty(TyKind::General),
                                         Atom::from(FIXME_SELF),
                                     )),
-                                    |env1| {
-                                        let interner = env1.interner();
+                                    |env| {
+                                        let interner = env.interner();
                                         Ok(opaque_ty
                                             .bounds
-                                            .lower(&env1)?
+                                            .lower(&env)?
                                             .iter()
                                             .flat_map(|qil| {
                                                 // Instantiate the bounds with the innermost bound variable, which represents Self, as the self type.
@@ -529,6 +529,7 @@ impl LowerProgram for Program {
                                                     .intern(interner),
                                                 )
                                             })
+                                            .chain(opaque_ty.where_clauses.lower(&env)?)
                                             .collect())
                                     },
                                 )?;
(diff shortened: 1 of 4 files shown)


chalk-integration/src/lowering.rs
```diff
@@ -529,6 +529,7 @@ impl LowerProgram for Program {
                                                     .intern(interner),
                                                 )
                                             })
+                                            .chain(opaque_ty.where_clauses.lower(&env)?)
```
nikomatsakis, line 532: I don't think this is quite right. The *role* of the where-clauses is distinct from the *role* of the bounds. The bounds tell you what the hidden-type must satisfy; the where-clauses tell you what the generic parameters that the opaque type is used with must satisfy. They do not, in and of themselves, let you make assumptions. In particular, imagine an example like this one:

```rust
opauqe type Foo<T: Eq>: Debug = ...;
```

Here the `T: Eq` is a *where-clause* -- i.e., the idea is that `Foo<T>` is some type which is `Debug`, and it only works where `T: Eq`.

tests/test/opaque_types.rs
```diff
@@ -46,6 +46,26 @@ fn opaque_reveal() {
...
+            struct Ty { }
+
+            trait Clone { }
+            trait Trait { }
+            impl Trait for Ty { }
+            opaque type T: Clone where T: Trait = Ty;
+        }
+
+        goal {
+            T: Trait
```
nikomatsakis, line 62: This test in particular doesn't seem right. The real role of where-clauses is to inform the "well-formedness conditions" for an opaque type, not to allow us to prove things about the placeholder type. This is the role of the bounds.

chalk-integration/src/lowering.rs
detrumi: Hmm, so these should then be a `Vec<TraitBound>` instead of a `<Vec<VariableKind>`, right? I'm not sure how to handle lowering for that correctly, since I don't think the current `LowerTypeKind` approach works for `TraitBound`'s. Would lowering for opaque types then be similar to `AssociatedTyLookup`?

jackh726:
Superseded by #563

PR #513, opened 2020-06-10
Recursive solver factoring and privacy

nikomatsakis:
This branch just starts to factor the recursive solver into a few more files, and tightens up some of the privacy. 

My goal was to try and better separate out the "caching and loop logic", which I hope to eventually move to salsa, from the chalk-specific logic. For now I got as far as moving most of the chalk-specific stuff into the `solve` module.

The next step is going to be trying to factor out the caching stuff into a module that is generic over chalk-specific things, but that's for another day.

### DIFF
--- a/chalk-solve/src/recursive/solve.rs
+++ b/chalk-solve/src/recursive/solve.rs
@@ -4,7 +4,7 @@ use super::lib::{Guidance, Minimums, Solution, UCanonicalGoal};
 use super::Solver;
 use crate::clauses::program_clauses_for_goal;
 use crate::infer::{InferenceTable, ParameterEnaVariableExt};
-use crate::solve::truncate;
+use crate::{solve::truncate, RustIrDatabase};
 use chalk_ir::fold::Fold;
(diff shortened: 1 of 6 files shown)


chalk-solve/src/recursive/solve.rs
```diff
@@ -194,9 +221,17 @@ impl<'me, I: Interner> Solver<'me, I> {
         environment: &Environment<I>,
         goal: &DomainGoal<I>,
     ) -> Result<Vec<ProgramClause<I>>, Floundered> {
-        program_clauses_for_goal(self.program, environment, goal)
+        program_clauses_for_goal(self.db(), environment, goal)
     }
 }
+
+impl<S, I> SolveIterationHelpers<I> for S
```
jackh726, line 228: Wait, extra?

nikomatsakis: There are two traits -- one is the public method, the other is the private helpers, and you need an impl for both.

nikomatsakis:
(Yes, it's just moving things around)

PR #514, opened 2020-06-11
Refactor ProgramClauseData to remove Implies variant

### LINKED ISSUE #511: simplify ProgramClauseImplication
Currently we have the [`ProgramClause`] type, which maps to an enum [`ProgramClauseData`] with two options:

* Implies -- a [`ProgramClauseImplication`], corresponding to `P :- Q1..Qn` (i.e., `P` is true if `Q1...Qn` are provable)
* ForAll -- which wraps the same [`ProgramClauseImplication`] in binders, corresponding to `forall<...> { P :- Q1..qn }`.

I think we should remove the `Implies` variant and just have that a `ProgramClause` maps to a `Binders<ProgramClauseImplication>`, but those binders may be empty.

This is a straight refactoring and simplification.

[`ProgramClause`]: http://rust-lang.github.io/chalk/chalk_ir/struct.ProgramClause.html
[`ProgramClauseData`]: http://rust-lang.github.io/chalk/chalk_ir/enum.ProgramClauseData.html
[`ProgramClauseImplication`]: http://rust-lang.github.io/chalk/chalk_ir/struct.ProgramClauseImplication.html

This issue has been assigned to @declanvk via [this comment](https://github.com/rust-lang/chalk/issues/511#issuecomment-642337691).

declanvk:
Closes #511 

~~As far as I can tell, I didn't change the logic of anything. However, when I checked the test suite there were still some failing tests. I've listed them here just in case they are relevant.~~

Initially failing tests
  
```
test test::coherence::downstream_impl_of_fundamental_43355 ... FAILED
test test::coherence::fundamental_traits ... FAILED
test test::coherence::local_negative_reasoning_in_coherence ... FAILED
test test::coherence::nonoverlapping_assoc_types ... FAILED
test test::coherence::overlapping_assoc_types ... FAILED
test test::coherence::overlapping_assoc_types_error ... FAILED
test test::coherence::two_blanket_impls ... FAILED
test test::misc::flounder_ambiguous ... FAILED
test test::projection::normalize_into_iterator ... FAILED
test test::projection::projection_from_env_a ... FAILED
test test::projection::projection_from_env_slow ... FAILED
test test::wf_goals::struct_wf ... FAILED
test test::wf_lowering::copy_constraints ... FAILED
test test::wf_lowering::drop_constraints ... FAILED
test test::wf_lowering::implied_bounds ... FAILED
```

Looking into the reason now.

### DIFF
--- a/chalk-ir/src/cast.rs
+++ b/chalk-ir/src/cast.rs
@@ -225,18 +225,6 @@ where
     }
 }
 
-impl<I: Interner> CastTo<ProgramClause<I>> for ProgramClauseImplication<I> {
-    fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData(Binders::empty(interner, self.shifted_in(interner))).intern(interner)
-    }
-}
-
-impl<I: Interner> CastTo<ProgramClause<I>> for Binders<ProgramClauseImplication<I>> {
-    fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData(self).intern(interner)
-    }
-}
-
 impl<T, U> CastTo<Option<U>> for Option<T>
 where
     T: CastTo<U>,
--- a/chalk-ir/src/cast.rs
+++ b/chalk-ir/src/cast.rs
@@ -205,7 +205,8 @@ where
             priority: ClausePriority::High,
         };
 
-        ProgramClauseData(Binders::empty(interner, implication)).intern(interner)
+        ProgramClauseData(Binders::empty(interner, implication.shifted_in(interner)))
+            .intern(interner)
     }
 }
 
@@ -226,7 +227,7 @@ where
 
 impl<I: Interner> CastTo<ProgramClause<I>> for ProgramClauseImplication<I> {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData(Binders::empty(interner, self)).intern(interner)
+        ProgramClauseData(Binders::empty(interner, self.shifted_in(interner))).intern(interner)
     }
 }
 
--- a/chalk-ir/src/cast.rs
+++ b/chalk-ir/src/cast.rs
@@ -199,12 +199,13 @@ where
     I: Interner,
 {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData::Implies(ProgramClauseImplication {
+        let implication = ProgramClauseImplication {
             consequence: self.cast(interner),
             conditions: Goals::new(interner),
             priority: ClausePriority::High,
-        })
-        .intern(interner)
+        };
+
+        ProgramClauseData(Binders::empty(interner, implication)).intern(interner)
     }
 }
 
@@ -214,7 +215,7 @@ where
     T: HasInterner<Interner = I> + CastTo<DomainGoal<I>>,
 {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData::ForAll(self.map(|bound| ProgramClauseImplication {
+        ProgramClauseData(self.map(|bound| ProgramClauseImplication {
             consequence: bound.cast(interner),
             conditions: Goals::new(interner),
             priority: ClausePriority::High,
@@ -225,13 +226,13 @@ where
 
 impl<I: Interner> CastTo<ProgramClause<I>> for ProgramClauseImplication<I> {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData::Implies(self).intern(interner)
+        ProgramClauseData(Binders::empty(interner, self)).intern(interner)
     }
 }
 
 impl<I: Interner> CastTo<ProgramClause<I>> for Binders<ProgramClauseImplication<I>> {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData::ForAll(self).intern(interner)
+        ProgramClauseData(self).intern(interner)
     }
 }
 
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -1732,10 +1732,7 @@ impl std::ops::BitAnd for ClausePriority {
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, Fold, HasInterner, Zip)]
-pub enum ProgramClauseData<I: Interner> {
-    Implies(ProgramClauseImplication<I>),
-    ForAll(Binders<ProgramClauseImplication<I>>),
-}
+pub struct ProgramClauseData<I: Interner>(pub Binders<ProgramClauseImplication<I>>);
 
 impl<I: Interner> ProgramClauseImplication<I> {
     pub fn into_from_env_clause(self, interner: &I) -> ProgramClauseImplication<I> {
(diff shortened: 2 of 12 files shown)


declanvk:
My best guess for the error is that by introducing an empty `Binders` it still changes the meaning, especially with respect to folding, via:

https://github.com/rust-lang/chalk/blob/a6400eba1d85c04c6d1c6184635f44a6ce57e1f2/chalk-ir/src/fold/binder_impls.rs#L38-L56

I managed to reproduce the same set of errors (the ones in the list above) with this commit alone: https://github.com/declanvk/chalk/commit/e96058569721eb71587ea294b82ee9c527ebcbf3

chalk-ir/src/cast.rs
```diff
@@ -225,13 +226,13 @@ where
 
 impl<I: Interner> CastTo<ProgramClause<I>> for ProgramClauseImplication<I> {
     fn cast_to(self, interner: &I) -> ProgramClause<I> {
-        ProgramClauseData::Implies(self).intern(interner)
+        ProgramClauseData(Binders::empty(interner, self)).intern(interner)
```
jackh726, line 229: This needs `self.shifted_in()`

chalk-solve/src/clauses/builder.rs
```diff
@@ -77,18 +77,13 @@ impl<'me, I: Interner> ClauseBuilder<'me, I> {
...
-                    VariableKinds::from(interner, self.binders.clone()),
-                    clause,
-                ))
-                .intern(interner),
-            );
-        }
+        self.clauses.push(
+            ProgramClauseData(Binders::new(
+                VariableKinds::from(interner, self.binders.clone()),
+                clause,
```
jackh726, line 83: So, we should be careful here. We indeed may need this to be `clause.shifted_in()` if `self.binders.len() == 0`. This is a little bit awkward. The builder I guess should always be assuming a level of binders.

declanvk:
That fixed it! I had to export the `Shift` folder from `chalk-ir`, let me know if I should refactor this at all.

jackh726:
Yeah, that was going to be my next comment: if we got rid of the Cast impl for one that does the shift, can we find the root "cause" there

nikomatsakis:
I wonder if those cast impls are even used. Probably.

jackh726:
@nikomatsakis nope, neither `Cast` impl is used :/

nikomatsakis:
Well then let's just remove them and avoid the problem =)

declanvk:
I removed the `Cast` impls and fixed the locations that failed by wrapping empty `Binders` and unconditionally applying `shifted_in`. I was contemplating some sort of `Binders` helper that looks like:
```rust
impl<I, T> Binders<T>
where
    T: Visit<I> + Fold<I, I, Result = T> + HasInterner<Interner = I>,
    I: Interner,
{
    pub fn empty_shift_if_required(interner: &T::Interner, value: T) -> Self {
        let value = if value.has_free_vars(interner) {
            // If this type contains "bound" types/lifetimes, then it
            // needs to be shifted across binders.
            value.shifted_in(interner)
        } else {
            value
        };

        Self::empty(interner, value)
    }
}
```
instead of having to apply `shifted_in` every time `Binders::empty` is used.

chalk-ir/src/cast.rs
```diff
@@ -225,12 +208,6 @@ where
     }
 }
 
-impl<I: Interner> CastTo<ProgramClause<I>> for ProgramClauseImplication<I> {
```
nikomatsakis, line 226: So I think *this* impl was completely unused...

chalk-ir/src/cast.rs
```diff
@@ -193,23 +193,6 @@ impl<I: Interner> CastTo<GenericArg<I>> for GenericArg<I> {
     }
 }
 
-impl<T, I> CastTo<ProgramClause<I>> for T
```
nikomatsakis, line 196: ...but maybe we should keep this impl, which would let us revert the diffs below.

chalk-ir/src/lib.rs
```diff
@@ -2110,11 +2110,12 @@ impl<I: Interner> Goal<I> {
...
+                let clauses = vec![DomainGoal::Compatible(()), DomainGoal::DownstreamType(ty)]
+                    .into_iter()
+                    .map(|clause| Binders::empty(interner, clause.shifted_in(interner)));
+
                 GoalData::Implies(
-                    ProgramClauses::from(
-                        interner,
-                        vec![DomainGoal::Compatible(()), DomainGoal::DownstreamType(ty)],
-                    ),
+                    ProgramClauses::from(interner, clauses),
```
nikomatsakis, line 2118: Hmm I think this code read more nicely before, and it's a good example that (a) is using precisely the pattern I was nervous about but (b) reads very naturally. i.e., here we have a goal `DownstreamType(ty)`, where the `ty` will contain a bound variable at some depth, and we are (implicitly, as part of the cast) inserting a `forall` binder around it, but the fact that the adjustment is happening without us having to talk about it feels very natural. i.e., we didn't say anything about this binder, so it's just managed for us.

chalk-solve/src/wf.rs
```diff
@@ -326,9 +326,14 @@ fn impl_wf_environment<'i, I: Interner>(
...
-        .into_iter()
-        .map(move |ty| ty.into_from_env_goal(interner).cast(interner));
+    let types_wf = types.into_iter().map(move |ty| {
+        let bound_goal = Binders::empty(
+            interner,
+            ty.into_from_env_goal(interner).shifted_in(interner),
+        );
+
+        bound_goal.cast(interner)
+    });
```
nikomatsakis, line 336: same here

nikomatsakis:
@declanvk 

> I was contemplating some sort of Binders helper that looks like

I would perhaps instead just make the `shifted_in` helper check if there are no free variables and return a clone if so -- although there is a bit of a complication in that we might need to account for the possibility of `T::Result` not being equal to `T`.

nikomatsakis:
But in any case that's just an optimization, so let's address it separately. It actually fits in with another theme of work I think we should pursue, which is propagating "flags" and the like, similar to what rustc has, which can enable us to create fast-paths quite efficiently.

chalk-ir/src/cast.rs
jackh726: Yes, this and the one underneath (for `Binders<ProgramClauseImplication<I>>`

chalk-ir/src/cast.rs
jackh726: I didn't check this one, whoops.

chalk-ir/src/lib.rs
jackh726: I do think we should maybe file an issue to follow up on this. I don't like the implicit `forall` here

nikomatsakis: Hmm, I convinced myself of the opposite, that the implicit forall was not a problem, so long as it doesn't bind anything (after all, it has no meaning then). Can you say a bit more about what's bothering you about it?

I think what's important to me is that we should be trying to make the high-level code relatively readable and "mappable" to the underlying logic with relatively little "noise". I think this code gets pretty dense and the more we have random iterates and meaningless shifts in there, the more confusing it is. But I *am* amenable to creating more explicit helpers to cover specific scenarios, and relying less on general traits like `Cast` that might be easily misused.

jackh726: So, I guess I haven't convinced myself one way or the other whether I do or don't like the implicit `forall`. I guess it's not that bad since we *are* shifting.

nikomatsakis: So, shall we leave the Cast impl in place and file a follow-up PR to consider removing it and replacing it with some other pattern?

Or should we leave *this* version? 

I guess I don't care that much either way. I have a slight preference for the former.

jackh726: I think we should just leave the Cast impl in place. I don't necessarily think we need a followup issue at this point; mostly because I can't really think of a better pattern. Thinking about this more, it's actually not as bad as originally sort of imagined it. Mostly because if you had something like `(^0.0).cast()` that would be shifted to `^1.0`. So you can't "lose" the information that you forgot to wrap your type in a `Binders`

nikomatsakis: OK.

declanvk:
I removed the unused `Cast` impls and put back the `CastTo<ProgramClause<I>> for T` one with a `shifted_in` inside of it.

Should I rebase/squash to address conflicts?

jackh726:
@declanvk Squashing doesn't matter. Rebase yes :)

PR #519, opened 2020-06-11
Add closures

jackh726:
c.c. #368

This works but isn't rebased.

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -147,13 +147,22 @@ impl<'k> Env<'k> {
         }
 
         if let Some(id) = self.closure_ids.get(&name.str) {
-            return Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
-                name: chalk_ir::TypeName::Closure(id.clone()),
-                // See note in `program`. Unlike rustc, we store upvars separately.
-                substitution: chalk_ir::Substitution::empty(interner),
-            })
-            .intern(interner)
-            .cast(interner));
+            let k = self.closure_kind(*id);
+            if k.binders.len(interner) > 0 {
+                return Err(RustIrError::IncorrectNumberOfTypeParameters {
+                    identifier: name.clone(),
+                    expected: k.binders.len(interner),
+                    actual: 0,
+                });
+            } else {
+                return Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
+                    name: chalk_ir::TypeName::Closure(id.clone()),
+                    // See note in `program`. Unlike rustc, we store upvars separately.
+                    substitution: chalk_ir::Substitution::empty(interner),
+                })
+                .intern(interner)
+                .cast(interner));
+            }
         }
 
         if let Some(id) = self.opaque_ty_ids.get(&name.str) {
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -2,8 +2,8 @@ use crate::interner::{ChalkFnAbi, ChalkIr};
 use chalk_ir::cast::{Cast, Caster};
 use chalk_ir::interner::{HasInterner, Interner};
 use chalk_ir::{
-    self, AdtId, AssocTypeId, BoundVar, ClausePriority, DebruijnIndex, FnDefId, ImplId, OpaqueTyId,
-    QuantifiedWhereClauses, Substitution, ToGenericArg, TraitId, TyKind,
+    self, AdtId, AssocTypeId, BoundVar, ClausePriority, ClosureId, DebruijnIndex, FnDefId, ImplId,
+    OpaqueTyId, QuantifiedWhereClauses, Substitution, ToGenericArg, TraitId, TyKind,
 };
 use chalk_ir::{debug, debug_heading};
 use chalk_parse::ast::*;
@@ -20,6 +20,7 @@ use crate::{Identifier as Ident, RawId, TypeKind, TypeSort};
 
 type AdtIds = BTreeMap<Ident, chalk_ir::AdtId<ChalkIr>>;
 type FnDefIds = BTreeMap<Ident, chalk_ir::FnDefId<ChalkIr>>;
+type ClosureIds = BTreeMap<Ident, chalk_ir::ClosureId<ChalkIr>>;
 type TraitIds = BTreeMap<Ident, chalk_ir::TraitId<ChalkIr>>;
 type OpaqueTyIds = BTreeMap<Ident, chalk_ir::OpaqueTyId<ChalkIr>>;
 type AdtKinds = BTreeMap<chalk_ir::AdtId<ChalkIr>, TypeKind>;
@@ -42,6 +43,7 @@ struct Env<'k> {
     fn_def_ids: &'k FnDefIds,
     fn_def_kinds: &'k FnDefKinds,
     fn_def_abis: &'k FnDefAbis,
+    closure_ids: &'k ClosureIds,
     trait_ids: &'k TraitIds,
     trait_kinds: &'k TraitKinds,
     opaque_ty_ids: &'k OpaqueTyIds,
@@ -139,6 +141,26 @@ impl<'k> Env<'k> {
                 .intern(interner)
                 .cast(interner));
             }
+        }
+
+        if let Some(id) = self.closure_ids.get(&name.str) {
+            return Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
+                name: chalk_ir::TypeName::Closure(id.clone()),
+                // XXX closure upvars
+                substitution: chalk_ir::Substitution::from(
+                    interner,
+                    Some(
+                        chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
+                            name: chalk_ir::TypeName::Tuple(0),
+                            substitution: chalk_ir::Substitution::empty(interner),
+                        })
+                        .intern(interner)
+                        .cast(interner),
+                    ),
+                ),
+            })
+            .intern(interner)
+            .cast(interner));
         }
 
         if let Some(id) = self.opaque_ty_ids.get(&name.str) {
@@ -318,6 +340,7 @@ impl LowerProgram for Program {
 
         let mut adt_ids = BTreeMap::new();
         let mut fn_def_ids = BTreeMap::new();
+        let mut closure_ids = BTreeMap::new();
         let mut trait_ids = BTreeMap::new();
         let mut opaque_ty_ids = BTreeMap::new();
         let mut adt_kinds = BTreeMap::new();
@@ -340,6 +363,10 @@ impl LowerProgram for Program {
                     fn_def_ids.insert(type_kind.name.clone(), id);
                     fn_def_kinds.insert(id, type_kind);
                     fn_def_abis.insert(id, defn.abi.lower()?);
+                }
+                Item::ClosureDefn(defn) => {
+                    let id = ClosureId(raw_id);
+                    closure_ids.insert(defn.name.str.clone(), id);
                 }
                 Item::TraitDefn(defn) => {
                     let type_kind = defn.lower_type_kind()?;
@@ -364,6 +391,7 @@ impl LowerProgram for Program {
 
         let mut adt_data = BTreeMap::new();
         let mut fn_def_data = BTreeMap::new();
+        let mut closure_data = BTreeMap::new();
         let mut trait_data = BTreeMap::new();
         let mut well_known_traits = BTreeMap::new();
         let mut impl_data = BTreeMap::new();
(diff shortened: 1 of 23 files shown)


jackh726:
Okay I just applied my changes in a new commit against master, since that was easier.

@nikomatsakis I would appreciate a review. The biggest "missing" part of this PR is tests. Not sure the best was to test closures currently.

chalk-solve/src/clauses/builtin_traits/copy.rs
```diff
@@ -54,6 +54,24 @@ pub fn add_copy_program_clauses<I: Interner>(
             TypeName::FnDef(_) => {
                 builder.push_fact(trait_ref.clone());
             }
+            TypeName::Closure(_) => {
+                let interner = db.interner();
+                match substitution
```
nikomatsakis, line 59: I think we should just require that `substitution.parameters(interner).last`, or even `substitution.parameters(interner)`, be `copy` (or clone, etc) without bothering to unpack that it is a tuple. This is the direction rustc is going as well, for various reasons.

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -19,11 +20,116 @@ use chalk_ir::{
...
+                            alias,
+                            ty: return_type,
+                        });
+                    }
+                });
+                Ok(())
+            }
+            TypeName::Closure(closure_id) => {
+                let closure_datum = db.closure_datum(closure_id, apply.substitution.clone());
+                let trait_matches = match well_known {
```
nikomatsakis, line 83: total nit, but maybe `match (well_known, closure_datum.kind) { ... }`?

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -19,11 +20,116 @@ use chalk_ir::{
...
+                        closure_datum.kind,
+                        ClosureKind::FnOnce | ClosureKind::FnMut | ClosureKind::Fn
+                    ),
+                    _ => false,
+                };
+                if trait_matches {
+                    builder.push_binders(
+                        &closure_datum.inputs_and_output,
+                        |builder, inputs_and_output| {
+                            let substitution = Substitution::from(
```
nikomatsakis, line 98: this logic...

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -19,11 +20,116 @@ use chalk_ir::{
...
+                                        .cloned()
+                                        .map(|ty| ty.cast(interner)),
+                                ),
+                            );
+                            builder.push_fact(TraitRef {
+                                trait_id,
+                                substitution: substitution.clone(),
+                            });
+
+                            if let WellKnownTrait::FnOnce = well_known {
```
nikomatsakis, line 113: ...and especially this logic...feels like it is duplicated with the `FnDef` case, can we maybe extract a helper function?

jackh726:
>  Downloaded lalrpop v0.19.0
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/lalrpop-0.19.0/Cargo.toml`

The heck. I'm just gonna restart CI.

chalk-solve/src/clauses/builtin_traits/copy.rs
jackh726: Actually that is kind of nice.

jackh726:
So, testing closures with no upvars is *fairly* easy (like you said, a `closure foo(...)` syntax isn't terrible). The question is more of how would we test upvars?

We could do something like
```
closure foo[...]()
```
where the `[...]` would be the upvars. But even still, the question is how do we write the upvars? `[Foo, &mut Bar]`?

nikomatsakis:
I was imagining 

```
closure foo<...>(self|&self|&mut self, t0...tn) -> treturn {
   t_upvar1,
   t_upvar2,
   ...
}
```

where the self part captures fn/fn-mut/fn-once.

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -25,107 +25,96 @@ pub fn add_fn_trait_program_clauses<I: Interner>(
 ) -> Result<(), Floundered> {
     let interner = db.interner();
     let trait_id = db.well_known_trait_id(well_known).unwrap();
+
+    let push_clauses = |builder: &mut ClauseBuilder<'_, I>,
```
nikomatsakis, line 29: This is a total style nit and more of a suggestion, but I find really big and complex closures hard to read, and would rather see this pulled out as a private helper function.

nikomatsakis: (I realize it does use a few variables from the environment, so it would require a few extra parameters; I think I just prefer to see those parameters so I can tell just what it needs, and also digest each function more in isolation. If it's a *lot* of parameters then I would try to use methods somehow.)

chalk-parse/src/parser.lalrpop
```diff
@@ -102,6 +103,21 @@ FnArgs: Vec<Ty> = {
...
+        kind: s,
+        argument_types: vec![],
+        return_type: Ty::Tuple { types: Vec::new() },
+    }
+}
+
+ClosureSelf: ClosureKind = {
+    "self" => ClosureKind::FnOnce,
+    "mutself" => ClosureKind::FnMut,
+    "refself" => ClosureKind::Fn,
```
jackh726, line 118: @nikomatsakis So question about lalrpop: If I use `self`, `&mut self`, and `&self` here instead, I get the following error in lowering:
`lowering error: parse error: UnrecognizedToken { token: (35, Token(7, "&"), 36), expected: ["\"&mut self\"", "\"&self\"", "\"self\""] }`
What's the "correct" way to do this?

nikomatsakis: @jackh726 you probably want `"&" "self"` or something like that, i.e., two distinct tokens.

jackh726:
This should be ready to review

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -19,11 +65,71 @@ use chalk_ir::{
...
+                    | (WellKnownTrait::FnMut, ClosureKind::Fn) => true,
+                    (WellKnownTrait::FnOnce, _) => true,
+                    _ => false,
+                };
+                if !trait_matches {
+                    return Ok(());
+                }
+                let inputs_and_output = &closure_datum.inputs_and_output;
+                builder.push_binders(inputs_and_output, |builder, inputs_and_output| {
+                    let arg_sub = inputs_and_output
```
nikomatsakis, line 117: Nit: it feels like the construction of `arg_sub` and `output_ty` could be moved into the helper routine as well..? I think they are identical

nikomatsakis: Ah, I see, there's a third call. I guess then there could be a second helper.

nikomatsakis: But it's ok as is.

chalk-solve/src/lib.rs
```diff
@@ -101,6 +101,14 @@ pub trait RustIrDatabase<I: Interner>: Debug {
 
     /// Check if a trait is object safe
     fn is_object_safe(&self, trait_id: TraitId<I>) -> bool;
+
+    fn closure_datum(
```
nikomatsakis, line 105: Doc comment

chalk-solve/src/lib.rs
```diff
@@ -101,6 +101,14 @@ pub trait RustIrDatabase<I: Interner>: Debug {
...
+        &self,
+        closure_id: ClosureId<I>,
+        substs: &Substitution<I>,
+    ) -> Arc<ClosureDatum<I>>;
+
+    fn closure_upvars(&self, closure_id: ClosureId<I>, substs: &Substitution<I>) -> Binders<Ty<I>>;
```
nikomatsakis, line 111: Doc comment -- for example, are there are any constraints chalk places like "this must be a tuple" etc? I don't think so.

chalk-solve/src/rust_ir.rs
```diff
@@ -168,6 +168,15 @@ pub struct FnDefDatumBound<I: Interner> {
     pub where_clauses: Vec<QuantifiedWhereClause<I>>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ClosureDatum<I: Interner> {
```
nikomatsakis, line 172: Doc comment, also for the fields -- it's nice to have something show up in rustdoc :)

chalk-solve/src/rust_ir.rs
```diff
@@ -591,3 +600,10 @@ impl Polarity {
         }
     }
 }
+
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
```
nikomatsakis, line 604: doc comment, although I confess this feels fairly self explanatory. 

> Indicates the "most permissive" `Fn`-like trait that the closure implements. If the closure kind for a closure is `FnMut`, for example, then the closure implements `FnMut` and `FnOnce`.

tests/test/closures.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        // A closure with non-Copy upvars is not copy
+        goal {
+            foobuzz: Copy
+        } yields {
+            "No possible solution"
+        }
+        // A closure with only Copy upvars is copy
+        goal {
+            foobar: Copy
```
nikomatsakis, line 99: I'm sort of surprised we don't see `foobar<'x>` here

tests/test/closures.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        goal {
+            forall<'a> {
+                Normalize(<foobar<'a> as FnOnce<(u8, f32)>>::Output -> u32)
+            }
+        } yields {
+            "Unique"
+        }
+        goal {
+            foobuzz: FnOnce<(u8, f32)>
```
nikomatsakis, line 249: Similarly I would expect `'a` etc here

chalk-solve/src/rust_ir.rs
```diff
@@ -168,6 +168,15 @@ pub struct FnDefDatumBound<I: Interner> {
     pub where_clauses: Vec<QuantifiedWhereClause<I>>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ClosureDatum<I: Interner> {
+    pub id: ClosureId<I>,
+
+    pub inputs_and_output: Binders<FnDefInputsAndOutputDatum<I>>,
```
nikomatsakis, line 175: probably worth pointing people at `inputs_and_output` from `FnDef` to explain the late-bound business

tests/test/closures.rs
```diff
@@ -0,0 +1,259 @@
...
+            closure foobuzz<'a>(self, a: u8, b: f32) -> u32 {
+                u8;
+                &'a u16;
+                &'a mut u32
+            }
+            closure foobar<'a>(self, a: u8, b: f32) -> u32 {
+                u8;
+                &'a u16
+            }
+        }
```
nikomatsakis, line 72: can we add a test with generic types? maybe

```
closure with_ty<T>(self) { T; }
```

and then

```
forall<T> { with_ty<T>: Copy }
```

should not be true, but `forall<T> { if (T: Copy) { .. } }` should be provable.

chalk-integration/src/lowering.rs
```diff
@@ -132,15 +137,25 @@ impl<'k> Env<'k> {
...
+                    name: chalk_ir::TypeName::FnDef(id.clone()),
                     substitution: chalk_ir::Substitution::empty(interner),
                 })
                 .intern(interner)
                 .cast(interner));
             }
         }
 
+        if let Some(id) = self.closure_ids.get(&name.str) {
+            return Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
```
nikomatsakis, line 150: I think we're missing the check for `k.binders.len() != 0` here?

nikomatsakis: Also this code looks like it should maybe be factored and shared better -- I presume this match arm covers only the case where there were no type parameters supplied. That probably didn't matter when lowering was simpler ...

nikomatsakis: I must be missing the case where type parameters are supplied...

chalk-solve/src/clauses/builtin_traits/fn_family.rs
jackh726: I went ahead and added a helper function for the fndef and closure case.

chalk-integration/src/lowering.rs
jackh726: Yeah, I wasn't handling this correctly.

jackh726:
Ok @nikomatsakis I addressed your comments. I wasn't handling the generic types correctly. I also just moved from the single `ClosureDatum` struct to just functions on `RustIrDatabase`, since that seems to be the direction we want to go. Can you look over again? I wrote some doc comments, but they could use a second pair of eyes for clarity.

jackh726:
Btw there might be a bug in here somewhere. Trying to update rustc to use this causes the closure tests to fail. So don't merge yet even if it looks good. Debugging...

jackh726:
False alarm :) this is good for review

Issue #418, opened 2020-04-28
subtyping and type equality


nikomatsakis:
Chalk currently has a notion of type *unification*, but it doesn't have a notion of subtyping. Since trait matching is always invariant, and thus never uses subtyping, you might not think this is important, but it turns out it is when you consider higher-ranked types and variance (annoyingly).

The key invariant is that `T1 = T2` if `T1 <: T2` and `T2 <: T1`. The way we handle this equality for higher-ranked types in the rust repo is to [translate `T1 = T2` into two subtyping requirements](https://github.com/rust-lang/rust/blob/db98d32ea04438cf759b749d5cbeba4f77222664/src/librustc_infer/infer/equate.rs#L133-L136). (In fact, rustc has two bits of code that handle this sort of thing, the other one is [defined here](https://github.com/rust-lang/rust/blob/db98d32ea04438cf759b749d5cbeba4f77222664/src/librustc_infer/infer/nll_relate/mod.rs#L616-L622), and is probably more the model I would want us to use in chalk -- it too translates equality into [two](https://github.com/rust-lang/rust/blob/db98d32ea04438cf759b749d5cbeba4f77222664/src/librustc_infer/infer/nll_relate/mod.rs#L652) [subtyping](https://github.com/rust-lang/rust/blob/db98d32ea04438cf759b749d5cbeba4f77222664/src/librustc_infer/infer/nll_relate/mod.rs#L696) checks.)

The reason that this is specific to higher-ranked types is that:

```
for<'a> A <: for<'b> B
```

if `forall<'b> { exists<'a> { A <: B } }`. In other words, you instantiate one side with `forall` and one with `exists`. For now I will leave it an exercise to the reader, but if you work through `T1 = for<'a, 'b> fn(&'a u32, &'b u32)` and `T2 = for<'c> fn(&'c u32)`, you will find that `T1 <: T2` and `T2 <: T1` both hold, and therefore `T1 = T2` should hold. ([I wrote some notes on why this is here](https://github.com/rust-lang/rust/issues/32330#issuecomment-202536977).) But chalk wouldn't handle that today.

[...]

PR #520, opened 2020-06-11
Add Variance and use it for zip

jackh726:
c.c. #418

I have to rebase this, but we can probably merge this initially. Only `Invariant` is used right now, but this threads it through.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -258,6 +258,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                             info!("program clause = {:#?}", clause);
                             let mut infer = infer.clone();
                             if let Ok(resolvent) = infer.resolvent_clause(
+                                context.unification_database(),
                                 context.interner(),
                                 &environment,
                                 &domain_goal,
@@ -630,6 +631,7 @@ impl<'forest, I: Interner, C: Context<I> + 'forest, CO: ContextOps<I, C> + 'fore
                 );
                 match strand.infer.apply_answer_subst(
                     self.context.interner(),
+                    self.context.unification_database(),
                     &mut strand.ex_clause,
                     &subgoal,
                     table_goal,
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -12,8 +12,8 @@ use crate::{
 
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    Canonical, ConstrainedSubst, Floundered, Goal, GoalData, InEnvironment, NoSolution,
-    Substitution, UCanonical, UniverseMap,
+    Canonical, ConstrainedSubst, FallibleOrFloundered, Floundered, Goal, GoalData, InEnvironment,
+    NoSolution, Substitution, UCanonical, UniverseMap,
 };
 use tracing::{debug, debug_span, info, instrument};
 
@@ -297,21 +297,23 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                 // simplified subgoals. You can think of this as
                 // applying built-in "meta program clauses" that
                 // reduce goals into Domain goals.
-                if let Ok(ex_clause) =
-                    Self::simplify_goal(context, &mut infer, subst, environment, goal)
-                {
-                    info!(
-                        ex_clause = ?infer.debug_ex_clause(context.interner(), &ex_clause),
-                        "pushing initial strand"
-                    );
-                    let strand = Strand {
-                        infer,
-                        ex_clause,
-                        selected_subgoal: None,
-                        last_pursued_time: TimeStamp::default(),
-                    };
-                    let canonical_strand = Self::canonicalize_strand(context, strand);
-                    table.enqueue_strand(canonical_strand);
+                match Self::simplify_goal(context, &mut infer, subst, environment, goal) {
+                    FallibleOrFloundered::Ok(ex_clause) => {
+                        info!(
+                            ex_clause = ?infer.debug_ex_clause(context.interner(), &ex_clause),
+                            "pushing initial strand"
+                        );
+                        let strand = Strand {
+                            infer,
+                            ex_clause,
+                            selected_subgoal: None,
+                            last_pursued_time: TimeStamp::default(),
+                        };
+                        let canonical_strand = Self::canonicalize_strand(context, strand);
+                        table.enqueue_strand(canonical_strand);
+                    }
+                    FallibleOrFloundered::NoSolution => {}
+                    FallibleOrFloundered::Floundered => table.mark_floundered(),
                 }
             }
         }
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -1,4 +1,4 @@
-use crate::context::{Context, ContextOps, InferenceTable};
+use crate::context::{Context, ContextOps, UnificationOps};
 use crate::forest::Forest;
 use crate::{ExClause, Literal, TimeStamp};
 
@@ -14,7 +14,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
     /// includes unifications that cannot be completed.
     pub(super) fn simplify_goal(
         context: &impl ContextOps<I, C>,
-        infer: &mut dyn InferenceTable<I, C>,
+        infer: &mut dyn UnificationOps<I, C>,
         subst: Substitution<I>,
         initial_environment: Environment<I>,
         initial_goal: Goal<I>,
@@ -63,6 +63,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                 }
                 GoalData::EqGoal(goal) => infer.unify_generic_args_into_ex_clause(
                     context.interner(),
+                    context.unification_database(),
                     &environment,
                     &goal.a,
                     &goal.b,
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -4,7 +4,8 @@ use crate::{ExClause, Literal, TimeStamp};
 
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    Environment, Fallible, Goal, GoalData, InEnvironment, QuantifierKind, Substitution, Variance,
+    Environment, FallibleOrFloundered, Goal, GoalData, InEnvironment, QuantifierKind, Substitution,
+    Variance,
 };
 use tracing::debug;
 
@@ -18,7 +19,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
         subst: Substitution<I>,
         initial_environment: Environment<I>,
         initial_goal: Goal<I>,
-    ) -> Fallible<ExClause<I>> {
+    ) -> FallibleOrFloundered<ExClause<I>> {
         let mut ex_clause = ExClause {
             subst,
             ambiguous: false,
@@ -61,7 +62,7 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                             subgoal.clone(),
                         )));
                 }
-                GoalData::EqGoal(goal) => infer.relate_generic_args_into_ex_clause(
+                GoalData::EqGoal(goal) => match infer.relate_generic_args_into_ex_clause(
                     context.interner(),
                     context.unification_database(),
                     &environment,
@@ -69,16 +70,34 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                     &goal.a,
                     &goal.b,
                     &mut ex_clause,
-                )?,
-                GoalData::SubtypeGoal(goal) => infer.relate_generic_args_into_ex_clause(
-                    context.interner(),
-                    context.unification_database(),
-                    &environment,
-                    Variance::Covariant,
-                    &goal.a,
-                    &goal.b,
-                    &mut ex_clause,
-                )?,
+                ) {
+                    Ok(()) => {}
+                    Err(_) => return FallibleOrFloundered::NoSolution,
+                },
+                GoalData::SubtypeGoal(goal) => {
+                    match (goal.a.ty(context.interner()), goal.b.ty(context.interner())) {
+                        (Some(a_ty), Some(b_ty)) => {
+                            if a_ty.inference_var(context.interner()).is_some()
+                                && b_ty.inference_var(context.interner()).is_some()
+                            {
+                                return FallibleOrFloundered::Floundered;
+                            }
+                        }
+                        _ => {}
+                    }
+                    match infer.relate_generic_args_into_ex_clause(
+                        context.interner(),
+                        context.unification_database(),
+                        &environment,
+                        Variance::Covariant,
+                        &goal.a,
+                        &goal.b,
+                        &mut ex_clause,
+                    ) {
+                        Ok(()) => {}
+                        Err(_) => return FallibleOrFloundered::NoSolution,
+                    }
+                }
                 GoalData::DomainGoal(domain_goal) => {
                     ex_clause
                         .subgoals
@@ -94,6 +113,6 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
             }
         }
 
-        Ok(ex_clause)
+        FallibleOrFloundered::Ok(ex_clause)
     }
 }
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -4,7 +4,7 @@ use crate::{ExClause, Literal, TimeStamp};
 
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    Environment, Fallible, Goal, GoalData, InEnvironment, QuantifierKind, Substitution,
+    Environment, Fallible, Goal, GoalData, InEnvironment, QuantifierKind, Substitution, Variance,
 };
 use tracing::debug;
 
@@ -61,15 +61,24 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                             subgoal.clone(),
                         )));
                 }
-                GoalData::EqGoal(goal) => infer.unify_generic_args_into_ex_clause(
+                GoalData::EqGoal(goal) => infer.relate_generic_args_into_ex_clause(
                     context.interner(),
                     context.unification_database(),
                     &environment,
+                    Variance::Invariant,
                     &goal.a,
                     &goal.b,
                     &mut ex_clause,
                 )?,
-                GoalData::SubtypeGoal(_goal) => todo!(),
+                GoalData::SubtypeGoal(goal) => infer.relate_generic_args_into_ex_clause(
+                    context.interner(),
+                    context.unification_database(),
+                    &environment,
+                    Variance::Covariant,
+                    &goal.a,
+                    &goal.b,
+                    &mut ex_clause,
+                )?,
                 GoalData::DomainGoal(domain_goal) => {
                     ex_clause
                         .subgoals
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -75,17 +75,12 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
                     Err(_) => return FallibleOrFloundered::NoSolution,
                 },
                 GoalData::SubtypeGoal(goal) => {
-                    match (goal.a.ty(context.interner()), goal.b.ty(context.interner())) {
-                        (Some(a_ty), Some(b_ty)) => {
-                            if a_ty.inference_var(context.interner()).is_some()
-                                && b_ty.inference_var(context.interner()).is_some()
-                            {
-                                return FallibleOrFloundered::Floundered;
-                            }
-                        }
-                        _ => {}
+                    if goal.a.inference_var(context.interner()).is_some()
+                        && goal.b.inference_var(context.interner()).is_some()
+                    {
+                        return FallibleOrFloundered::Floundered;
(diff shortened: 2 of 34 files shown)


chalk-solve/src/infer/unify.rs
```diff
@@ -157,11 +158,11 @@ impl<'t, I: Interner> Unifier<'t, I> {
             | (&TyData::Dyn(_), &TyData::Function(_)) => Err(NoSolution),
 
             (&TyData::Placeholder(ref p1), &TyData::Placeholder(ref p2)) => {
-                Zip::zip_with(self, p1, p2)
+                Zip::zip_with(self, variance, p1, p2)
             }
 
             (&TyData::Apply(ref apply1), &TyData::Apply(ref apply2)) => {
-                Zip::zip_with(self, apply1, apply2)
+                Zip::zip_with(self, variance, apply1, apply2)
```
nikomatsakis, line 165: Where things are going to get more interesting is that we need to be applying variance *transformations* at points like this. i suspect the right way to manage this is going to be to remove this impl:

https://github.com/rust-lang/chalk/blob/1781c431c149e18aeee6329c8fd502544efd0f6e/chalk-ir/src/zip.rs#L281-L289

and instead have a specialization function for zipping substitutions that takes some way to "fetch" the variance for each element and which [transforms](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.Variance.html#method.xform) the variance appropriate. Similar to [this code in rustc](https://github.com/rust-lang/rust/blob/7c78a5f97de07a185eebae5a5de436c80d8ba9d4/src/librustc_middle/ty/relate.rs#L136-L150).

chalk-solve/src/infer/test.rs
```diff
@@ -12,14 +12,26 @@ fn infer() {
     let a = table.new_variable(U0).to_ty(interner);
     let b = table.new_variable(U0).to_ty(interner);
     table
-        .unify(interner, &environment0, &a, &ty!(apply (item 0) (expr b)))
+        .unify(
```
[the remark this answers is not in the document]
jackh726: Yeah, probably rename. `relate` works for me

chalk-solve/src/infer/unify.rs
jackh726: Yeah...this was the bit that I was unsure about. Still don't know how sure I am. But I can look a bit into how rustc does it.

jackh726:
@nikomatsakis this is still very much a WIP, but I've done a bit more work and if you have time to look at it, comments would be helpful :)

chalk-ir/src/lib.rs
```diff
@@ -2898,3 +2939,16 @@ pub struct AnswerSubst<I: Interner> {
...
 }
+
+/// Logic to decide the Variance for a given subst
+pub trait UnificationDatabase<I>
+where
+    Self: std::fmt::Debug,
+    I: Interner,
+{
+    /// Gets the variances for the substitution of a fn def
+    fn fn_def_variance(&self, fn_def_id: FnDefId<I>) -> Vec<Variance>;
```
nikomatsakis, line 2950: `I::Variances` would probably be better

chalk-ir/src/zip.rs
```diff
@@ -21,52 +21,91 @@ use std::sync::Arc;
...
+        &mut self,
+        variances: Option<Vec<Variance>>,
+        a: &[GenericArg<I>],
+        b: &[GenericArg<I>],
+    ) -> Fallible<()>
+    where
+        Self: Sized,
+    {
+        for (i, (a, b)) in a.iter().zip(b.iter()).enumerate() {
+            let variance = variances.as_ref().map_or(Variance::Invariant, |v| v[i]);
```
nikomatsakis, line 60:
```suggestion
            let variance = variances.as_ref().map(|v| v[i]).unwrap_or(Variance::Invariant);
```

ymmv but I personally find this formulation much easier to read than `map_or`

chalk-ir/src/zip.rs
```diff
@@ -21,52 +21,91 @@ use std::sync::Arc;
...
+        b: &Binders<T>,
+    ) -> Fallible<()>
     where
         T: HasInterner<Interner = I> + Zip<I> + Fold<I, I, Result = T>;
 
-    /// Retreives the interner from the underlying zipper object
+    /// Zips two substs
+    fn zip_substs(
+        &mut self,
+        variances: Option<Vec<Variance>>,
```
nikomatsakis, line 52: this needs to have a `variance0: Variance` parameter...

chalk-ir/src/zip.rs
```diff
@@ -21,52 +21,91 @@ use std::sync::Arc;
...
+        a: &[GenericArg<I>],
+        b: &[GenericArg<I>],
+    ) -> Fallible<()>
+    where
+        Self: Sized,
+    {
+        for (i, (a, b)) in a.iter().zip(b.iter()).enumerate() {
+            let variance = variances.as_ref().map_or(Variance::Invariant, |v| v[i]);
+            Zip::zip_with(self, variance, a, b)?;
```
nikomatsakis, line 61: ...and then here you would do `variance0.xform(variance)`, where [this is the `xform` function](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.Variance.html#method.xform)

chalk-ir/src/zip.rs
```diff
@@ -324,21 +452,184 @@ impl<I: Interner> Zip<I> for VariableKind<I> {
...
+                };
+                Zip::zip_with(
+                    zipper,
+                    variance,
+                    a.substitution.iter(interner).last().unwrap(),
+                    b.substitution.iter(interner).last().unwrap(),
+                )?;
+                Ok(())
+            }
+            _ => zipper.zip_substs(
```
nikomatsakis, line 586: Nit: I'd prefer to see this `_` written out explicitly. My usual rubric is "if a new variant is added, how likely is it that it wants custom handling?" I'd say the answer here is "reasonably likely"

chalk-ir/src/zip.rs
```diff
@@ -324,21 +452,184 @@ impl<I: Interner> Zip<I> for VariableKind<I> {
...
+    fn zip_with<'i, Z: Zipper<'i, I>>(
+        zipper: &mut Z,
+        variance: Variance,
+        a: &Self,
+        b: &Self,
+    ) -> Fallible<()>
+    where
+        I: 'i,
+    {
+        Zip::zip_with(zipper, variance, &a.bounds, &b.bounds)?;
```
nikomatsakis, line 605: It probably falls out elsewhere but I think this wants to be invariant...

chalk-ir/src/zip.rs
```diff
@@ -324,21 +452,184 @@ impl<I: Interner> Zip<I> for VariableKind<I> {
...
+        _variance: Variance,
+        a: &Self,
+        b: &Self,
+    ) -> Fallible<()>
+    where
+        I: 'i,
+    {
+        let interner = zipper.interner();
+        zipper.zip_substs(
+            None,
```
nikomatsakis, line 623: These are the fn parameters, right? If so, we should comment, but also this should be contravariant.

chalk-ir/src/zip.rs
```diff
@@ -324,21 +452,184 @@ impl<I: Interner> Zip<I> for VariableKind<I> {
...
+            &a.0.parameters(interner)[..a.0.len(interner) - 1],
+            &b.0.parameters(interner)[..b.0.len(interner) - 1],
+        )?;
+        Zip::zip_with(
+            zipper,
+            Variance::Contravariant,
```
nikomatsakis, line 629: If this is the return type, it should be Covariant.

chalk-ir/src/zip.rs
nikomatsakis: Well, actually, it should be `variance.xform(Contravariant)`

chalk-ir/src/zip.rs
nikomatsakis: Well actually it should be `variance.xform(Covariant)`

nikomatsakis: which is always just `variance`

chalk-solve/src/infer/unify.rs
```diff
@@ -472,31 +493,85 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
+                .push(InEnvironment::new(self.environment, b_outlives_a));
+        }
+        if matches!(variance, Variance::Invariant | Variance::Covariant) {
+            let a_outlives_b = GoalData::AddRegionConstraint(a, b).intern(interner);
+            self.goals
+                .push(InEnvironment::new(self.environment, a_outlives_b));
+        }
+    }
+
+    fn with_local_variance<R>(
```
nikomatsakis, line 510: not sure what this does...

nikomatsakis: kind of looks like the xform function

chalk-ir/src/zip.rs
nikomatsakis: Oh and I think we want some other bits of variance here...

* [`Ref`](http://rust-lang.github.io/chalk/chalk_ir/enum.TypeName.html#variant.Ref) is contravariant in its lifetime parameter, and either covariant (if Imm) or invariant (if Mutable) in its type parameter
* [`Raw`](http://rust-lang.github.io/chalk/chalk_ir/enum.TypeName.html#variant.Raw) is covariant if const, invariant if mutable

probably more...

chalk-ir/src/lib.rs
jackh726: Hmm true. But, I honestly would prefer to avoid creating these at all. I'm wondering if maybe a
```
fn with_fn_def_variances<R>(&self, fn_def_id: FnDefId<I>, op: impl Fn(&[Variance])) -> R;
```
would be better.

chalk-ir/src/zip.rs
jackh726: I can change this, but I actually just copied this bit from rustc.

chalk-ir/src/lib.rs
nikomatsakis: That is not possible with a `dyn` trait. But there could be something like `&self -> &[Variance]`, potentially, though that's kind of limiting.

chalk-ir/src/zip.rs
nikomatsakis: Doesn't matter too much. I just have a thing against `map_or` :P

chalk-ir/src/lib.rs
jackh726: Well, we actually have to get this via `RustIrDatabase::unification_database()`, so it doesn't have to be `dyn`.

nikomatsakis: Well, we're not generic over the `RustIrDatabase`, so that means it must be `dyn`. Anyway I think we *should* make it `dyn` or at least `dyn` compatible.

chalk-engine/src/logic.rs
```diff
@@ -296,21 +297,23 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
...
+                            infer,
+                            ex_clause,
+                            selected_subgoal: None,
+                            last_pursued_time: TimeStamp::default(),
+                        };
+                        let canonical_strand = Self::canonicalize_strand(context, strand);
+                        table.enqueue_strand(canonical_strand);
+                    }
+                    FallibleOrFloundered::NoSolution => {}
+                    FallibleOrFloundered::Floundered => table.mark_floundered(),
```
jackh726, line 316: So, this is somewhat of a question of what to do here. If the goal is a `SubtypeGoal` and both types are `InferenceVar`s, do we want to `Flounder` or have `NoSolution`?

chalk-engine/src/simplify.rs
```diff
@@ -61,13 +62,42 @@ impl<I: Interner, C: Context<I>> Forest<I, C> {
...
+                        (Some(a_ty), Some(b_ty)) => {
+                            if a_ty.inference_var(context.interner()).is_some()
+                                && b_ty.inference_var(context.interner()).is_some()
+                            {
+                                return FallibleOrFloundered::Floundered;
+                            }
+                        }
+                        _ => {}
+                    }
+                    match infer.relate_generic_args_into_ex_clause(
```
jackh726, line 88: So, I made `SubtypeGoal` actually have two `GenericArg`s to be able to reuse this function. (And similarly the one in `recursive`). But I think I just might change this function to be generic over `Zip`

chalk-integration/src/program.rs
```diff
@@ -326,6 +327,30 @@ impl tls::DebugContext for Program {
     }
 }
 
+impl UnificationDatabase<ChalkIr> for Program {
```
jackh726, line 330: I should have these actually use something we can test. But I'm not actually sure the best way to approach this.

chalk-ir/src/lib.rs
```diff
@@ -1731,6 +1845,16 @@ pub struct EqGoal<I: Interner> {
 
 impl<I: Interner> Copy for EqGoal<I> where I::InternedGenericArg: Copy {}
 
+/// Subtype goal: tries to prove that `a` is a subtype of `b`
+#[derive(Clone, PartialEq, Eq, Hash, Fold, Visit, Zip)]
+#[allow(missing_docs)]
+pub struct SubtypeGoal<I: Interner> {
```
jackh726, line 1851: The alternative here was to add a `variance: Variance` field to `EqGoal` and change it to `RelateGoal`

chalk-ir/src/zip.rs
```diff
@@ -335,21 +472,220 @@ impl<I: Interner> Zip<I> for VariableKind<I> {
...
+        zipper: &mut Z,
+        variance: Variance,
+        a: &Self,
+        b: &Self,
+    ) -> Fallible<()>
+    where
+        I: 'i,
+    {
+        let interner = zipper.interner();
+        Zip::zip_with(zipper, variance, &a.trait_id, &b.trait_id)?;
```
jackh726, line 515: This should probably just be `Invariant`

jackh726: In addition to the other ids.

jackh726: But it also doesn't really matter since these only check for equality.

chalk-parse/src/parser.lalrpop
```diff
@@ -464,6 +464,8 @@ LeafGoal: LeafGoal = {
     <dg: DomainGoal> => LeafGoal::DomainGoal { goal: dg },
 
     <a:GenericArg> "=" <b:GenericArg> => LeafGoal::UnifyGenericArgs { a, b },
+
+    <a:GenericArg> "subtype" <b:GenericArg> => LeafGoal::SubtypeGenericArgs { a, b },
```
jackh726, line 468: Help? Neither `"<:"` and `"<" ":"` work.

chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -139,13 +138,15 @@ pub fn add_fn_trait_program_clauses<I: Interner>(
             _ => Ok(()),
         },
         TyData::Function(fn_val) => {
-            let (binders, orig_sub) = fn_val.into_binders_and_value(interner);
-            let bound_ref = Binders::new(VariableKinds::from(interner, binders), orig_sub);
+            let bound_ref = fn_val.as_binders(interner);
+            //let (binders, orig_sub) = fn_val.into_binders_and_value(interner);
+            //let bound_ref = Binders::new(VariableKinds::from(interner, binders), orig_sub);
```
jackh726, line 143: I can remove these.

chalk-solve/src/infer/test.rs
```diff
@@ -1,9 +1,23 @@
...
-use super::unify::UnificationResult;
+use super::unify::RelationResult;
 use super::*;
 use chalk_integration::interner::ChalkIr;
 
+// We just use a vec of 20 `Invariant`, since this is zipped and no substs are
+// longer than this
+#[derive(Debug)]
+struct TestDatabase;
+impl UnificationDatabase<ChalkIr> for TestDatabase {
```
jackh726, line 11: Might be worth making `TestDatabase` stateful for each function. But that adds a bunch more to each use, so I didn't do it yet. With more tests for subtyping, it might be needed.

chalk-solve/src/infer/unify.rs
```diff
@@ -57,53 +63,68 @@ impl<'t, I: Interner> Unifier<'t, I> {
             table: table,
             goals: vec![],
             interner,
+            db,
         }
     }
 
-    /// The main entry point for the `Unifier` type and really the
+    /// The main entry points for the `Unifier` type and really the
```
jackh726, line 70: Whoops.

chalk-solve/src/infer/unify.rs
```diff
@@ -57,53 +63,68 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
+                        Ok(())
+                    },
                 }
             }
 
-            // Unifying an inference variable with a non-inference variable.
+            // Relating an inference variable with a non-inference variable.
             (&TyData::InferenceVar(var, kind), ty_data @ &TyData::Apply(_))
             | (&TyData::InferenceVar(var, kind), ty_data @ &TyData::Placeholder(_))
             | (&TyData::InferenceVar(var, kind), ty_data @ &TyData::Dyn(_))
```
jackh726, line 130: I haven't updated this. I'm not sure if there's a way to do this *without* generalization?

chalk-solve/src/recursive/fulfill.rs
```diff
@@ -323,7 +338,20 @@ impl<'s, I: Interner, Solver: SolveDatabase<I>, Infer: RecursiveInferenceTable<I
...
-                self.unify(&environment, &a, &b)?;
+                self.unify(&environment, Variance::Invariant, &a, &b)?;
+            }
+            GoalData::SubtypeGoal(SubtypeGoal { a, b }) => {
+                match (a.ty(self.interner()), b.ty(self.interner())) {
+                    (Some(a_ty), Some(b_ty)) => {
+                        if a_ty.inference_var(self.interner()).is_some()
+                            && b_ty.inference_var(self.interner()).is_some()
+                        {
+                            return Err(NoSolution);
```
jackh726, line 349: This in particular is different than `chalk-engine`. There, we `Flounder`, here I didn't do all the refactoring to be able to handle this yet.

tests/test/unify.rs
```diff
@@ -333,3 +333,19 @@ fn quantified_types() {
         }
     }
 }
+
+#[test]
+fn subtype() {
```
jackh726, line 338: I honestly have no clue about tests.

chalk-engine/src/logic.rs
nikomatsakis: Flounder, definitely not `NoSolution`. There is a bit of an interesting problem here around cycles -- I have to remember -- rustc does something special to extend the occurs check to detect cycles like `?A <: ?B` and `?B = Vec<?A>`... not sure if we'll have to reproduce that, but we can worry about that once we get to the generalizer.

chalk-integration/src/program.rs
nikomatsakis: I think we want to add explicit annotations probably into the chalk syntax, rather than reproduce rustc's inference.

chalk-ir/src/zip.rs
nikomatsakis: Yeah, irrelevant.

chalk-solve/src/solve/slg/resolvent.rs
```diff
@@ -391,22 +408,28 @@ impl<'i, I: Interner> Zipper<'i, I> for AnswerSubstitutor<'i, I> {
...
-            (TyData::Function(answer), TyData::Function(pending)) => {
-                self.outer_binder.shift_in();
-                Zip::zip_with(self, &answer.substitution, &pending.substitution)?;
-                self.outer_binder.shift_out();
-                Ok(())
-            }
+            (TyData::Function(answer), TyData::Function(pending)) => Zip::zip_with(
+                self,
+                variance,
+                &answer.clone().into_binders(interner),
```
nikomatsakis, line 430: I guess this is accounting for the shifting?

tests/test/unify.rs
nikomatsakis: Well, you might test the region constraints that arise from `&'a u32 <: &'b u32`, for example -- but I can push some more tests later on. If we had the ability to annotate structs with variance etc I guess that'd be better.

chalk-solve/src/infer/test.rs
jackh726: So, rather it being an empty struct, it could be `variances: Variances`, and then each test can decide what the variances that should be returned are.

chalk-engine/src/logic.rs
jackh726: From what I can tell, rustc has a separate unification table for subtypes that is used only for the occurs check.

chalk-parse/src/parser.lalrpop
jackh726: That worked

chalk-solve/src/infer/unify.rs
jackh726: I added a FIXME

chalk-solve/src/recursive/fulfill.rs
jackh726: I just set `self.cannot_solve =true` and continued processing. Probably not the correct thing to do. But I think that will also lead to `Floundering`.

jackh726:
Closing in favor of #609

PR #521, opened 2020-06-12
Recursive minimum

nikomatsakis:
This makes two changes:

* Correct the minimum tracking for negative goals -- actually, I should try to craft a test that shows how the old behavior was wrong
* Pass the minimums using the stack, rather than as an extra variable. The motivation here is that this "minimums" information would eventually move into salsa and we would prefer then for it to be passed "behind the scenes".

### DIFF
--- a/chalk-solve/src/recursive.rs
+++ b/chalk-solve/src/recursive.rs
@@ -129,8 +129,8 @@ impl<'me, I: Interner> Solver<'me, I> {
         // - None < Some(CannotProve)
         // the function which maps the loop iteration to `answer` is a nondecreasing function
         // so this function will eventually be constant and the loop terminates.
+        let minimums = &mut Minimums::new();
         loop {
-            let minimums = &mut Minimums::new();
             let (current_answer, current_prio) = self.solve_iteration(&canonical_goal, minimums);
 
             debug!(
--- a/chalk-solve/src/recursive.rs
+++ b/chalk-solve/src/recursive.rs
@@ -5,7 +5,7 @@ mod search_graph;
 mod solve;
 mod stack;
 
-use self::lib::{Minimums, Solution, UCanonicalGoal};
+use self::lib::{Solution, UCanonicalGoal};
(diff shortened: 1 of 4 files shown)


nikomatsakis:
OK, as I dig in, I'm realizing that the handling of negative goals and recursion in the recursive solver is really not quite right. We're basically mishandling *exactly* the case around negative cycles, e.g. a rule like `A :- not { A }`. The correct answer according to the Well-Formed Semantics here is Ambiguous, I believe -- you can't declare that A is true, because there is no rule that would justify it, but if you say that A is *false*, then there is a rule that justifies A as true.

We actually kind of removed support for these sorts of cycles from the SLG solver, I believe it now propagates back a "floundered" style error of some kind, and perhaps we should do the same in the recursive solver. This is precisely the case that SLG solvers aim to support with "delayed goals". 

As a side note, I'm increasingly of the opinion that I'd like to remove `not` goals altogether but it would require us to address coherence and specialization in some other way.

nikomatsakis:
Current behavior of the recursive solver in that scenario is to make `A` provable. If we fix various bugs in the recursive solver (I have some fixes locally), then we wind up with an infinite cycle:

* You start out assuming A cannot be proven, and you get back that it can
* Then you assume it *can* be proven, iterate, and find out that it *can't*
* So then you assume that it can...

etc. Basically you never converge on a fixed point.

nikomatsakis:
(I have some thoughts on the right fix here and will try to find some time to code it up this weekend; interestingly it definitely impacts the idea of what to do for salsa, since I forgot about the need to handle negative (i.e., non-monotonic) edges within cycles. Pain in the neck.)

nikomatsakis:
I'm going to close this PR and open a new one for the work on fixing how we handle minimums and negative edges. Once that is done, we'll have to figure out how to reconcile those changes with my desire to not thread minimums around explicitly (and in particular with salsa; it definitely adds a wrench in the works).

PR #523, opened 2020-06-13
Add method to get repr data of an ADT to ChalkDatabase

### LINKED ISSUE #517: Add callback to get repr data
To perform some of wf checks, we need to know whether or not struct has `#[repr(C)]` or `#[repr(packed)]` (only those two). So we need to add a database callback to get that info given an adt id and extend the parser to account for it in ChalkDatabase.

This issue has been assigned to @voidc via [this comment](https://github.com/rust-lang/chalk/issues/517#issuecomment-643694192).

voidc:
Hi everyone! I'm Dominik and I'm interested in getting involved in rustc and chalk development. #517  seems like a good place to start and get familiar with the code.
With this WIP PR I would like to get some feedback on whether I understood the issue correctly and this is going in the right direction.

closes #517

### DIFF
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -59,8 +59,13 @@ WellKnownTrait: WellKnownTrait = {
      "#" "[" "lang" "(" "unsize" ")" "]" => WellKnownTrait::Unsize,
 };
 
+StructRepr: StructRepr = {
+     "#" "[" "repr" "(" "C" ")" "]" => StructRepr::C,
+     "#" "[" "repr" "(" "packed" ")" "]" => StructRepr::Packed,
+};
+
 StructDefn: StructDefn = {
-    <upstream:UpstreamKeyword?> <fundamental:FundamentalKeyword?> <phantom_data:PhantomDataKeyword?>
+    <upstream:UpstreamKeyword?> <fundamental:FundamentalKeyword?> <phantom_data:PhantomDataKeyword?> <repr:StructRepr?>
         "struct" <n:Id><p:Angle<VariableKind>>
         <w:QuantifiedWhereClauses> "{" <f:Fields> "}" => StructDefn
     {
@@ -73,6 +78,7 @@ StructDefn: StructDefn = {
             fundamental: fundamental.is_some(),
             phantom_data: phantom_data.is_some(),
         },
+        repr: repr.unwrap_or_default(),
     }
 };
 
--- a/chalk-parse/src/parser.lalrpop
+++ b/chalk-parse/src/parser.lalrpop
@@ -59,11 +59,10 @@ WellKnownTrait: WellKnownTrait = {
      "#" "[" "lang" "(" "unsize" ")" "]" => WellKnownTrait::Unsize,
 };
 
-StructReprC: () = "#" "[" "repr(C)" "]";
-StructReprPacked: () = "#" "[" "repr(packed)" "]";
+StructRepr: Atom = "#" "[" "repr" "(" <name:Id> ")" "]" => name.str;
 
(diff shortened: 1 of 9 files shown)


Areredify:
Good start, couple of things I want to point out:
a) `#[repr(C)]` and `#[repr(packed)]` are not mutually exclusive, so you shouldn't use an enum to represent repr.
b) Storing repr data in `AdtDatum` is what we wanted to avoid by using callbacks to `RustIrDatabase`; you need to store it in `Program` instead.

voidc:
Thanks for the feedback; I revised my PR accordingly. I'm not sure why the tests are failing though.

Areredify:
Hmm, not sure, it doesn't like the letter "C" for some reason

voidc:
I think what is going is that due to [this](https://github.com/rust-lang/chalk/pull/523/files#diff-6fa8bab6c97bead41db277caab4db7a2R61) line the character C becomes a separate token which breaks occurrences of "C" as an identifier. I'm pretty clueless on how to fix this.

Areredify:
I have no idea how lalrpop works internally, @nikomatsakis should know

nathanwhit:
I think that (roughly) what happens is that the identifier `C` becomes a terminal, and so it behaves like a reserved identifier/keyword. It's the same behavior as if, for instance, you declared a struct named `impl` which the parser would reject because "impl" behaves as a reserved keyword. We don't want this for `C`, however, since we only care about `C` (and `packed` for that matter) in the context of a `repr` attribute.

I'll leave suggestions to fix it in a separate comment.

chalk-parse/src/parser.lalrpop
```diff
@@ -58,8 +58,11 @@ WellKnownTrait: WellKnownTrait = {
      "#" "[" "lang" "(" "unsize" ")" "]" => WellKnownTrait::Unsize,
 };
 
+StructReprC: () = "#" "[" "repr" "(" "C" ")" "]";
```
nathanwhit, line 61:
```suggestion
StructReprC: () = "#" "[" "repr(C)" "]";
```

chalk-parse/src/parser.lalrpop
```diff
@@ -58,8 +58,11 @@ WellKnownTrait: WellKnownTrait = {
...
+StructReprPacked: () = "#" "[" "repr" "(" "packed" ")" "]";
```
nathanwhit, line 62:
```suggestion
StructReprPacked: () = "#" "[" "repr(packed)" "]";
```

chalk-parse/src/parser.lalrpop
```diff
@@ -58,8 +58,11 @@ WellKnownTrait: WellKnownTrait = {
      "#" "[" "lang" "(" "unsize" ")" "]" => WellKnownTrait::Unsize,
 };
 
+StructReprC: () = "#" "[" "repr(C)" "]";
+StructReprPacked: () = "#" "[" "repr(packed)" "]";
```
nikomatsakis, line 62: I think for this we should just parse any identifier and then generate a list of identifiers in the AST, perhaps, which lowering can match against "C" or "packed", or else have a `match` statement that examines the identifier in the parser:

```
StructRepr: XXX = "#" "[" "repr" "(" <name:Ident>")" "]" => {
    match name {
        "C" => ...
        "repr" => ...
    }
}
```

voidc: I implemented your suggestion so that StructRepr now parses to an Atom.

jackh726:
I think this looks good. @Areredify do you think we need anything other than `repr(C)` and `repr(packed)`?

Areredify:
Not for the foreseeable future I think

nikomatsakis:
Thanks @voidc!

Issue #506, opened 2020-06-09
refactor how RustIr manages ADTs (and maybe other types)


nikomatsakis:
Currently the `RustIrDatabase` trait has methods that return `Arc<SomeRustIrType>`, but this doesn't map that well to rustc, as it must create these awkward types. We could instead have finer-grained methods. For the case of ADTs, we might have e.g.:

* num_variants(adt_id)
* num_fields(adt_id, variant_index)
* field_type(adt_id, variant_index, field_index)

PR #524, opened 2020-06-14
Support for ADTs

### LINKED ISSUE #505: extend chalk ADTs with information about variants
We have a notion of ADTs but it can't capture the notion of having multiple variants and fields per variant. Part of #368.

This might be best done by also fixing #506.

This issue has been assigned to @doctorn via [this comment](https://github.com/rust-lang/chalk/issues/505#issuecomment-641578908).

doctorn:
This is far from complete (I have practically no tests right now and I haven't really attempted any of the refactoring suggested in #506) however I thought this would be a good point to get some eyes on this as it doesn't break any tests... Any thoughts?

Will eventually resolve #505

### DIFF
--- a/chalk-parse/src/ast.rs
+++ b/chalk-parse/src/ast.rs
@@ -20,7 +20,7 @@ pub struct Program {
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub enum Item {
-    StructDefn(StructDefn),
+    AdtDefn(AdtDefn),
     FnDefn(FnDefn),
     ClosureDefn(ClosureDefn),
     TraitDefn(TraitDefn),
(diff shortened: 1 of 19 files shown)


chalk-parse/src/ast.rs
```diff
@@ -29,19 +29,26 @@ pub enum Item {
...
-    pub flags: StructFlags,
 }
 
 #[derive(Clone, PartialEq, Eq, Debug)]
-pub struct StructFlags {
+pub struct AdtFlags {
     pub upstream: bool,
     pub fundamental: bool,
     pub phantom_data: bool,
+    pub struct_: bool,
```
jackh726, line 51: This should probably be a enum of `Struct, Enum, Union`

chalk-solve/src/clauses/builtin_traits/sized.rs
```diff
@@ -14,22 +14,37 @@ fn push_adt_sized_conditions<I: Interner>(
...
 
     let interner = db.interner();
 
-    // To check if an ADT type S<..> is Sized, we only have to look at its last field.
-    // This is because the WF checks for ADTs require that all the other fields must be Sized.
-    let last_field_ty = adt_datum
+    // To check if an ADT type S<..> is Sized, we only have to look at the last field of each
+    // of its variants. This is because the WF checks for ADTs require that all the other
+    // fields must be Sized.
+    let last_field_tys = adt_datum
```
nathanwhit, line 40: WF checks for enums require that *all* fields of all variants are `Sized`, so all well-formed enums are `Sized`. As a result, here we only need to check if the last field is `Sized` if the ADT is a struct.

chalk-solve/src/rust_ir.rs
```diff
@@ -93,15 +93,24 @@ impl<I: Interner> AdtDatum<I> {
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
 pub struct AdtDatumBound<I: Interner> {
-    pub fields: Vec<Ty<I>>,
+    pub variants: Vec<AdtDatumVariant<I>>,
     pub where_clauses: Vec<QuantifiedWhereClause<I>>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, HasInterner)]
+pub struct AdtDatumVariant<I: Interner> {
```
nathanwhit, line 101: `AdtDatumVariant` reads a bit strangely to me, maybe `AdtVariantDatum`?

chalk-solve/src/rust_ir.rs
```diff
@@ -93,15 +93,24 @@ impl<I: Interner> AdtDatum<I> {
...
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct AdtFlags {
     pub upstream: bool,
     pub fundamental: bool,
     pub phantom_data: bool,
+    /// `true` if the ADT was lowered from a `struct` (this information is required
+    /// for determining if we have built-in `Unsize` impls as `enum`s can never be
+    /// `Unsize`).
+    pub struct_: bool,
```
nathanwhit, line 113: We should use an enum (as described in [this comment](https://github.com/rust-lang/chalk/pull/524/files#r439853744)) here as well.

doctorn:
Rebased - I'm just working on the final tests now

jackh726:
So, we previously talked about merging *last week*. I went ahead and rebased and added the tests and updated for the new display tests. So, I'm gonna go ahead and merge this.

Thanks @doctorn

PR #525, opened 2020-06-14
Introduce Tracing into Chalk (Third Time's the Charm)

nathanwhit:
This PR is a continuation of #409. The changes made here are largely the same as that PR, but rewritten on top of the current `master` and with a few minor adjustments to the macros. I also opted to take the suggestion in [this comment](https://github.com/rust-lang/chalk/pull/409#issuecomment-642771768) and make `chalk-ir` not depend on tracing to potentially ease `rustc` integration.


jackh726:
Awesome @nathanwhit! Can you also add a bit in the book of how to use this? (i.e. if I want to look at the debug output, what would I use instead of `CHALK_DEBUG`?) And theoretically, how would this be different for debugging the rustc integration?

chalk-ir/src/lib.rs
```diff
@@ -51,8 +51,31 @@ pub mod fold;
...
+    #[macro_export]
+    macro_rules! debug_span {
+        ($($t: tt)*) => {
+            let __span = tracing::debug_span!($($t)*);
+            let __span = __span.enter();
+        };
+    }
+
+    #[macro_export]
+    macro_rules! info_span {
```
jackh726, line 64: ...this is only used *once* in `chalk-engine`...

chalk-ir/src/lib.rs
```diff
@@ -51,8 +51,31 @@ pub mod fold;
...
+        ($($t: tt)*) => {
+            let __span = tracing::info_span!($($t)*);
+            let __span = __span.enter();
+        };
+    }
+
+    #[macro_export]
+    macro_rules! trace_span {
```
jackh726, line 72: ...and this isn't used at all.

nathanwhit:
Applied your suggestions and added some documentation!

As for debugging the rustc integration, all you _should_ have to do is add some code to register a "subscriber" to collect the tracing data, like [this](https://github.com/nathanwhit/chalk/blob/9303b01db458531e9c0f1ed05277d46fe16351c0/src/main.rs#L301-L305). You would probably want to place it at an entry point to chalk (and from that point onward logs will be collected), but you could add it anywhere really, as long as it gets executed.

jackh726:
Would that, theoretically, capture other crates' tracing calls? I'll have to test this out, but I overall feel pretty good about landing this. @nikomatsakis what do you think? I would rather land this sooner rather than later, if everything looks good, so it doesn't get stale (again).

nathanwhit:
It *could* potentially capture other crates tracing calls, but only if you register the tracing subscriber as the global default (which is the case for the code I linked). If instead you set a subscriber as a *local* default using [this method](https://docs.rs/tracing/0.1.15/tracing/subscriber/fn.set_default.html) or [this one](https://docs.rs/tracing/0.1.15/tracing/subscriber/fn.with_default.html), like in [this snippet here](https://github.com/nathanwhit/chalk/blob/9303b01db458531e9c0f1ed05277d46fe16351c0/tests/test/mod.rs#L211-L217), then the subscriber would only capture tracing calls that are made transitively within that scope (if that makes sense).

nikomatsakis:
I approve to land, yes.

nikomatsakis:
Let's try it out

PR #531, opened 2020-06-17
Ignore auto traits order

### LINKED ISSUE #510: Support different trait order in trait object unification
Currently when doing unification of trait objects, we check that they are exactly equal, including order of auto traits and associated type clauses. We need to fix that and allow trait objects have any order of where clauses on them. Note that after implementing it some `Unsize` tests should begin to fail, you would need to bless them (and remove "FIXME" comments)

This issue has been assigned to @zaharidichev via [this comment](https://github.com/rust-lang/chalk/issues/510#issuecomment-642025909).

zaharidichev:
This PR aims to fix #510 by sorting the auto traits defs durgin lowering

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -181,7 +181,6 @@ impl<'k> Env<'k> {
     }
 
     fn auto_trait(&self, id: chalk_ir::TraitId<ChalkIr>) -> bool {
-        println!("Looking UP {:?}", id);
         self.auto_traits[&id]
     }
 
@@ -1793,10 +1792,7 @@ impl LowerGoal<LoweredProgram> for Goal {
         let auto_traits = program
             .trait_data
             .iter()
-            .map(|(&trait_id, datum)| {
-                println!("DATA: {:?}", trait_id);
-                (trait_id, datum.flags.auto)
-            })
+            .map(|(&trait_id, datum)| (trait_id, datum.flags.auto))
             .collect();
 
         let fn_def_abis: BTreeMap<_, _> = program
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -26,6 +26,7 @@ type AdtKinds = BTreeMap<chalk_ir::AdtId<ChalkIr>, TypeKind>;
 type FnDefKinds = BTreeMap<chalk_ir::FnDefId<ChalkIr>, TypeKind>;
 type FnDefAbis = BTreeMap<FnDefId<ChalkIr>, <ChalkIr as Interner>::FnAbi>;
 type TraitKinds = BTreeMap<chalk_ir::TraitId<ChalkIr>, TypeKind>;
+type AutoTraits = BTreeMap<chalk_ir::TraitId<ChalkIr>, bool>;
 type OpaqueTyKinds = BTreeMap<chalk_ir::OpaqueTyId<ChalkIr>, TypeKind>;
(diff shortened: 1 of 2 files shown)


tests/test/builtin_impls/unsize.rs
```diff
@@ -76,7 +76,7 @@ fn dyn_to_dyn_unsizing() {
                 dyn Principal + Auto1 + 'a: Unsize<dyn Auto1 + Principal + 'a>
             }
         } yields {
-            "No possible solution"
+            "Unique; substitution [], lifetime constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_0 }]"
```
nikomatsakis, line 79: can you remove the "FIXME" comment on line 72?

tests/test/builtin_impls/unsize.rs
```diff
@@ -68,9 +68,7 @@ fn dyn_to_dyn_unsizing() {
             "Unique; substitution [], lifetime constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_0 }]"
         }
 
-        // Different order of traits in target and source
-        // FIXME: this doesn't work because trait object unification
-        // respects where clause order, which it shouldn't
+        // Different order of traits in target and sourcex
```
nikomatsakis, line 71: Sorry, I didn't mention it, but I think there is maybe a similar comment elsewhere in the file, the other test you edited...?

zaharidichev:
@jackh726 Rustc does it during ast conversion right [here](https://github.com/rust-lang/rust/blob/3ddf48053e83f4949d00dcaf6eb8d9e28fc6bb95/src/librustc_typeck/astconv.rs#L1703). I assumed this would be the equivalent of it. Does that make sense?

tests/test/builtin_impls/unsize.rs
zaharidichev: it just says, "See above", which I think is still relevant, no ?

chalk-integration/src/lowering.rs
```diff
@@ -1753,6 +1789,16 @@ impl LowerGoal<LoweredProgram> for Goal {
                 ((datum.trait_id, datum.name.clone()), lookup)
             })
             .collect();
+
+        let auto_traits = program
+            .trait_data
+            .iter()
+            .map(|(&trait_id, datum)| {
+                println!("DATA: {:?}", trait_id);
```
jackh726, line 1797: Remove this

chalk-integration/src/lowering.rs
```diff
@@ -178,6 +180,11 @@ impl<'k> Env<'k> {
         Err(RustIrError::NotStruct(name.clone()))
     }
 
+    fn auto_trait(&self, id: chalk_ir::TraitId<ChalkIr>) -> bool {
+        println!("Looking UP {:?}", id);
```
jackh726, line 184: And this

Areredify:
I think we should maybe add some tests for trait object unification?

jackh726:
@Areredify what do you mean? Did we miss tests for `dyn` Unsizing?

zaharidichev:
@Areredify is this in tests/unify.rs ?

Areredify:
@jackh726  We didn't, but I would say it's a roundabout wait to test that

jackh726:
Ah, I see. Hmm, I think this is fine. It might be worth adding specific tests for unification of *all* types. But let's do that separately. (Maybe file an issue?)

zaharidichev:
@Areredify I can take the unification tests issue if you provide a pinch of guidance.

jackh726:
Let's do that in a separate PR :)

PR #534, opened 2020-06-17
Change README to be more clear

### LINKED ISSUE #121: README doesn't make it clear exactly what chalk is
> A Prolog-ish interpreter written in Rust, intended perhaps for use in the compiler, but also for experimentation.

This makes it sound like a (relatively) general-purpose logic programming language, perhaps with a fairly Prolog-like syntax, that would be useful for the Rust compiler, and possibly unrelated purposes.

Looking at [`libstd.chalk`](https://github.com/rust-lang-nursery/chalk/blob/master/libstd.chalk) however, shows this is not the case, since it's input is (a subset of) Rust syntax, and not some generic language.

I'm not sure exactly how it should be described, but this should probably be clarified.

jackh726:
Specifically mention it's about traits.

Closes #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@
 
 # chalk
 
-A [Prolog-ish][Prolog] interpreter written in Rust, intended perhaps for use in
-the compiler, but also for experimentation.
+Chalk is a library that implements the Rust trait system, based on [Prolog-ish][Prolog] logic rules.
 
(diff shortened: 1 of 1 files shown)


Mcat12:
@jackh726 Can someone also update the GitHub tagline to match this?

jackh726:
Yeah, I saw that. I can't edit it, but @nikomatsakis can.

PR #535, opened 2020-06-20
Support logs in lowering tests

Mcat12:
Not sure if this was intentional, but now lowering test logs are shown with `CHALK_DEBUG`.


jackh726:
Can we actually export this from `chalk-solve`? That would make it simple to log in rustc.

PR #536, opened 2020-06-20
Document all public items in chalk-ir

detrumi:


chalk-ir/src/debug.rs
```diff
@@ -464,6 +477,7 @@ impl<'a, I: Interner> Debug for SubstitutionDebug<'a, I> {
 }
 
 impl<I: Interner> Substitution<I> {
+    ///
```
bjorn3, line 480: Accidental?

chalk-ir/src/lib.rs
```diff
@@ -13,6 +16,7 @@ use std::marker::PhantomData;
 
 pub use crate::debug::SeparatorTraitRef;
 
+/// uninhabited (empty) type, used in combination with `PhantomData`
```
bjorn3, line 19:
```suggestion
/// Uninhabited (empty) type, used in combination with `PhantomData`.
```

chalk-ir/src/lib.rs
```diff
@@ -63,16 +67,19 @@ pub mod debug;
 /// The set of assumptions we've made so far, and the current number of
 /// universal (forall) quantifiers we're within.
 pub struct Environment<I: Interner> {
+    /// The clauses in the environment.
     pub clauses: ProgramClauses<I>,
 }
 
 impl<I: Interner> Environment<I> {
+    /// Creates a new environment
```
bjorn3, line 75:
```suggestion
    /// Creates a new environment.
```

chalk-ir/src/lib.rs
```diff
@@ -1231,6 +1389,7 @@ pub enum WellFormed<I: Interner> {
     Ty(Ty<I>),
 }
 
+/// Checks wether a type or trait ref can be derived based on things in scope.
```
nathanwhit, line 1392:
```suggestion
/// Checks whether a type or trait ref can be derived from the contents of the environment.
```

chalk-ir/src/lib.rs
```diff
@@ -1388,20 +1552,24 @@ impl<I: Interner> QuantifiedWhereClause<I> {
...
     pub fn new(interner: &I) -> Self {
         Self::from(interner, None::<QuantifiedWhereClause<I>>)
     }
 
+    /// Get the interned quantified where clauses.
     pub fn interned(&self) -> &I::InternedQuantifiedWhereClauses {
         &self.interned
     }
 
+    /// Creates a list of quantified where clauses using an iterator.
```
nathanwhit, line 1572:
```suggestion
    /// Creates a list of quantified where clauses from an iterator.
```

chalk-ir/src/lib.rs
```diff
@@ -1415,6 +1583,7 @@ impl<I: Interner> QuantifiedWhereClauses<I> {
         .unwrap()
     }
 
+    /// Tries to create a list of quantified where clauses using an iterator.
```
nathanwhit, line 1586:
```suggestion
    /// Tries to create a list of quantified where clauses from an iterator.
```

chalk-ir/src/lib.rs
```diff
@@ -1713,14 +1899,23 @@ where
 /// conditions.
 #[derive(Clone, PartialEq, Eq, Hash, Fold, Visit, HasInterner, Zip)]
 pub struct ProgramClauseImplication<I: Interner> {
+    /// The consequense of the clause, which holds when the conditions hold
```
nathanwhit, line 1902:
```suggestion
    /// The consequence of the clause, which holds if the conditions hold.
```

chalk-ir/src/lib.rs
```diff
@@ -1713,14 +1899,23 @@ where
...
     pub consequence: DomainGoal<I>,
+
+    /// The condition goals that should hold.
     pub conditions: Goals<I>,
+
+    /// The implication  priority
```
nathanwhit, line 1908:
```suggestion
    /// The priority of the implication.
```

chalk-ir/src/lib.rs
```diff
@@ -1816,6 +2024,7 @@ impl<I: Interner> ProgramClauses<I> {
         .unwrap()
     }
 
+    /// Tries to create a list of program clauses using an iterator.
```
nathanwhit, line 2027:
```suggestion
    /// Tries to create a list of program clauses from an iterator.
```

chalk-ir/src/lib.rs
```diff
@@ -803,53 +871,71 @@ impl<I: Interner> Const<I> {
     }
 }
 
+/// Constant data, containing the type it's known by and its value.
```
nathanwhit, line 874:
```suggestion
/// Constant data, containing the constant's type and value.
```

chalk-ir/src/lib.rs
```diff
@@ -2198,14 +2458,30 @@ pub enum GoalData<I: Interner> {
...
+
+    /// Existential quantifier `Exists`.
+    ///
+    /// A formula with the existential quantifier `exists(x). P(x)` is satisfiable
+    /// if and only if there exists at least one value for all possible values of x
+    /// which satisfies the subformula `P(x)`.
+
+    /// In the context of chalk, the existential quantifier usually demands the
+    /// existence of exactly one instance (i.e. type) that satisfies the formula
+    /// (i.e. type constraints). More than one instance means that the result is ambiguous.
```
jackh726, line 2484: Mmm, this is...somewhat correct. If using the slg solver with `solve_multiple`, you'll get back multiple answers. But this is right in that it *does* say "usually"

chalk-ir/src/interner.rs
```diff
@@ -227,6 +231,13 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
         None
     }
 
+    /// Prints the debug representation of a function-def-id. To get good
+    /// results, this requires inspecting TLS, and is difficult to
```
jackh726, line 235: This isn't a comment *this* PR exactly (since this just matches other functions). But I honestly would rather we no mention TLS in all of these functions.

chalk-ir/src/lib.rs
```diff
@@ -2385,16 +2675,18 @@ where
     }
 }
 
+/// Utility for converting a list of all the binders into scope
+/// into references to those binders. Simply pair the binders with
+/// the indices, and invoke `to_generic_arg()` on the `(binder,
```
jackh726, line 2680: Another comment not specific to this PR: I feel like this should have been `(index, binder)` so that `variable_kinds.enumerate()` can be casted directly.

chalk-ir/src/lib.rs
```diff
@@ -2461,13 +2753,26 @@ impl<'i, I: Interner> Folder<'i, I> for &SubstFolder<'i, I> {
 /// additionally be solved.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Fold, Visit, HasInterner)]
 pub struct ConstrainedSubst<I: Interner> {
-    pub subst: Substitution<I>, /* NB: The `is_trivial` routine relies on the fact that `subst` is folded first. */
+    /// The substitution that is being constrained.
+    ///
+    /// NB: The `is_trivial` routine relies on the fact that `subst` is folded first.
+    pub subst: Substitution<I>, /*  */
```
jackh726, line 2759: Extra comments now

nikomatsakis:
Can we enable the lint that warns about public items w/o doc comments, too?

chalk-ir/src/lib.rs
detrumi: Might not have been the best idea to copy this part from the book then (though it's sometimes tricky to decide what should go where)

jackh726: I think this is fine, the "usually" makes it so.

jackh726:
@nikomatsakis that was added :)

chalk-ir/src/interner.rs
detrumi: Yeah, and it adds a lot of noise to the source, even more so since there's 25 of them. I'll just remove the part about TLS.

detrumi:
> Can we enable the lint that warns about public items w/o doc comments, too?

Ah yes, should've mentioned that I enabled that for `chalk-ir`.
I don't think it makes as much sense for the other crates at the moment, but for `chalk-ir` it's a great fit.

chalk-ir/src/lib.rs
detrumi: Heh, good catch. I glanced right over that the first 5 times I read your comment :laughing:

chalk-ir/src/lib.rs
detrumi: Good idea, I'll just include it.

PR #538, opened 2020-06-21
Fix coherence issue with associated types in generic bound

### LINKED ISSUE #515: Coherence issue with overlapping imps with assoc types
Given 
```
trait Iterator {
    type Item;
}

trait Trait {}

impl<T> Trait for T where T: Iterator<Item = u32> {} 
impl<T> Trait for T where T: Iterator<Item = u32> {} 

struct Foo {}
```
we don't throw a coherence error. Adding `impl Iterator for Foo { type Item = u32; }` makes it appear, but it should work without it too.

Mcat12:
The normalize clause needs to be simplified, and may not be the right approach, but it works.

Fixes #515

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -364,18 +364,18 @@ fn program_clauses_that_could_match<I: Interner>(
                     )?;
                 }
 
+                push_program_clauses_for_associated_type_values_in_impls_of(
+                    builder,
+                    trait_id,
+                    trait_parameters,
+                );
+
                 push_clauses_for_compatible_normalize(
                     db,
                     builder,
                     interner,
                     trait_id,
                     proj.associated_ty_id,
-                );
-
-                push_program_clauses_for_associated_type_values_in_impls_of(
-                    builder,
-                    trait_id,
-                    trait_parameters,
                 );
             }
             AliasTy::Opaque(_) => (),
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -389,10 +389,12 @@ fn program_clauses_that_could_match<I: Interner>(
 /// Adds clauses to allow normalizing possible downstream associated type
 /// implementations when in the "compatible" mode. Example clauses:
 ///
-///     for<type, type, type> Normalize(<^0.0 as Trait<^0.1>>::Item -> ^0.2)
-///         :- Compatible, Implemented(^0.0: Trait<^0.1>), DownstreamType(^0.1), CannotProve
-///     for<type, type, type> Normalize(<^0.0 as Trait<^0.1>>::Item -> ^0.2)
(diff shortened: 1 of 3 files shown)


chalk-solve/src/clauses.rs
```diff
@@ -377,6 +384,57 @@ fn program_clauses_that_could_match<I: Interner>(
...
+/// forall<T1, T2> Normalize(<T1 as Trait>::Item -> T2)
+///     :- Implemented(T1: Trait), Compatible, DownstreamType(T1), CannotProve
+fn push_clauses_for_compatible_normalize<I: Interner>(
+    builder: &mut ClauseBuilder<'_, I>,
+    interner: &I,
+    trait_id: TraitId<I>,
+    associated_ty_id: AssocTypeId<I>,
+) {
+    // FIXME: Simplify?
+    builder.push_binders(
```
Mcat12, line 398: Building this clause is long and cumbersome. Is there a better way? Most clauses seem to re-use binders, but I'm not sure if that is possible here.

chalk-solve/src/clauses.rs
```diff
@@ -377,6 +384,57 @@ fn program_clauses_that_could_match<I: Interner>(
     Ok(clauses)
 }
 
+/// Adds the clause:
+///
+/// forall<T1, T2> Normalize(<T1 as Trait>::Item -> T2)
+///     :- Implemented(T1: Trait), Compatible, DownstreamType(T1), CannotProve
```
Mcat12, line 390: I'm not sure if this is the right way to fix this coherence problem, but it seems to be working.

chalk-solve/src/clauses.rs
```diff
@@ -377,6 +384,57 @@ fn program_clauses_that_could_match<I: Interner>(
     Ok(clauses)
 }
 
+/// Adds the clause:
+///
+/// for<type, type> Normalize(<^0.0 as Trait>::Item -> ^0.1)
+///     :- Implemented(^0.0: Trait), Compatible, DownstreamType(^0.0), CannotProve
```
nikomatsakis, line 390: Ok, so, I think this rule is right in spirit, but it won't work for more complex traits like

```rust
trait Foo<T> {
    type Bar;
}
```

Here, we would need something like

```
for<A, B, C> {
    Normalize(<A as Foo<B>>::Bar -> C) if
        Compatible, Implemented(A: Foo<B>), DownstreamType(A), CannotProve
}

for<A, B, C> {
    Normalize(<A as Foo<B>>::Bar -> C) if
        Compatible, Implemented(A: Foo<B>), FullyVisible(A), DownstreamType(B), CannotProve
}
```

nikomatsakis: Compare to this logic

https://github.com/rust-lang/chalk/blob/b8a90d8122ec0589969c89b938777c8baf58be5b/chalk-solve/src/clauses/program_clauses.rs#L607-L623

Mcat12: I think the latest commit fixes this.

PR #539, opened 2020-06-22
Add more Copy implementations

matthewjasper:
This is mainly so that `QuantifiedWhereClause` can be put in a
`rustc_middle::ty::List`, the others are added in case they're useful.


jackh726:
@matthewjasper can you rebase?

PR #541, opened 2020-06-23
Fix built-in Fn impls when generics are involved

flodiebold:
There were two problems:
 - the FnDef impl was using a wrong substitution for the FnDef (calling
   `builder.substitution_in_scope()`, which wasn't even necessarily the same
   number of parameters -- not sure what the intention was there)
 - when looking for `Normalize` clauses, the self type wasn't generalized (so
   bound variables were handled wrongly). (This lead to crashes when trying to
   integrate it in rust-analyzer: rust-analyzer/rust-analyzer#4982)

### DIFF
--- a/tests/test/fn_def.rs
+++ b/tests/test/fn_def.rs
@@ -96,5 +96,73 @@ fn fn_def_implements_fn_traits() {
         } yields {
             "Unique"
         }
+
+        goal {
+            bar: Fn<(i32,)>
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            Normalize(<bar as FnOnce<(i32,)>>::Output -> ())
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            baz: Fn<(i32,)>
+        } yields {
+            "Unique"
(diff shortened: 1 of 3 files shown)


tests/test/fn_def.rs
```diff
@@ -96,5 +96,73 @@ fn fn_def_implements_fn_traits() {
...
+
+        goal {
+            forall<T> { foo<T>: Fn<(T,)> }
+        } yields {
+            "Unique"
+        }
+
+        goal {
+            exists<T> { Normalize(<foo<T> as FnOnce<(T,)>>::Output -> T) }
+        } yields[SolverChoice::recursive()] {
```
jackh726, line 158: Is this associated type ambiguity for SLG? If so, can you add a comment (next to the wrong output)?

flodiebold: Oh hm actually I just added that while debugging, let me check whether it fails.

flodiebold: No, works fine, I've removed it.

PR #543, opened 2020-06-27
Incorporate region constraints into program clauses

### LINKED ISSUE #508: refactor how "lifetime outlives" goals are integrated into the solver

nathanwhit:
Closes #508

This PR attempts the refactoring proposed in #508, which is to incorporate region constraints into program clauses.

I haven't been able to find a good way to simplify unification with this approach, which is a bit of a downside compared to the current approach.

### DIFF
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -8,7 +8,7 @@
 use crate::{CompleteAnswer, ExClause};
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    AnswerSubst, Binders, Canonical, ConstrainedSubst, Constraint, DomainGoal, Environment,
+    AnswerSubst, Binders, Canonical, ConstrainedSubst, Constraints, DomainGoal, Environment,
     Fallible, Floundered, GenericArg, Goal, InEnvironment, ProgramClause, ProgramClauses,
     Substitution, UCanonical, UniverseMap,
 };
@@ -94,7 +94,7 @@ pub trait ContextOps<I: Interner, C: Context<I>>: Sized + Clone + Debug {
     ) -> (
         C::InferenceTable,
         Substitution<I>,
-        Vec<InEnvironment<Constraint<I>>>,
+        Constraints<I>,
         Vec<InEnvironment<Goal<I>>>,
     );
 
@@ -181,7 +181,7 @@ pub trait UnificationOps<I: Interner, C: Context<I>> {
         &mut self,
         interner: &I,
         subst: Substitution<I>,
-        constraints: Vec<InEnvironment<Constraint<I>>>,
+        constraints: Constraints<I>,
     ) -> Canonical<ConstrainedSubst<I>>;
 
     // Used by: logic
@@ -189,7 +189,7 @@ pub trait UnificationOps<I: Interner, C: Context<I>> {
         &mut self,
         interner: &I,
         subst: Substitution<I>,
-        constraints: Vec<InEnvironment<Constraint<I>>>,
+        constraints: Constraints<I>,
         delayed_subgoals: Vec<InEnvironment<Goal<I>>>,
     ) -> Canonical<AnswerSubst<I>>;
 
--- a/chalk-engine/src/context.rs
+++ b/chalk-engine/src/context.rs
@@ -8,7 +8,7 @@
 use crate::{CompleteAnswer, ExClause};
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    AnswerSubst, Binders, Canonical, ConstrainedSubst, Constraints, DomainGoal, Environment,
+    AnswerSubst, Binders, Canonical, ConstrainedSubst, Constraint, DomainGoal, Environment,
     Fallible, Floundered, GenericArg, Goal, InEnvironment, ProgramClause, ProgramClauses,
     Substitution, UCanonical, UniverseMap,
 };
@@ -94,7 +94,7 @@ pub trait ContextOps<I: Interner, C: Context<I>>: Sized + Clone + Debug {
     ) -> (
         C::InferenceTable,
         Substitution<I>,
-        Constraints<I>,
+        Vec<InEnvironment<Constraint<I>>>,
         Vec<InEnvironment<Goal<I>>>,
     );
 
@@ -181,7 +181,7 @@ pub trait UnificationOps<I: Interner, C: Context<I>> {
         &mut self,
         interner: &I,
         subst: Substitution<I>,
-        constraints: Constraints<I>,
+        constraints: Vec<InEnvironment<Constraint<I>>>,
     ) -> Canonical<ConstrainedSubst<I>>;
 
     // Used by: logic
@@ -189,7 +189,7 @@ pub trait UnificationOps<I: Interner, C: Context<I>> {
         &mut self,
         interner: &I,
         subst: Substitution<I>,
-        constraints: Constraints<I>,
+        constraints: Vec<InEnvironment<Constraint<I>>>,
         delayed_subgoals: Vec<InEnvironment<Goal<I>>>,
     ) -> Canonical<AnswerSubst<I>>;
 
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -2558,9 +2558,6 @@ pub enum GoalData<I: Interner> {
     /// proven via program clauses
     DomainGoal(DomainGoal<I>),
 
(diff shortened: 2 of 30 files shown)


chalk-ir/src/lib.rs
```diff
@@ -2009,6 +2009,9 @@ pub struct ProgramClauseImplication<I: Interner> {
     /// The condition goals that should hold.
     pub conditions: Goals<I>,
 
+    /// The lifetime constraints that should be proven.
+    pub constraints: Vec<InEnvironment<Constraint<I>>>,
```
jackh726, line 2013: Can you make this an interned type?

chalk-solve/src/infer/unify.rs
```diff
@@ -472,14 +478,15 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
 
-    fn push_lifetime_eq_subgoal(&mut self, a: Lifetime<I>, b: Lifetime<I>) {
-        let interner = self.interner;
-        let b_outlives_a = GoalData::AddRegionConstraint(b.clone(), a.clone()).intern(interner);
-        self.goals
-            .push(InEnvironment::new(self.environment, b_outlives_a));
-        let a_outlives_b = GoalData::AddRegionConstraint(a, b).intern(interner);
-        self.goals
-            .push(InEnvironment::new(self.environment, a_outlives_b));
+    fn push_lifetime_eq_constraint(&mut self, a: Lifetime<I>, b: Lifetime<I>) {
```
nikomatsakis, line 481: This setup is fine, but I think that the idea would be to push a [`LifetimeOutlives`](http://rust-lang.github.io/chalk/chalk_ir/struct.LifetimeOutlives.html) domain goal, and that we would have chalk-solve have a rule that says "to prove `LifetimeOutlives('a, 'b)`, add the region constraints that `'a: 'b`"

jackh726: Hmm, yeah that's actually probably a better way to approach this.

nathanwhit:
Just pushed some new changes to make the constraints an interned type, and to apply @nikomatsakis's suggestion for factoring constraints out of unification.

I did run into a problem though: after changing unification to emit `LifetimeOutlives` goals instead of constraints, a few of the tests started failing for (only) the recursive solver. From what I can tell the issue is that in the recursive solver when we try to solve an outlives goal with a bound lifetime like `'^0_0 : 'a` we find program clauses for the goal [before instantiating it](https://github.com/rust-lang/chalk/blob/2bd06ea1cac18cc83958667122d3d4488f6a9835/chalk-solve/src/recursive/solve.rs#L76), which yields a program clause with the [bound variable shifted in](https://github.com/rust-lang/chalk/blob/2bd06ea1cac18cc83958667122d3d4488f6a9835/chalk-solve/src/clauses/builder.rs#L80-L85), `^1_0 : 'a`. We then try to solve that program clause derived from the non-instantiated goal `'0_0 : 'a` in an effort to solve the instantiated goal `?0 : 'a`, and fail when trying to instantiate the program clause [here](https://github.com/rust-lang/chalk/blob/2bd06ea1cac18cc83958667122d3d4488f6a9835/chalk-solve/src/recursive/fulfill.rs#L173-L179) because the assertion [here](https://github.com/rust-lang/chalk/blob/2bd06ea1cac18cc83958667122d3d4488f6a9835/chalk-ir/src/lib.rs#L2853) fails.

The most recent commit 8dfa89af0aa2533afc9d53e91fc1ec82504b239e fixes the test failures, but it doesn't feel like the *right* fix, so any thoughts/feedback would be helpful.

chalk-engine/src/context.rs
```diff
@@ -94,7 +94,7 @@ pub trait ContextOps<I: Interner, C: Context<I>>: Sized + Clone + Debug {
     ) -> (
         C::InferenceTable,
         Substitution<I>,
-        Vec<InEnvironment<Constraint<I>>>,
+        Constraints<I>,
```
jackh726, line 97: I feel like it isn't worth using the interned `Constraints` here. We don't really intern anything else in `chalk-engine`. Maybe we should, but for consistency, let's just leave this as a `Vec`

jackh726:
So, I haven't looked much into it, but I think this is related to the problem we run into with the builtin traits. For that we use `generalize` (https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/clauses/builtin_traits.rs#L21). But I honestly still haven't looked at the recursive solver a ton. So cc @flodiebold and @nikomatsakis

chalk-ir/src/lib.rs
```diff
@@ -2629,6 +2630,79 @@ pub enum Constraint<I: Interner> {
...
+            interned: I::intern_constraints(interner, constraints.into_iter().casted(interner))?,
+        })
+    }
+
+    /// Get the interned representation of the constraints.
+    pub fn interned(&self) -> &I::InternedConstraints {
+        &self.interned
+    }
+
+    /// Index into the list of constraints.
```
nikomatsakis, line 2670: We gotta find a way to remove this boilerplate

chalk-solve/src/recursive/fulfill.rs
```diff
@@ -173,11 +174,17 @@ impl<'s, I: Interner, Solver: SolveDatabase<I>, Infer: RecursiveInferenceTable<I
         let ProgramClauseImplication {
             consequence,
             conditions,
+            constraints,
             priority: _,
         } = fulfill
             .infer
             .instantiate_binders_existentially(fulfill.solver.interner(), clause);
```
nikomatsakis, line 181: Hmm I suspect the problem is that somehow this is not instantiating the `^1` in the program clause constraints...

chalk-solve/src/recursive/solve.rs
```diff
@@ -68,12 +69,12 @@ pub(super) trait SolveIteration<I: Interner>: SolveDatabase<I> {
                 // or from the lowered program, which includes fallback
                 // clauses. We try each approach in turn:
 
-                let InEnvironment { environment, goal } = &canonical_goal.canonical.value;
-
+                let (_, _, InEnvironment { environment, goal }) =
+                    self.new_inference_table(&canonical_goal);
```
nikomatsakis, line 73: What motivated this change?

chalk-solve/src/clauses.rs
```diff
@@ -314,12 +314,15 @@ fn program_clauses_that_could_match<I: Interner>(
                 .to_program_clauses(builder),
         },
         DomainGoal::Holds(WhereClause::LifetimeOutlives(LifetimeOutlives { a, b })) => {
-            builder.push_clause(
-                WhereClause::LifetimeOutlives(LifetimeOutlives {
+            builder.push_fact_with_constraints(
```
nikomatsakis, line 317: The problem with the recursive solver is that this is generating an overly specific rule. It is creating a rule that is **specific** to a, b, but we can make a more general rule like `forall<a, b> { Outlives('a: 'b) :- [] && Constraint('a: 'b)` -- sorry, not sure of the notation. =)

We do need to add a helper analogous to [`ClauseBuilder::push_bound_ty`](http://rust-lang.github.io/chalk/chalk_solve/clauses/builder/struct.ClauseBuilder.html#method.push_bound_ty) but for lifetimes. Then we can do something like this:

```rust
builder.push_bound_lifetime(|builder, a| {
  builder.push_bound_lifetime(|builder, b| {
    builder.push_fact_with_constraints(
      LifetimeOutlives { a, b },
      Some(/* new constraint that a: b */)
    )
  })
})
```

chalk-ir/src/lib.rs
nathanwhit: Definitely agree, we have basically this exact code duplicated for at least 4 other types. I'll open an issue for it.

chalk-solve/src/recursive/solve.rs
nathanwhit: It was basically a result of noticing that in the SLG solver (which worked properly), but not the recursive solver, we instantiate goals before finding relevant program clauses, then wondering if replicating that might work here. It (surprisingly) did fix all the test failures, but it wasn't really driven by a strong intuition.

chalk-solve/src/clauses.rs
nathanwhit: Ah I see, that makes sense. Thanks for the help!

PR #544, opened 2020-06-28
Use tracing-tree for logging / clean up debug output

nathanwhit:
This PR makes it so we use the `tracing-tree` crate for logging output. This adds indentation to the logs, and generally makes the logging output much more readable compared to the current state. 

I also cleaned up many of the logging calls to properly use the `tracing` macros (e.g. create a new span when entering a function instead of just emitting an event).

### DIFF
--- a/chalk-solve/src/logging.rs
+++ b/chalk-solve/src/logging.rs
@@ -7,6 +7,6 @@ pub fn with_tracing_logs<T>(action: impl FnOnce() -> T) -> T {
     let filter = EnvFilter::from_env("CHALK_DEBUG");
     let subscriber = Registry::default()
         .with(filter)
-        .with(HierarchicalLayer::new(4));
+        .with(HierarchicalLayer::new(2));
     tracing::subscriber::with_default(subscriber, action)
 }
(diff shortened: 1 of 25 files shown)


chalk-solve/src/logging.rs
```diff
@@ -1,13 +1,12 @@
...
 pub fn with_tracing_logs<T>(action: impl FnOnce() -> T) -> T {
     let filter = EnvFilter::from_env("CHALK_DEBUG");
-    let subscriber = FmtSubscriber::builder()
-        .with_env_filter(filter)
-        .with_ansi(false)
-        .without_time()
-        .finish();
+    let subscriber = Registry::default()
+        .with(filter)
+        .with(HierarchicalLayer::new(4));
```
Mcat12, line 10: Especially with the recursive solver, we might want only 2 space indentation.

nikomatsakis:
That said, 👍 to merging, I trust this makes sense even if I don't et it yet

nathanwhit:
> Woah, stuff like `debug!(?foo)` is surprising to me. Where can I read up on that?

It definitely does take a bit of getting used to, the [tracing docs](https://docs.rs/tracing/0.1.15/tracing) are quite helpful in explaining it. The macro syntax specifically is covered [here](https://docs.rs/tracing/0.1.15/tracing/#using-the-macros).

(As a side note, I've been meaning to add some documentation in the `chalk` book about using the macros, it seems like it would be good to have some of the basics covered there with a link to the full docs.)

PR #545, opened 2020-06-28
Cleanup function names in `chalk-ir`

### LINKED ISSUE #530: Make function names in `chalk_ir` more uniform
Right now, the naming of function names in `chalk_ir` is a little bit non-uniform. For example:,
* `Substitution` has `from` and `from_fallible` for an `IntoIterator`, `from1` for a single, and `empty` for nothing
* `Binders` has `new` for a given `VariableKinds` and `empty` for no bound vars
* `VariableKinds` has `from` and `from_fallible` for an `IntoIterator`, `new` for nothing, and no `from1`
* `CanonicalVarKinds` is similar to `VariableKinds`
* `Goals` is also similar to the above two

Additionally `Substitution` uses `parameters`, `Binders` doesn't have a way to access the inner `Binders`, and the others use `as_slice`.

This is honestly a bit of a pain point since I always seem to use the wrong function name.

nathanwhit:
Closes #530.

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -2166,7 +2166,7 @@ pub struct VariableKinds<I: Interner> {
 }
 
 impl<I: Interner> VariableKinds<I> {
-    /// Creates an empty list of canonical variable kinds.
+    /// Creates an empty list of variable kinds.
     pub fn empty(interner: &I) -> Self {
         Self::from(interner, None::<VariableKind<I>>)
     }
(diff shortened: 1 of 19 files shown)


chalk-ir/src/lib.rs
```diff
@@ -1803,8 +1802,8 @@ impl<T: HasInterner> Binders<T> {
     /// Wraps the given value in a binder without variables, i.e. `for<>
     /// (value)`. Since our deBruijn indices count binders, not variables, this
     /// is sometimes useful.
-    pub fn empty(interner: &T::Interner, value: T) -> Self {
-        let binders = VariableKinds::new(interner);
+    pub fn without_vars(interner: &T::Interner, value: T) -> Self {
```
jackh726, line 1805: I think I actually like `empty` here. Or `dummy` if we want to match rustc.

nikomatsakis: I agree that `without_vars` wasn't very clear to me.

PR #549, opened 2020-07-01
Feature gate heavy `tracing` dependencies

nathanwhit:
There were some [concerns raised](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/tracing-tree/near/202540840) about chalk's dependency on `tracing`, `tracing-subscriber`, and `tracing-tree`, particularly in how they might affect the compilation time and binary size of rust-analyzer.

This PR experimentally gates `tracing-subscriber`, `tracing-tree`, and the `attributes` feature of `tracing` behind a new feature `tracing-full`. The handling of the `instrument` attribute macro from `tracing` (which only works with the `attributes` feature enabled) is in particular a bit hacky. In order to avoid manually adding `cfg_attr` to each usage of the `instrument` macro, this PR creates a wrapper macro exported from `chalk-derive` which does so.

### DIFF
--- a/chalk-solve/Cargo.toml
+++ b/chalk-solve/Cargo.toml
@@ -14,8 +14,8 @@ ena = "0.14.0"
 itertools = "0.9.0"
 petgraph = "0.5.0"
 tracing = "0.1"
-tracing-subscriber = "0.2"
-tracing-tree = "0.1.1"
+tracing-subscriber = { version = "0.2", optional = true }
+tracing-tree = { version = "0.1.1", optional = true }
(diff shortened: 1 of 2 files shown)


nathanwhit:
> I'm okay with the `tracing-full` feature flag, but I don't think we want to gate `instrument` at all.
I looked into this a bit more and indeed gating `instrument` doesn't seem to have much of a benefit in terms of compilation time or binary size (which makes sense, the only heavy dependency it uses is `syn` which we already depend on).

Just pushed some changes so now we don't gate `instrument`.

PR #551, opened 2020-07-02
Add type outlives goal

### LINKED ISSUE #509: Add type outlives goal to Unsize
In builtin impls for `Unsize` for `T` -> `dyn Trait + 'a` unsizing we need to check that `T` outlives `'a`, i.e. use `TypeOutlives(T, 'a)` which we don't currently have in chalk. So this is a tracking issue to add it once type outlives part of #435 lands.

nathanwhit:
Addresses part of #435. Closes #509.

### DIFF
--- a/tests/test/builtin_impls/unsize.rs
+++ b/tests/test/builtin_impls/unsize.rs
@@ -192,7 +192,7 @@ fn ty_to_dyn_unsizing() {
                 Foo: Unsize<dyn Principal + Auto + 'a>
             }
         } yields {
-            "Unique; substitution [], lifetime constraints []"
+            "Unique; substitution [], lifetime constraints [InEnvironment { environment: Env([]), goal: Foo: '!1_0 }]"
         }
 
         // Principal is not the first trait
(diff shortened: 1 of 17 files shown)


jackh726:
Can we also address this `FIXME` in this PR: 
https://github.com/rust-lang/chalk/blob/master/tests/test/builtin_impls/unsize.rs#L217

Issue is here: https://github.com/rust-lang/chalk/issues/509

I'm somewhat okay if we don't. But, would be nice.

nathanwhit:
> Can we also address this `FIXME` in this PR:
> https://github.com/rust-lang/chalk/blob/master/tests/test/builtin_impls/unsize.rs#L217
> 
> Issue is here: #509
> 
> I'm somewhat okay if we don't. But, would be nice.

Done!

tests/test/builtin_impls/unsize.rs
```diff
@@ -210,7 +210,7 @@ fn ty_to_dyn_unsizing() {
                 Foo: Unsize<dyn Auto + 'a>
             }
         } yields {
-            "Unique; substitution [], lifetime constraints []"
+            "Unique; substitution [], lifetime constraints [InEnvironment { environment: Env([]), goal: Foo: '!1_0 }]"
         }
 
         // TypeOutlives test
```
jackh726, line 216: Can you remove the `FIXME` the line below this?

Issue #550, opened 2020-07-01
Reduce boilerplate for interned type definitions


nathanwhit:
Right now we repeat code [like this](https://github.com/rust-lang/chalk/blob/e7d8893231cda2ea18099d3e0e90d780925ff0b9/chalk-ir/src/lib.rs#L2672-L2705) in at least 4-5 places across `chalk-ir`. This not only creates bloat, making the definitions in`chalk-ir` harder to read, but also makes adding new interned types very tedious. We should find a way to reduce the amount of code duplication and boilerplate here.

nikomatsakis:
Macros or derive might be an option here -- perhaps a trait for interned sequences? Something like that.

nathanwhit:
Closed by #554.

PR #552, opened 2020-07-02
Reduce boilerplate for interned sequence types

nathanwhit:
This is a first attempt at #550. It eliminates ~200 LOC from `chalk_ir/lib.rs`. This PR adds a new trait `Sequence`, which contains the duplicated logic. The downside of this is that it requires bringing the trait in scope in order to use the methods. The second downside is that we currently use a `from` method for creating sequence types, so when moved into a trait it creates a conflict with `From::from`. I chose to rename `from` to `from_iter` to avoid this, but suggestions are welcome.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -12,7 +12,7 @@ use crate::{
 
 use chalk_ir::interner::Interner;
 use chalk_ir::{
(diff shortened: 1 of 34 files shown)


jackh726:
An alternative to this would be to use a macro?

PR #553, opened 2020-07-02
Upgrade dependencies (incl. Salsa) 

nathanwhit:
This PR upgrades all dependencies, including upgrading salsa to `0.14`. To upgrade salsa, there were some changes needed to account for the removal of the `volatile` attribute, so it might be good if someone who knows salsa a bit better double checked this.

### DIFF
--- a/chalk-integration/src/query.rs
+++ b/chalk-integration/src/query.rs
@@ -17,7 +17,10 @@ use chalk_solve::RustIrDatabase;
 use chalk_solve::Solver;
 use chalk_solve::SolverChoice;
 use salsa::Database;
+use std::clone::Clone;
+use std::cmp::{Eq, PartialEq};
 use std::collections::BTreeMap;
+use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 use std::sync::Mutex;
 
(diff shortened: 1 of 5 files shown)


chalk-integration/src/query.rs
```diff
@@ -50,7 +51,7 @@ pub trait LoweringDatabase: RustIrDatabase<ChalkIr> {
     /// cached state becomes invalid, so the query is marked as
     /// volatile, thus ensuring that the solver is recreated in every
     /// revision (i.e., each time source program changes).
-    #[salsa::volatile]
+    #[salsa::dependencies]
```
nikomatsakis, line 54: Hmm, I don't know that this has quite the same semantics. I think this is going to create a fresh solver every time. This was an issue that was raised over on the salsa repo -- that we lost a way to have values that are not `Eq`.

detrumi: So is there a way to work around this, or are we stuck waiting for a solution to the salsa issue?

nathanwhit: > Hmm, I don't know that this has quite the same semantics. I think this is going to create a fresh solver every time. This was an issue that was raised over on the salsa repo -- that we lost a way to have values that are not `Eq`.

Ah yes, you're right. I was hoping they had the same semantics going off [their descriptions](https://salsa-rs.github.io/salsa/rfcs/RFC0001-Query-Group-Traits.html#controlling-query-modes), but looking at the source code they definitely aren't the same.

> So is there a way to work around this, or are we stuck waiting for a solution to the salsa issue?

We could use a hacky workaround (which @nikomatsakis briefly alluded to). I just pushed some changes that *should* work, but it's not ideal.

nathanwhit: > Hmm, I don't know that this has quite the same semantics. I think this is going to create a fresh solver every time. This was an issue that was raised over on the salsa repo -- that we lost a way to have values that are not `Eq`.

Ah yes, you're right. I was hoping they had the same semantics going off [their descriptions](https://salsa-rs.github.io/salsa/rfcs/RFC0001-Query-Group-Traits.html#controlling-query-modes), but looking at the source code they definitely aren't the same.

> So is there a way to work around this, or are we stuck waiting for a solution to the salsa issue?

We could use a hacky workaround (which @nikomatsakis briefly alluded to). I just pushed some changes that *should* work, but it's not ideal.

PR #554, opened 2020-07-03
Further reduce interned type boilerplate

detrumi:
Builds on top of #550, using a macro to generate the sequence structs, `interned` methods and `Sequence` impls.


jackh726:
This is more of what I was expecting. But if we do this, it's easy enough to just generate the `Sequence` functions in the macro. 

I guess this is a fundamental tradeoff: would we rather have more "code" or use traits to share "code". The trait route would result in less code generation, but requires we bring the trait into scope.

As a side point, I don't really like `Sequence`, I would instead prefer something like `InternedSlice`

PR #555, opened 2020-07-03
Don't flounder on int/float vars

flodiebold:
If the self type in our goal is e.g. an int var, we don't need to flounder even if the trait is non-enumerable because the search can still be restricted to a few types (rustc can do this already).

This works with the SLG solver, but the recursive solver passes goals to the program clauses code before instantiating them, and from just the BoundVars we can't tell their kind. Maybe we need to change the recursive solver to instantiate first after all? @nikomatsakis any ideas / suggestions?

### DIFF
--- a/chalk-engine/src/slg.rs
+++ b/chalk-engine/src/slg.rs
@@ -98,7 +98,12 @@ impl<'me, I: Interner> context::ContextOps<I, SlgContext<I>> for SlgContextOps<'
         goal: &DomainGoal<I>,
         _infer: &mut TruncatingInferenceTable<I>,
     ) -> Result<Vec<ProgramClause<I>>, Floundered> {
-        let clauses: Vec<_> = program_clauses_for_goal(self.program, environment, goal)?;
+        let clauses: Vec<_> = program_clauses_for_goal(
+            self.program,
+            environment,
+            goal,
+            &CanonicalVarKinds::empty(self.program.interner()),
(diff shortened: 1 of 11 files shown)


nikomatsakis:
I've come to think that the recursive solver's approach, of asking for program clauses *before* instantiation, is more correct. We could just pass the canonical binders as well, so that it can inspect the variable kinds.

flodiebold:
> We could just pass the canonical binders as well, so that it can inspect the variable kinds.

I think a prerequisite for that would be to convert the SLG solver to also ask for clauses before instantiation, probably? Since it requires changing the interface to take a `Canonical`.

flodiebold:
@jackh726 @nikomatsakis here's a variant that works and avoids refactoring the SLG solver for now. It's slightly hacky, but maybe good enough until we can get the two solvers aligned?

flodiebold:
Added the FIXME. I also realized that the `impls_for_trait` code needs the binders, so I added them as a parameter to the method.

jackh726:
Where does `impls_for_trait` need the binders?

flodiebold:
In rust-analyzer (or rustc, if it used the SLG solver). We need to know that we're actually looking for impls on integer or float types, so that we can search only those and not have to return all impls.

flodiebold:
I'll merge this now so we can use it in RA. Follow-up issue is #568.

PR #563, opened 2020-07-09
Support where clauses on opaque types

nathanwhit:
This PR iterates upon #498 to try to address the review comments there.

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -128,13 +128,14 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
 }
 
 impl<I: Interner> ToProgramClauses<I> for OpaqueTyDatum<I> {
-    /// Given `opaque type T<..>: A + B = HiddenTy;`, we generate:
+    /// Given `opaque type T<U>: A + B = HiddenTy where U: C;`, we generate:
     ///
     /// ```notrust
-    /// AliasEq(T<..> = HiddenTy) :- Reveal.
-    /// AliasEq(T<..> = !T<..>).
-    /// Implemented(!T<..>: A).
-    /// Implemented(!T<..>: B).
+    /// AliasEq(T<U> = HiddenTy) :- Reveal.
+    /// AliasEq(T<U> = !T<U>).
+    /// WF(T<U>) :- WF(U: C).
+    /// Implemented(!T<U>: A).
+    /// Implemented(!T<U>: B).
     /// ```
     /// where `!T<..>` is the placeholder for the unnormalized type `T<..>`.
     #[instrument(level = "debug", skip(builder))]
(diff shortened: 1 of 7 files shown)


chalk-solve/src/rust_ir.rs
```diff
@@ -645,6 +645,8 @@ pub struct OpaqueTyDatum<I: Interner> {
 pub struct OpaqueTyDatumBound<I: Interner> {
     /// Trait bounds for the opaque type.
```
nikomatsakis, line 646: Pre-existing, but can we clarify this by adding: "These are bounds that the hidden type must meet."

chalk-solve/src/rust_ir.rs
```diff
@@ -645,6 +645,8 @@ pub struct OpaqueTyDatum<I: Interner> {
...
     pub bounds: Binders<Vec<QuantifiedWhereClause<I>>>,
+    /// Where clauses that inform well-formedness conditions for the opaque type.
```
nikomatsakis, line 648: Clarify by adding: These are conditions on the generic parameters of an opaque type which must be true for a reference to the opaque type to be well-formed. (Is that indeed clearer? Maybe not.)

tests/test/opaque_types.rs
```diff
@@ -46,6 +46,49 @@ fn opaque_reveal() {
...
+            "Unique; substitution []"
+        }
+
+        goal {
+            WellFormed(T)
+        } yields {
+            "No possible solution"
+        }
+
+        goal {
```
nikomatsakis, line 80: maybe we should add a test `forall<U> { WellFormed(S<U>) }`? which would error, presumably.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -120,13 +120,14 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
     /// ```notrust
-    /// AliasEq(T<..> = HiddenTy) :- Reveal.
-    /// AliasEq(T<..> = !T<..>).
-    /// Implemented(!T<..>: A).
-    /// Implemented(!T<..>: B).
+    /// AliasEq(T<U> = HiddenTy) :- Reveal.
+    /// AliasEq(T<U> = !T<U>).
+    /// WF(T<U>) :- WF(U: C).
+    /// Implemented(!T<U>: A).
+    /// Implemented(!T<U>: B).
```
nikomatsakis, line 130: So this seems right, but also sort of...incomplete, in that we never use these where clauses as hypotheses. I guess this is because we would want to do that in the well-formedness checking for the opaque type **definition**, which is not yet implemented?

nathanwhit: That was my thinking as well. I was thinking those checks would probably belong around [here](https://github.com/rust-lang/chalk/blob/0f43952c643248811a730c2eef63ce2fe60d1a64/chalk-solve/src/wf.rs#L142) based on what we do for ADTs and trait impls. I figure we could save that for a follow-up PR.

PR #565, opened 2020-07-09
Remove unit fields from enum variants

nathanwhit:
Currently in a few places (e.g. `GoalData::CannotProve(())`, `DomainGoal::Reveal(())`) we have enum variants which have one field of unit type. There's a comment noting that it was a hack in order to make our macros work better, but it's no longer needed.


nikomatsakis:
yay!

Issue #299, opened 2019-11-24
(tests only) Add a "simple" mode


jackh726:
So, when solving a goal, `chalk-solve` automatically adds various clauses for potential downstream impls, `Compatible`, etc. Additionally, any tests not in `test/slg` also are checked that they don't violate orhpan rules, etc. While, this is definitely good in general, it makes actually debugging of logic (using `CHALK_DEBUG=1` or `2` much more noisy.

I think it would be super helpful to somehow have a mode for tests where only the clauses specified by the test are used and the program isn't checked. Importantly, this would be completely opt-in (so maybe an environment variable) and would never be used by default for any tests.

This would be pretty easy to implement:

This `checked_program` call would need to be conditional:
https://github.com/rust-lang/chalk/blob/3ef20308e7882a3e5fb71a6f28caab8cc71cf5f0/tests/test/mod.rs#L148

Also, the extra clauses get added in `program_clauses.rs`, for example:
https://github.com/rust-lang/chalk/blob/3ef20308e7882a3e5fb71a6f28caab8cc71cf5f0/chalk-solve/src/clauses/program_clauses.rs#L183
https://github.com/rust-lang/chalk/blob/3ef20308e7882a3e5fb71a6f28caab8cc71cf5f0/chalk-solve/src/clauses/program_clauses.rs#L415

Again, the purpose of this is to make debugging logic much easier. There will almost certainly be tests that fail with the option enabled, but that's okay because it isn't for them.

Mcat12:
@jackh726 is this issue ready to go? It's labeled with "design" (aka "needs design work"), but seems like it is straightforward enough to start implementing.

Maybe, we need to decide if this is important or if it will be used. Do you envision this only being used during test creation or debugging? Also, would this option be per-test (add syntax to `test` macro) or for all tests (environment variable)?

jackh726:
@Mcat12 yeah so I marked this a "design" more so because it "needs decision". It's actually pretty straightforward to implement.

(I wanted to just link a comment @nikomatsakis made in the last meeting, since it's relevant https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/design.20meeting.202020.2E05.2E26/near/198818703)

[...]

PR #566, opened 2020-07-09
Only generate clauses requiring Compatible when it is in the environment

jackh726:
cc. #299

I'm not sure if there's anything *else* we want to do for that issue. But, as is, this should generate many fewer clauses, which is a win not only because it's less work for the solvers to do, but also makes debugging a bit nicer.

### DIFF
--- a/chalk-integration/src/query.rs
+++ b/chalk-integration/src/query.rs
@@ -172,20 +172,22 @@ fn environment(db: &impl LoweringDatabase) -> Result<Arc<ProgramEnvironment>, Ch
 
     let builder = &mut ClauseBuilder::new(db, &mut program_clauses);
 
+    let env = chalk_ir::Environment::new(builder.interner());
+
     program
         .associated_ty_data
         .values()
-        .for_each(|d| d.to_program_clauses(builder));
+        .for_each(|d| d.to_program_clauses(builder, &env));
 
     program
         .trait_data
         .values()
-        .for_each(|d| d.to_program_clauses(builder));
+        .for_each(|d| d.to_program_clauses(builder, &env));
 
     program
         .adt_data
         .values()
-        .for_each(|d| d.to_program_clauses(builder));
+        .for_each(|d| d.to_program_clauses(builder, &env));
 
     for (&auto_trait_id, _) in program
         .trait_data
@@ -201,12 +203,12 @@ fn environment(db: &impl LoweringDatabase) -> Result<Arc<ProgramEnvironment>, Ch
         // If we encounter a negative impl, do not generate any rule. Negative impls
         // are currently just there to deactivate default impls for auto traits.
         if datum.is_positive() {
-            datum.to_program_clauses(builder);
+            datum.to_program_clauses(builder, &env);
             datum
                 .associated_ty_value_ids
                 .iter()
                 .map(|&atv_id| db.associated_ty_value(atv_id))
-                .for_each(|atv| atv.to_program_clauses(builder));
+                .for_each(|atv| atv.to_program_clauses(builder, &env));
         }
     }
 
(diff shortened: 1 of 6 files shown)


chalk-integration/src/query.rs
```diff
@@ -172,20 +172,22 @@ fn environment(db: &impl LoweringDatabase) -> Result<Arc<ProgramEnvironment>, Ch
 
     let builder = &mut ClauseBuilder::new(db, &mut program_clauses);
```
jackh726, line 174: This function (`environment`), from what I can tell is a bit of a relic. It's only used `bench` and in the REPL (when you call "lowered"). It can probably just be removed (along with `bench` tbh). But that's a separate PR.

chalk-solve/src/clauses.rs
```diff
@@ -283,7 +284,25 @@ fn program_clauses_that_could_match<I: Interner>(
...
                 })
                 | TyData::Alias(AliasTy::Opaque(OpaqueTy { opaque_ty_id, .. })) => {
-                    db.opaque_ty_data(*opaque_ty_id).to_program_clauses(builder);
+                    db.opaque_ty_data(*opaque_ty_id)
+                        .to_program_clauses(builder, environment);
+                }
+                _ => {}
+            }
+
+            // We don't actually do anything here, but we need to record the types it when logging
```
jackh726, line 293: This surprised me a bit. But I can't think of a better way to handle it.
The problem here is that `foo: Sized` (where `foo` is a function) doesn't actually call `fn_def_datum`. So the `LoggingRustIrDatabase` doesn't pick it up. This could be removed when the that actually records the *goals* too.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -596,29 +616,55 @@ impl<I: Interner> ToProgramClauses<I> for TraitDatum<I> {
...
-            // be implemented with the same genericity as the struct definition,
-            // i.e. Drop implementation for `struct S<T: Eq> {}` is forced to be
-            // `impl Drop<T: Eq> for S<T> { ... }`. That means that orphan rules
-            // prevent Drop from being implemented in downstream crates.
-            if self.well_known != Some(WellKnownTrait::Drop) {
-                // Add all cases for potential downstream impls that could exist
-                for i in 0..type_parameters.len() {
+            if environment.has_compatible_clause(interner) {
+                // Note: even though we do check for a `Compatible` clause here,
+                // we also keep it as a condition for the clauses below, purely
```
flodiebold, line 621: I think that makes sense -- conceptually, I think the solver should work the same if all clauses were being returned all the time, the clauses code should only decide to not generate non-applicable clauses.

chalk-ir/src/lib.rs
```diff
@@ -90,6 +90,25 @@ impl<I: Interner> Environment<I> {
...
+    /// True if any of the clauses in the environment have a consequence of `Compatible`.
+    /// Panics if the conditions or constraints of that clause are not empty.
+    pub fn has_compatible_clause(&self, interner: &I) -> bool {
+        self.clauses.as_slice(interner).iter().any(|c| {
+            let ProgramClauseData(implication) = c.data(interner);
+            match implication.skip_binders().consequence {
+                DomainGoal::Compatible => {
+                    // We currently don't generate `Compatible` with any conditions or constraints
+                    // If this was needed, for whatever reason, then a third "yes, but must evaluate"
+                    // return value would have to be added.
```
flodiebold, line 103: not really necessary if the clauses code continues to generate clauses that have `Compatible` as a condition where required, right?

jackh726: Yes, that's true. This wording is a bit conservative.

chalk-solve/src/clauses/program_clauses.rs
nikomatsakis: Yes, agreed

PR #569, opened 2020-07-11
Guard against infinite loop in recursive solver

flodiebold:
This happened in some tests in rust-analyzer. I think it can only happen with overlapping impls, so I don't know how to actually test it. (In rust-analyzer, it happened because of built-in impls provided both by RA and now Chalk.)

### DIFF
--- a/chalk-recursive/src/fulfill.rs
+++ b/chalk-recursive/src/fulfill.rs
@@ -447,9 +447,12 @@ impl<'s, I: Interner, Solver: SolveDatabase<I>, Infer: RecursiveInferenceTable<I
                             solution,
                         } = self.prove(wc, minimums)?;
 
-                        if solution.has_definite() {
-                            if let Some(constrained_subst) =
-                                solution.constrained_subst(self.interner())
+                        if let Some(constrained_subst) = solution.definite_subst(self.interner()) {
+                            if !constrained_subst.value.subst.is_empty(self.interner())
+                                || !constrained_subst
+                                    .value
+                                    .constraints
+                                    .is_empty(self.interner())
                             {
                                 self.apply_solution(free_vars, universes, constrained_subst);
                                 progress = true;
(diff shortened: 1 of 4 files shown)


tests/test/misc.rs
```diff
@@ -727,3 +727,34 @@ fn canonicalization_regression() {
         }
     }
 }
+
+#[test]
+#[ignore]
+// this is a regression test for an infinite loop, but it doesn't actually work
+// because the code fails coherence checking.
```
nikomatsakis, line 734: hmm I feel like we *used* to have a way (and should have a way) to disable coherence, I guess that would help?

chalk-recursive/src/fulfill.rs
```diff
@@ -447,9 +447,12 @@ impl<'s, I: Interner, Solver: SolveDatabase<I>, Infer: RecursiveInferenceTable<I
                             solution,
                         } = self.prove(wc, minimums)?;
 
-                        if solution.has_definite() {
-                            if let Some(constrained_subst) =
-                                solution.constrained_subst(self.interner())
+                        if let Some(constrained_subst) = solution.definite_subst(self.interner()) {
+                            if !constrained_subst.value.subst.is_empty(self.interner())
```
nikomatsakis, line 451: can we add a comment explaining this condition?

tests/test/misc.rs
flodiebold: It would, yeah. (I tried the test manually by disabling the coherence check in `checked_program`.) I didn't find such a way, but I could try adding it.

nikomatsakis: I think adding it makes sense, not sure how annoying it would be to do though. Maybe it has to be an "input" into the salsa setup?

nikomatsakis:
@flodiebold want to add a comment here?

flodiebold:
Yeah, I actually wanted to look into making it testable, but didn't get to it yet.

flodiebold:
@jackh726 @nikomatsakis I've added a comment and a way to disable coherence in tests.

PR #570, opened 2020-07-11
Cleanup & comment .chalk file writer tests

daboross:
This addresses another issue raised in the final review of #430 - mainly, that none of the tests explained at all what they're doing, and multiple tests were often merged into one test function.

We've made sure that each test makes sense, has comments, and only tests a single item. This involved removing reduntant tests, and creating new test functions. The comments are useful when a test's purpose is obscured by the substantial boilerplate it requires.


chalk-solve/src/display.rs
```diff
@@ -25,7 +25,7 @@ pub use self::state::*;
 
 use self::utils::as_display;
 
-pub fn write_item<F, I, T>(f: &mut F, ws: &WriterState<'_, I>, v: &T) -> Result
+fn write_item<F, I, T>(f: &mut F, ws: &WriterState<'_, I>, v: &T) -> Result
```
daboross, line 28: This is technically unrelated, but fairly insubstantial.

detrumi: Well, as long as `write_items` is still public and usable, this should be fine

daboross: :+1:! I think it is a positive change, just that it probably should have ended up part of #558 rather than this PR.

PR #572, opened 2020-07-12
Add readme for chalk-recursive

nathanwhit:
This should fix the publishing failure.


jackh726:
We should also manually update the schedule so the publish is done today. @nathanwhit do you want to do this? Basically, just update the cran schedule to run sometime soon, let it run, then update it back. If not, I can do it.

nathanwhit:
Sure thing! I'll take care of that.

PR #575, opened 2020-07-18
Add default `RustIrDatabase::*_name` impls

daboross:
This adds default implementations for `RustIrDatabase::*_name` methods. They use `Interner::debug_*` methods, and sanitize the output by replacing all non-alphanumeric-ascii chars with `_`.

This also removes all custom implementations of these methods, except the one for `Program::assoc_type_name`. We've kept that one, as the default impl ends up returning `_TraitName__AssocName_`, and for display tests, we need it to be exactly `AssocName`.

We've discussed the justification for this in [the .chalk file writer zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/.2Echalk.20file.20writer), and [the first .chalk file writer pull request](https://github.com/rust-lang/chalk/pull/430). But for anyone reading this, it boils down to the fact that we don't want an to impose an undue burden on `RustIrDatabase` implementors, especially for a debug tool. Rustc integration shouldn't have to accommodate `LoggingRustIrDatabase`.

### DIFF
--- a/chalk-integration/src/program.rs
+++ b/chalk-integration/src/program.rs
@@ -485,31 +485,17 @@ impl RustIrDatabase<ChalkIr> for Program {
         substs.clone()
     }
 
-    fn trait_name(&self, trait_id: TraitId<ChalkIr>) -> String {
-        self.trait_kinds.get(&trait_id).unwrap().name.to_string()
-    }
-
-    fn adt_name(&self, struct_id: AdtId<ChalkIr>) -> String {
-        self.adt_kinds.get(&struct_id).unwrap().name.to_string()
-    }
-
+    // The default implementation for `RustIrDatabase::assoc_type_name` outputs
+    // the name in the format `(Trait::AssocTypeName)`, which is reformatted to
+    // `_Trait__AssocTypeName_`. This doesn't match the input names, which is
+    // normally acceptable, but causes the re-parse tests for the .chalk syntax
+    // writer to fail. This is because they use the `Eq` implementation on
+    // Program, which checks for name equality.
     fn assoc_type_name(&self, assoc_type_id: AssocTypeId<ChalkIr>) -> String {
         self.associated_ty_data
             .get(&assoc_type_id)
             .unwrap()
             .name
             .to_string()
-    }
-
-    fn opaque_type_name(&self, opaque_ty_id: OpaqueTyId<ChalkIr>) -> String {
-        self.opaque_ty_kinds
-            .get(&opaque_ty_id)
-            .unwrap()
-            .name
-            .to_string()
-    }
-
-    fn fn_def_name(&self, fn_def_id: FnDefId<ChalkIr>) -> String {
-        self.fn_def_kinds.get(&fn_def_id).unwrap().name.to_string()
     }
 }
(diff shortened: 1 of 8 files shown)


daboross:
Forgot to run all the tests!

There's an additional bug related to the WriterState we need to fix in order to get this working.

daboross:
Fixed that bug! This should be ready.

Incidentally, that bug exposed another bug related to how we're doing name deduplication. We'll submit a fix for that in a separate PR as soon as we've written tests.

chalk-solve/src/display/utils.rs
```diff
@@ -23,3 +23,32 @@ macro_rules! write_joined_non_empty_list {
...
+    .expect("expected writing to a String to succeed");
+    if debug_out.is_empty() {
+        return "Unknown".to_owned();
+    }
+
+    // now the actual sanitization
+    debug_out
+        .chars()
+        .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
+        .collect()
```
detrumi, line 53: Nit: this can be replaced by the following:
```
debug_out.replace(|c: char| c.is_ascii_alphanumeric(), "_");
```

chalk-solve/src/lib.rs
```diff
@@ -144,21 +163,35 @@ pub trait RustIrDatabase<I: Interner>: Debug {
         substs: &Substitution<I>,
     ) -> Substitution<I>;
 
-    /// Retrieves a trait's original name. No uniqueness guarantees.
-    /// TODO: remove this, use only interner debug methods
-    fn trait_name(&self, trait_id: TraitId<I>) -> String;
+    /// Retrieves a trait's original name. No uniqueness guarantees, but must
+    /// a valid Rust identifier.
+    fn trait_name(&self, trait_id: TraitId<I>) -> String {
+        crate::display::sanitize_debug_name(|f| I::debug_trait_id(trait_id, f))
```
detrumi, line 169: Maybe add `use crate::display::sanitize_debug_name` in this file?

flodiebold:
> This should also help on RA's side, because displaying the names there is tricky as well.

It shouldn't be *that* tricky, we just haven't implemented it :wink:

chalk-integration/src/program.rs
```diff
@@ -471,31 +485,17 @@ impl RustIrDatabase<ChalkIr> for Program {
...
-    fn adt_name(&self, struct_id: AdtId<ChalkIr>) -> String {
-        self.adt_kinds.get(&struct_id).unwrap().name.to_string()
-    }
-
+    // The default implementation for `RustIrDatabase::assoc_type_name` outputs
+    // the name in the format `(Trait::AssocTypeName)`, which is reformatted to
+    // `_Trait__AssocTypeName_`. This doesn't match the input names, which is
+    // normally acceptable, but causes the re-parse tests for the .chalk syntax
+    // writer to fail. This is because they use the `Eq` implementation on
+    // Program, which checks for name equality.
```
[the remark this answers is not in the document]
daboross: The display test suite currently relies on exact name equality, so we need to do _something_ to get the `assoc_type_name` type name output to exactly match the input.

One alternative would be to make a `RustIrDatabase` wrapper for `Program` which just overwrites `assoc_type_name`, but I'm not sure that would be worth it. Especially since having this one method overwrite also means that the debug printer can have fully exact names when printing output from the chalk command line, and this code is run in no other situations. At least to my knowledge, neither rustc nor rust-analyzer integration uses `Program`, right?

We could try to create a test equality that doesn't depend on names, but it would also be more effort - and I'm not sure we want to allow names to slip in those tests without us noticing. They're otherwise fairly exact output correctness tests.

jackh726:
@detrumi we have bors now :) `@bors r+`

PR #578, opened 2020-07-21
Add workflow_dispatch for manual publish

pksunkara:
As discussed in https://github.com/rust-analyzer/rust-analyzer/pull/4982


pksunkara:
I tested it and the button worked on my fork.

PR #579, opened 2020-07-22
Check well-formedness of opaque type declarations

nathanwhit:


tests/test/wf_lowering.rs
```diff
@@ -1103,3 +1103,17 @@ fn no_unsize_impls() {
         }
     }
 }
+
+#[test]
+fn ill_formed_opaque_ty() {
+    lowering_error! {
+        program {
+            trait Foo {}
+            struct Bar {}
```
nikomatsakis, line 1112: try this test

```rust
trait Foo { }
struct NotFoo { }
struct IsFoo { }
impl Foo for IsFoo { }
opaque type T: Foo = NotFoo;
```

I think your code will accept it, but I could be wrong

chalk-solve/src/wf.rs
```diff
@@ -269,6 +275,70 @@ where
...
+        let mut gb = GoalBuilder::new(self.db);
+
+        let datum = self.db.opaque_ty_data(opaque_ty_id);
+        let bound = &datum.bound;
+
+        // We make a goal like
+        //
+        // forall<T>
+        let goal = gb.forall(&bound, opaque_ty_id, |gb, _, bound, opaque_ty_id| {
+            // exists<Self>
```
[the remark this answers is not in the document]
nathanwhit: Yup we should. Right now I substitute the hidden type for `Self` for the bounds, but not for the where clauses (which just doesn't make sense).

tests/test/wf_lowering.rs
nathanwhit: It does reject this actually, since you still need to prove `NotFoo: Foo`. But, regardless, your point still stands

nikomatsakis:
@nathanwhit if you rebase, I'm happy to merge

PR #580, opened 2020-07-22
Don't fail CI if only nightly fails

nathanwhit:
This way the check using stable Rust will still run even if nightly fails. This is motivated by the ICE on nightly recently, which currently causes the stable check to halt.


jackh726:
Yeah, this is probably what we want. We test on nightly basically only to catch regressions right?

PR #581, opened 2020-07-22
Upgrade salsa

nathanwhit:
The update was *mostly* painless, but there were a few queries where we needed to go from `LoweringDatabase` -> `RustIrDatabase` which posed a bit of an issue. This worked fine with generics, but now salsa passes a trait object to queries (and trait object upcasting isn't a thing yet) so I had to use a bit of a workaround.


nikomatsakis:
Thanks!

jackh726:
@nikomatsakis we have bors now :) `@bors r+`

nikomatsakis:
Yeah, I realized that later, whoops!

PR #582, opened 2020-07-22
Create single Github actions job to gate bors on

nathanwhit:
This is a preliminary step for getting bors integrated into the chalk repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,3 +111,23 @@ jobs:
 
       - name: Build Chalk book
         run: cd book && ./mdbook build
+
+  end-success:
+    name: bors build finished
+    if: success()
+    runs-on: ubuntu-latest
+    needs: [test, fmt, mdbook-linkcheck]
+
+    steps:
+      - name: Mark the job as successful
+        run: exit 0
+
+  end-failure:
+    name: bors build finished
+    if: "!success()"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jackh726:
Hmm. So how will this work will mdbook-linkcheck? I kind of feel like we shouldn't gate anything on it. Also, seems like it generates to *separate* checks (maybe one for stable and one for nightly test)

nathanwhit:
> Hmm. So how will this work will mdbook-linkcheck? I kind of feel like we shouldn't gate anything on it. Also, seems like it generates to _separate_ checks (maybe one for stable and one for nightly test)

Makes sense to not gate on linkcheck, so I updated the workflow to reflect that. 

As for the two checks (which was adopted from [here](https://github.com/rust-lang/crater/blob/c3f462bdab37a93c24b2b172b90564749e892cbc/.github/workflows/bors.yml#L126-L150) as suggested by `pietroalbini`) as far as I know it's due to limitations in GHA, which make it so that you need to have two separate (and mutually exclusive) checks –– one that just returns `0` if all other checks passed, and another that just returns `1` if any other check fails.

jackh726:
Oh, duh. Right. Um, do they have to be named "bors build finished" or could they be "bors build finished success" and "bors build finished failed" to make them more clear

nathanwhit:
As far as I know, I think they do need to have the same name (unfortunately).

PR #585, opened 2020-07-25
.chalk syntax writer: Fix name de-duplication bug

super-tuple:
While working on https://github.com/rust-lang/chalk/pull/575, we found a bug in the name de-duplication logic. To ensure we have unique names, we store a mapping between ids and names for ids that don't have unique names. This state was managed inside of `write_items`, but we call `write_items` twice when we output the data from the `LoggingRustIrDatabase`. It is called once to output the logged items, and once to output the necessary stub items. The id to name mappings are lost between these calls, which breaks the unique name guarantee.

We fixed this by creating the mapping information store outside of the display code, and passing it into calls to `write_items`. We also significantly improved the name de-duplication tests.

### DIFF
--- a/chalk-solve/src/display/state.rs
+++ b/chalk-solve/src/display/state.rs
@@ -148,12 +148,14 @@ where
     ///
     /// `f` will be run on the internal database, and the returned result will
     /// wrap the result from `f`. For consistency, `f` should always contain the
-    /// given database, and must keep the same ID<->item relationships.  
+    /// given database, and must keep the same ID<->item relationships.
     pub(super) fn wrap_db_ref<'a, DB2: ?Sized, P2, F>(&'a self, f: F) -> WriterState<I, DB2, P2>
     where
-        P: 'a,
         DB2: RustIrDatabase<I>,
         P2: Borrow<DB2>,
+        // We need to pass in `&'a P` specifically to guarantee that the `&P`
+        // can outlive the function body, and thus that it's safe to store `&P`
+        // in `P2`.
         F: FnOnce(&'a P) -> P2,
     {
         WriterState {
--- a/chalk-solve/src/display/state.rs
+++ b/chalk-solve/src/display/state.rs
@@ -1,11 +1,13 @@
 //! Persistent state passed down between writers.
 //!
-//! This is essentially `WriterState` and other things supporting that.
+//! This is essentially `InternalWriterState` and other things supporting that.
 use std::{
-    cell::RefCell,
+    borrow::Borrow,
     collections::BTreeMap,
-    fmt::{Display, Formatter, Result},
+    fmt::{Debug, Display, Formatter, Result},
+    marker::PhantomData,
     rc::Rc,
+    sync::{Arc, Mutex},
 };
 
 use crate::RustIrDatabase;
@@ -34,8 +36,14 @@ impl Display for InvertedBoundVar {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+enum UnifiedId<I: Interner> {
+    AdtId(I::InternedAdtId),
+    DefId(I::DefId),
+}
+
 #[derive(Debug)]
-pub struct DefIdAliases<T: Ord> {
+pub struct IdAliasStore<T: Ord> {
     /// Map from the DefIds we've encountered to a u32 alias id unique to all ids
     /// the same name.
     aliases: BTreeMap<T, u32>,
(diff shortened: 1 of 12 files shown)


chalk-solve/src/logging_db.rs
```diff
@@ -327,45 +325,45 @@ where
...
-        self.db.borrow().opaque_ty_data(id)
+        self.db.opaque_ty_data(id)
     }
 
     fn hidden_opaque_type(&self, id: OpaqueTyId<I>) -> Ty<I> {
-        self.db.borrow().hidden_opaque_type(id)
+        self.db.hidden_opaque_type(id)
     }
 
     fn impls_for_trait(
```
detrumi, line 369: You missed a `db.borrow()` call here, and in `closure_inputs_and_output` as well

chalk-solve/src/display/state.rs
```diff
@@ -71,35 +79,134 @@ impl<T: Copy + Ord> DefIdAliases<T> {
...
+
+    /// Returns a new version of self containing a wrapped database which
+    /// references the outer data.
+    ///
+    /// `f` will be run on the internal database, and the returned result will
+    /// wrap the result from `f`. For consistency, `f` should always contain the
+    /// given database, and must keep the same ID<->item relationships.  
+    pub(super) fn wrap_db_ref<'a, DB2: ?Sized, P2, F>(&'a self, f: F) -> WriterState<I, DB2, P2>
+    where
+        P: 'a,
```
detrumi, line 154: This bound doesn't seem to be needed

chalk-solve/src/display/state.rs
```diff
@@ -109,12 +216,22 @@ impl<'a, I: Interner> WriterState<'a, I> {
         std::iter::repeat("  ").take(self.indent_level).format("")
     }
 
-    pub(super) fn alias_for_id_name(&self, id: I::DefId, name: String) -> impl Display {
-        self.def_id_aliases.borrow_mut().alias_for_id_name(id, name)
+    pub(super) fn alias_for_adt_id_name(&self, id: I::InternedAdtId, name: String) -> impl Display {
```
detrumi, line 219: I wonder if these could be simplified by making this `id: impl Into<UnifiedId<I>>`.

daboross: We tried to, but I think using `Into<UnifiedId<I>>` in general has some significant issues. Specifically, we would need to have two implementations

```rust
impl<I: Interner> From<I::DefId> for UnifiedId<I> { ... }

impl<I: Interner> From<I::InternedAdtId> for UnifiedId<I> { ... }
```

and these end up conflicting, since there's no guarantee that `DefId` and `InternedAdtId` are actually different types for an arbitrary interner.

chalk-solve/src/display/state.rs
daboross: I think we need it because `f` takes `&'a P`, that being sound requires `P: 'a`, and since we're just introducing `'a` here, there's no guarantee it really is shorter than `P`?

Specifically, we get:

```
error[E0495]: cannot infer an appropriate lifetime for borrow expression due to conflicting requirements
   --> chalk-solve\src\display\state.rs:160:19
    |
160 |             db: f(&self.db),
    |                   ^^^^^^^^
    |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the method body at 152:5...
   --> chalk-solve\src\display\state.rs:152:5
    |
152 | /     pub(super) fn wrap_db_ref<'a, DB2: ?Sized, P2, F>(&self, f: F) -> WriterState<I, DB2, P2>
153 | |     where
154 | |         P: 'a,
155 | |         DB2: RustIrDatabase<I>,
...   |
163 | |         }
164 | |     }
    | |_____^
note: ...so that reference does not outlive borrowed content
   --> chalk-solve\src\display\state.rs:160:19
    |
160 |             db: f(&self.db),
    |                   ^^^^^^^^
note: but, the lifetime must be valid for the lifetime `'a` as defined on the method body at 152:31...
   --> chalk-solve\src\display\state.rs:152:31
    |
152 |     pub(super) fn wrap_db_ref<'a, DB2: ?Sized, P2, F>(&self, f: F) -> WriterState<I, DB2, P2>
    |                               ^^
    |
160 |             db: f(&self.db),
    |                   ^^^^^^^^

error: aborting due to previous error
```

We need `'a` at all, because we want to be able to create a `P2` which directly wraps `P`, and in order to do that the `&P` needs a lifetime that's guaranteed to outlive the function. If it was just `FnOnce(&P) -> P2`, then the `&P` could have been created inside `wrap_db_ref`, and then it really would be unsound to have `P2` wrapping `P`.

PR #589, opened 2020-08-02
[WIP] Convert from semantic to syntactic equality

nathanwhit:
A continuation of #401. Currently only projections are handled, but all tests pass. Still working on the lowering for lifetimes and function types.


Areredify:
@nathanwhit would you mind me repicking this up?

PR #591, opened 2020-08-04
Handle well known traits for more types

### LINKED ISSUE #533: Reorganize(/cleanup) tests
We've recently added a bunch of builtin types and traits, and a lot of tests associated with them. We should decide whether we're organizing tests by the *trait* the implement, or the *type* that implements.

Right now, there are some tests in `test/builtin_impls`, but much more in `test/*`. For example, there is `test/builtin_impl/copy.rs` which has *only* the test for tuples being copy. All the other `Copy`-related tests are in separate files. For example, `fn_def_is_copy` is in `test/fn_def.rs`.

I imagine it'll be easiest to just have each type be a separate file, so testing for `Copy`, `Clone`, `WF`, etc. would all be in one file. This would roughly line up with each row in http://rust-lang.github.io/chalk/book/clauses/well_known_traits.html being in a separate file.

I somewhat think that `Unsize` might be different. The tests are somewhat complicated, and I would feel better about being able to reason about the same logic across different types. Additionally, we can somewhat justify this since these tests are really testing the conversion of one type to another.

Somewhat related to this, it might be nice to go through and document *why* we have each test (as well as the individual goals in each test). Though, I'm not sure if that should be a part of this issue or a separate issue.

### LINKED ISSUE #590: Regression for well known trait impls for builtin types
So, this regression happened sometime between 0.14 and 0.20.

Basically, if we have a goal like
```
Canonical {
  value: InEnvironment {
    environment: Env([]),
    goal: Implemented(Ty(^0.0): TraitId(DefId(2:1942 ~ core[d282]::marker[0]::Sized[0]))),
  },
  binders: [U0 with kind integer type],
}
```
then we don't generate the clause for `Sized` impl (at least with SLG).

We pass `Implemented(Ty(?0i): TraitId(DefId(2:1942 ~ core[d282]::marker[0]::Sized[0])))` to `program_clauses_for_goal`; the `?0i` self type is not covered in `add_sized_program_clauses`. ~~I haven't tested `Copy`/`Clone`, but I'm almost certain the same problem occurs.~~ Edit: forgot that these are in libcore, so aren't builtin for ints/floats.

nathanwhit:
Adds clauses for well known traits (mostly `Sized`, `Copy`) for more types, including inference vars and bound vars. Also reorganizes tests for well known traits and adds some missing tests.

Fixes #590. 
Fixes #533.


tests/test/refs.rs
```diff
@@ -29,6 +29,30 @@ fn immut_refs_are_sized() {
     }
 }
 
+#[test]
+fn immut_refs_are_copy_clone() {
```
jackh726, line 33: This isn't correct, since these are in `libcore`.

tests/test/refs.rs
```diff
@@ -57,3 +81,27 @@ fn mut_refs_are_sized() {
         }
     }
 }
+
+#[test]
+fn mut_refs_are_not_copy_clone() {
```
jackh726, line 86: `libcore`

tests/test/scalars.rs
```diff
@@ -136,3 +136,73 @@ fn scalars_are_well_formed() {
...
+
+#[test]
+fn scalars_are_copy_clone() {
+    test! {
+        program {
+            #[lang(copy)] trait Copy { }
+            #[lang(clone)] trait Clone { }
+        }
+
+        goal { i8: Copy } yields { "Unique" }
```
jackh726, line 174: All of these are `libcore`

chalk-solve/src/clauses/builtin_traits/copy.rs
```diff
@@ -58,11 +64,40 @@ pub fn add_copy_program_clauses<I: Interner>(
                 let upvars = upvars.substitute(db.interner(), &closure_fn_substitution);
                 needs_impl_for_tys(db, builder, trait_ref, Some(upvars).into_iter());
             }
-            _ => {}
+
+            TypeName::Ref(Mutability::Not)
+            | TypeName::Raw(_)
+            | TypeName::Scalar(_)
+            | TypeName::Never => builder.push_fact(trait_ref.clone()),
```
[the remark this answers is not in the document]
nathanwhit: (I don't know how I always manage to forget this 😓)

tests/test/functions.rs
```diff
@@ -1,60 +1,246 @@
...
 
+        goal {
+            fn([u8]): Copy
+        } yields {
+            "Unique; substitution [], lifetime constraints []"
+        }
     }
 }
 
+use crate::test::*;
```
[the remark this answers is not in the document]
nathanwhit: Copy-and-pasted when rearranging tests, whoops.

jackh726:
I'm also going to do a release after this gets merged to make sure things are good with rustc.

PR #592, opened 2020-08-05
Turn a few generics into dynamic dispatch.

### LINKED ISSUE #562: Use `dyn Solver` instead of `S: Solver` in `chalk_solve`
See https://github.com/rust-lang/chalk/pull/546#discussion_r448618711

crlf0710:
Fixes #562.

r? @jackh726
cc @nikomatsakis

### DIFF
--- a/chalk-integration/src/query.rs
+++ b/chalk-integration/src/query.rs
@@ -8,7 +8,6 @@ use crate::program::Program;
 use crate::program_environment::ProgramEnvironment;
 use crate::tls;
 use crate::SolverChoice;
-use crate::SolverImpl;
 use chalk_ir::TraitId;
 use chalk_solve::clauses::builder::ClauseBuilder;
 use chalk_solve::clauses::program_clauses::ToProgramClauses;
@@ -16,6 +15,7 @@ use chalk_solve::coherence::orphan;
 use chalk_solve::coherence::{CoherenceSolver, SpecializationPriorities};
 use chalk_solve::wf;
 use chalk_solve::RustIrDatabase;
+use chalk_solve::Solver;
 use salsa::Database;
 use std::clone::Clone;
 use std::cmp::{Eq, PartialEq};
@@ -57,7 +57,7 @@ pub trait LoweringDatabase:
     /// volatile, thus ensuring that the solver is recreated in every
     /// revision (i.e., each time source program changes).
     // HACK: salsa requires that queries return types that implement `Eq`
-    fn solver(&self) -> ArcEq<Mutex<SolverImpl>>;
+    fn solver(&self) -> ArcEq<Mutex<Box<dyn Solver<ChalkIr>>>>;
 }
 
 // Needed to go from dyn LoweringDatabase -> dyn RustIrDatabase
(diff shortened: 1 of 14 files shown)


chalk-solve/src/coherence/solve.rs
```diff
@@ -131,7 +130,7 @@ impl<I: Interner, S: Solver<I>, SC: Into<S> + Copy> CoherenceSolver<'_, I, S, SC
             .negate(interner);
 
         let canonical_goal = &goal.into_closed_goal(interner);
-        let solution = self.solver_choice.into().solve(self.db, canonical_goal);
+        let solution = self.solver.solve(self.db, canonical_goal);
```
jackh726, line 133: So...this is not the same things semantically. Before, we use a fresh solver for each goal. Now, we use a single solver for all goals. Importantly, the solver can (and does) keep some state between goals. It's a bit surprising to me that nothing fails here. (Really, I think the only thing we have to be careful about is placeholders). I think this needs to stay as creating a new solver for each goal, but with some convincing (from myself or otherwise), this could be changed now or later.

crlf0710:
Updated to make coherence and wf solvers create  fresh inner solver every time.

chalk-solve/src/coherence.rs
```diff
@@ -11,11 +11,10 @@ use std::sync::Arc;
 pub mod orphan;
 mod solve;
 
-pub struct CoherenceSolver<'db, I: Interner, S, SC: Into<S> + Copy> {
+pub struct CoherenceSolver<'db, 'sb, I: Interner> {
```
jackh726, line 14: I think this (and `WfSolver`) could just have a single `'a` lifetime.

chalk-solve/src/coherence.rs
```diff
@@ -72,24 +71,25 @@ impl<I: Interner> SpecializationPriorities<I> {
...
-            solver_choice,
+            solver_builder,
             trait_id,
-            _solver: std::marker::PhantomData,
         }
     }
 
     pub fn specialization_priorities(
-        &self,
+        &mut self,
```
jackh726, line 92: Why do these need `&mut self` now?

chalk-integration/src/query.rs
```diff
@@ -132,12 +132,9 @@ fn orphan_check(db: &dyn LoweringDatabase) -> Result<(), ChalkError> {
 
     tls::set_current_program(&program, || -> Result<(), ChalkError> {
         let local_impls = program.local_impl_ids();
+        let mut solver = db.solver_choice().into_solver();
```
jackh726, line 135: So to be semantically the same, this should be *inside* the loop (and then you could just pass the solver by value). But I think we only generate a `LocalImplAllowed` goal with no placeholders, so this should be fine to reuse.

chalk-solve/src/coherence/solve.rs
```diff
@@ -11,9 +10,9 @@ use chalk_ir::*;
 use itertools::Itertools;
 use tracing::{debug, instrument};
 
-impl<I: Interner, S: Solver<I>, SC: Into<S> + Copy> CoherenceSolver<'_, I, S, SC> {
+impl<I: Interner> CoherenceSolver<'_, '_, I> {
     pub(super) fn visit_specializations_of_trait(
-        &self,
+        &mut self,
```
jackh726, line 15: Same here: Why now `&mut self`

chalk-solve/src/coherence.rs
crlf0710: Indeed! Done.

chalk-solve/src/coherence.rs
crlf0710: It's no longer needed now. I have changed them back.

chalk-solve/src/coherence/solve.rs
crlf0710: Done.

chalk-integration/src/query.rs
crlf0710: I've moved it inside the loop for consistency.

jackh726:
Thanks! @bors r+

PR #596, opened 2020-08-19
Support const evaulation

### LINKED ISSUE #548: Support const evaluation

josh65536:
Fixes #548 
For Chalk's tests, a special-formatted `Option<u32>` is used, where `Some(n)` evaluates to `n` and `None` evaluates to a specific but unknown constant. `Some(n)` is formatted as `n?` and `None` is formatted as `?`.


chalk-ir/src/lib.rs
```diff
@@ -929,8 +930,16 @@ impl<I: Interner> Const<I> {
             ConstValue::InferenceVar(_) => false,
             ConstValue::Placeholder(_) => false,
             ConstValue::Concrete(_) => false,
+            // We assume for now that an unevaluated const expression doesn't contain bound vars
+            ConstValue::Unevaluated(_) => false,
```
jackh726, line 934: Hmm. Not sure that this assumption is correct. If the `Unevaluated` had an expression like `{N + 1}`, then there `N` would probably be a bound var and we would want to shift. But I haven't thought about this enough to know either way right now.

chalk-ir/src/lib.rs
```diff
@@ -942,6 +951,23 @@ pub struct ConstData<I: Interner> {
     pub value: ConstValue<I>,
 }
 
+impl<I: Interner> ConstData<I> {
+    /// Tries to evaluate the constant.
+    /// This is guaranteed to return a concrete constant or an error.
+    pub fn try_eval(&self, interner: &I) -> Result<ConstData<I>, ConstEvalError> {
```
jackh726, line 957: Why have this function? Instead of just `try_eval` on `Unevaluated`?

chalk-ir/src/lib.rs
jackh726: But I guess for now, this is probably fine. But I think I would like a followup issue filed with or without a `FIXME` here.

chalk-ir/src/lib.rs
```diff
@@ -979,6 +1013,60 @@ impl<I: Interner> ConcreteConst<I> {
...
+impl<I: Interner> UnevaluatedConst<I> {
+    /// Attempts to evaluate the const expression.
+    pub fn try_eval(&self, ty: &Ty<I>, interner: &I) -> Result<ConcreteConst<I>, ConstEvalError> {
+        Ok(ConcreteConst {
+            interned: interner.try_eval_const(&ty.interned, &self.interned)?,
+        })
+    }
+
+    /// Checks whether two unevaluated constants are equal.
+    pub fn const_eq(&self, ty: &Ty<I>, other: &UnevaluatedConst<I>, interner: &I) -> bool {
```
jackh726, line 1032: Don't think this is needed.

chalk-ir/src/lib.rs
```diff
@@ -979,6 +1013,60 @@ impl<I: Interner> ConcreteConst<I> {
...
+
+impl Debug for UnevaluatedConstData {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self.0 {
+            Some(n) => write!(fmt, "{}?", n)?,
+            None => write!(fmt, "?")?,
+        }
+        Ok(())
+    }
+}
```
jackh726, line 1051: I don't think this should be here. In `chalk-integration`, we should just `Option<u32>` directly for `InternedUnevaluatedConst`

chalk-ir/src/lib.rs
```diff
@@ -979,6 +1013,60 @@ impl<I: Interner> ConcreteConst<I> {
...
+
+/// An error that can occur when evaluating a const expression.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum ConstEvalError {
```
jackh726, line 1055: Probably also want an `Unevaluatable`. Though I don't exactly know how Chalk would handle these differently. (`TooGeneric` is obvious `Ambiguous`, but would `Unevaluable` return `Ambiguous` or `NoSolution`?)

chalk-parse/src/parser.lalrpop
```diff
@@ -610,3 +611,8 @@ LifetimeId: Identifier = {
 };
 
 ConstValue: u32 = <s:r"[0-9]+"> => u32::from_str_radix(s, 10).unwrap();
+
+UnevaluatedConst: Option<u32> = {
+    "?" => None,
+    <s:ConstValue> "?" => Some(<>),
+}
```
jackh726, line 618: Newline

tests/test/constants.rs
```diff
@@ -105,6 +122,99 @@ fn generic_impl() {
...
+
+            impl Trait for S<3?> {}
+        }
+
+        goal {
+            exists<const N> {
+                S<N>: Trait
+            }
+        } yields {
+            "Unique; substitution [?0 := 3?], lifetime constraints []"
```
jackh726, line 141: Ah, interesting. I have to think if this is something we want to be able to support. (Giving solutions with unevaluated consts). It's pretty clear that these would *have* to be evaluable, since they would have to pass WF/coherence checks.

tests/test/constants.rs
```diff
@@ -105,6 +122,99 @@ fn generic_impl() {
     }
 }
 
+#[test]
```
jackh726, line 125: A comment on each goal and what it's purpose is would be helpful.

tests/test/constants.rs
```diff
@@ -105,6 +122,99 @@ fn generic_impl() {
...
+        goal {
+            S<0>: Trait
+        } yields {
+            "Ambiguous; no inference guidance"
+        }
+
+        goal {
+            S<?>: Trait
+        } yields {
+            "Unique"
```
jackh726, line 204: See, this should be `Ambiguous`, for a similar reason that two `InferenceVar`s aren't the same: we don't actually know this would evaluate to the same const.

tests/test/constants.rs
```diff
@@ -105,6 +122,99 @@ fn generic_impl() {
...
+        }
+
+        // We don't know if the constant evaluates to an invalid value
+        goal {
+            exists<N> {
```
jackh726: Should this be `const N`?

chalk-ir/src/lib.rs
josh65536: Which is why I said "for now".

chalk-ir/src/lib.rs
josh65536: Oops, I had this for convenience before and some changes I made removed all the calls I made to this function.

tests/test/constants.rs
josh65536: I've been assuming that for chalk testing purposes that `?` evaluates to a specific but unknown constant. I don't feel like breaking reflexivity.

jackh726: But that's not true. The point here is that `?` represents a constant that can't be evaluated *yet*. Importantly, if you have two constants that can't be evaluated (e.g. `{N+1}` and `{X +1}`), then we can't say that if there is a solution that covers one case, there is a solution that covers all cases.

chalk-ir/src/lib.rs
josh65536: The only reason this exists is to change the output formatting to reflect the test input.

jackh726: Ah okay. So, having a newtype wrapper for `Option<u32>` is fine, but it should still go in `chalk-integration`.

tests/test/constants.rs
josh65536: I don't see the rest of the tests doing this, though

jackh726: A lot of tests haven't done this. Recently though, we have been trying to at least make sure *new* tests get comments.

josh65536: I added comments on the new tests

tests/test/constants.rs
josh65536: As it turns out, removing `const_eq` for unevaluated consts and assuming that two `?`s don't necessarily evaluate to the same value breaks `LocalImplAllowed(Baz: Foo<?>)` in the `unevaluated_const_unknown` coherence test.

jackh726: Oops, I never responded to this. Sorry about that.

So, I think theoretically we could always say that `impl Foo<?0> for Baz {}` never violates orphan rules, since both `Foo` and `Baz` are local. *However*, I think for unevaluated consts, it's okay to say it's `Ambiguous` if we get `TooGeneric`. I would have to look more into the orphan rules code to figure out how to best handle this. For this PR, we can mark a comment or `FIXME` and file a followup issue for more discussion.

tests/test/constants.rs
jackh726: Great thanks!

chalk-engine/src/slg/aggregate.rs
```diff
@@ -483,6 +483,44 @@ impl<I: Interner> AntiUnifier<'_, '_, I> {
                 }
             }
 
+            (ConstValue::Concrete(e), ConstValue::Unevaluated(u))
```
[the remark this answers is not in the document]
josh65536: I don't see why not. Chalk can return associated type projections in answers, and an unevalutated const is basically a projection.

tests/test/coherence.rs
```diff
@@ -517,17 +517,22 @@ fn orphan_check() {
...
-
-            impl Foo<?> for Baz {}
-        }
-    }
-}
+// FIXME: The following test fails by violation of the orphan rules.
+// This is because right now, unevaluated const expressions have a
+// temporary reflexivity hole, where ? != ?,
+// which messes with the SLG resolvent.
+// The test should be uncommented when the reflexivity hole is fixed.
```
jackh726, line 524: This isn't a reflexivity hole. `Unevaluated` won't ever equal `Unevaluated`. The only reason why we *might* be okay with allowing this test to pass orphan rules is because we that both `Foo` and `Baz` are local, so this should pass regardless of the value `?`. That's what this `FIXME` should be.

tests/test/coherence.rs
```diff
@@ -579,29 +584,35 @@ fn unevaluated_const_intersect() {
...
-            "overlapping impls of trait `Foo`"
-        }
-    }
-}
+// FIXME: The following test fails by violation of the orphan rules.
+// This is because right now, unevaluated const expressions have a
+// temporary reflexivity hole, where ? != ?,
+// which messes with the SLG resolvent.
+// The test should be uncommented when the reflexivity hole is fixed.
+// It is supposed to fail because of overlapping impls.
```
jackh726, line 592: See above. What error do we get here if we leave these uncommented? With these, we actually don't *know* that these are overlapping impls (could `impl Foo<3>` and `impl Foo<4>`). But this still shouldn't lower successfully.

tests/test/constants.rs
```diff
@@ -188,41 +188,55 @@ fn unevaluated_impl() {
...
-            }
-        } yields {
-            "Unique; substitution [?0 := ?], lifetime constraints []"
-        }
-    }
-}
+// FIXME: This won't even lower
+// because `impl Trait for S<?>` fails the orphan rules
+// because `LocalImplAllowed(S<?>: Trait)` can't be proven
+// because `?` != `?`.
```
jackh726, line 194: Should be able to disable coherence now (like in #569). Let's do that so we still can have some of these tests.

chalk-solve/src/infer/unify.rs
```diff
@@ -464,25 +464,22 @@ impl<'t, I: Interner> Unifier<'t, I> {
             }
 
             (&ConstValue::Unevaluated(ref expr1), &ConstValue::Unevaluated(ref expr2)) => {
```
jackh726, line 466: This isn't completely necessary, but what you could do (and in other places with this pattern) is something similar to the `normalize_shallow` we do 
```
let a_val = a_val.try_eval(a_ty, interner);
let b_val = b_val.try_eval(a_ty, interner);
```
which would just add a new function of `ConstValue` `try_eval`. Then this would deduplicate/cleanup this logic a bit.

tests/test/coherence.rs
josh65536: Right now it's `structural mismatch between answer '?' and pending goal '?'`. Is the resolvent allowed to create a `CannotProve` goal? Also, in the case where chalk can't prove that impls don't overlap, it assumes that they do.

chalk-ir/src/interner.rs
```diff
@@ -93,6 +93,15 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
...
     type InternedConcreteConst: Debug + Clone + Eq + Hash;
 
+    /// "Interned" representation of an unevaluated const expression. In normal user code,
+    /// `Self::InternedUnevaluatedConst` is not referenced. Instead,
+    /// we refer to `UnevaluatedConst<Self>`, which wraps this type.
+    ///
+    /// `InternedUnevaluatedConst` instances are not created by chalk,
+    /// it can only evaluate them with the [`try_eval_const`] method
+    /// and check for (syntactic for now) equality between them.
+    type InternedUnevaluatedConst: Debug + Clone + Eq + Hash;
```
nikomatsakis, line 103: Do we anticipate that this will carry a "substitutions"?

nikomatsakis: I expect not, right?

chalk-ir/src/interner.rs
```diff
@@ -456,14 +465,21 @@ pub trait Interner: Debug + Copy + Eq + Ord + Hash {
...
         c1: &Self::InternedConcreteConst,
         c2: &Self::InternedConcreteConst,
     ) -> bool;
 
+    /// Attempt to evaluate a const to a concrete const.
+    fn try_eval_const(
+        &self,
+        ty: &Self::InternedType,
+        constant: &Self::InternedUnevaluatedConst,
+    ) -> Result<Self::InternedConcreteConst, ConstEvalError>;
```
nikomatsakis, line 481: I kind of expect this to take a set of substitutions. Consider how the [rustc variant looks](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.ConstKind.html#variant.Unevaluated).

chalk-solve/src/infer/unify.rs
```diff
@@ -415,19 +420,26 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
-                debug!(?var, ty=?a, "unify_var_ty");
-                self.unify_var_const(var, a)
+                debug!(?var, ty=?a, "unify_var_const");
+                self.unify_var_const(var, &a)
             }
 
+            // Note that we assume that the unevaluated const doesn't contain inference variables.
+            // Also, for now, the unevalutable unevaluated const is too generic.
+            (&ConstValue::InferenceVar(_), &ConstValue::Unevaluated(_))
+            | (&ConstValue::Unevaluated(_), &ConstValue::InferenceVar(_)) => Ok(self.goals.push(
```
nikomatsakis, line 439: Hmm, I think I expected us to unify an inference variable here...

chalk-solve/src/infer/unify.rs
```diff
@@ -440,9 +452,23 @@ impl<'t, I: Interner> Unifier<'t, I> {
                 }
             }
 
+            // For now, the unevalutable unevaluated const is too generic.
+            (&ConstValue::Concrete(_), &ConstValue::Unevaluated(_))
+            | (&ConstValue::Unevaluated(_), &ConstValue::Concrete(_)) => Ok(self.goals.push(
```
nikomatsakis, line 457: ...and then to generate a subobligation here, kind of like associated type normalization. Is this planned for "future PRs" or is my idea of how to approach this outdated :)

tests/test/coherence.rs
```diff
@@ -516,3 +516,94 @@ fn orphan_check() {
...
+
+            impl Foo<3?> for Baz {}
+            impl Foo<3?> for Baz {}
+        } error_msg {
+            "overlapping impls of trait `Foo`"
+        }
+    }
+}
+
+// FIXME: The following test fails by structural mismatch of ? and ?
```
nikomatsakis, line 583: What does this FIXME mean exactly?

jackh726:
We discussed this in our [meeting Tuesday](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/meeting.202020-10-20/near/213976710). The conclusion was that the PR as-is isn't correct (see @nikomatsakis's review). Part of this is we need better direction for how consts will be handled.

I'm going to go ahead and close this for now. @josh65536 if you still want to work on this, feel free to reopen. Either Niko or I can answer questions on here or, preferably, Zulip. We really would like to get this over the finish line :)

PR #598, opened 2020-08-19
Properly unify consts with vars

### LINKED ISSUE #597: Exists-forall gives false solution for consts
Solving
```
exists<const C> {
    forall<const D> {
        C = D
    }
}
```
gives `Unique; substitution [?0 := !1_0], lifetime constraints []` instead of the expected `No possible solution`.

nathanwhit:
Before, we incorrectly unified consts with vars unconditionally.

Fixes #597.


Areredify:
Oh yeah, it was most certainly an oversight. I actually saw it when working on universe canonicalization but forgot to push it separately

jackh726:
@bors r+

book links are broken with latest rust directory moving around business, but we can leave that for another PR

PR #599, opened 2020-09-01
Refactor FnPointer and FnDef

### LINKED ISSUE #574: Refactor `FnPointer` and `FnDef` to share function signature logic
Right now we have a lot of code duplication in handling the ABI, safety, variadic-ness, and other information for function pointers and function definitions. We should encapsulate this in something like a `FnSig` struct, and use that in both `FnPointer` and `FnDef`.

memoryleak47:
Fixes #574.


memoryleak47:
Whoops, I didn't know about `cargo test --all`.
I'll look into it.

PR #600, opened 2020-09-02
Fix links in chalk book

nathanwhit:
A few links in the book were broken due to the directory changes in rustc, causing the link check to fail.


detrumi:
Thanks!

@bors r+

PR #601, opened 2020-09-04
Extern types

### LINKED ISSUE #595: Support extern opaque types.
```rust
#![feature(extern_types)]

extern "C" {
    type A;
    fn foo() -> Box<A>;
    fn bar();
}

impl A {
    fn bar(&self) {
        unsafe {bar()};
    }
}

fn main() {
    let a = unsafe { foo() } ;
    a.bar();
}
```
This code compiles fine in rustc, but rust-analyzer gives `Box<{unknown}>` inlay hint to a.

memoryleak47:
Closes  #595.


tests/test/extern_types.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[test]
+fn extern_ty_single() {
+    lowering_success! {
+        program {
+            extern { type A; }
+        }
+    }
+}
+
```
jackh726, line 26: Can you provide like a one-line comment on each of these for *why* this test exists (and passes as it is).
Honestly, it's probably:
`extern types are always well-formed`
and
`extern types don't implement any builtin traits`

jackh726 (approved):
Couple nits, but pretty straightforward!

PR #602, opened 2020-09-04
Simplify lowering

detrumi:
- Moved lowering functions that just take `Env` to a single `Lower` trait with associated `type Lowered`, to get rid of all the single-use traits. This retains `.lower(env)` chaining for the most important types.
- Changed lowering functions with extra params to freestanding functions. These could in principle be handled by the same trait, by making the argument generic, but that would make it less clear, and having lowering functions that take extra parameters named differenlty is a good thing, so they stand out.
- Same for lowering functions that don't take an environment. These often aren't chained anyways, and having names like  `lower_adt_repr` makes the code clearer IMO.
- Use macros to reduce boilerplate around `LowerParameterMap` (also removing some unused functions there) and `lower_type_kind` (renamed `OpaqueTyDefn.identifier` to `OpaqueTyDefn.name` to line things up better)
- Re-use `lookup_apply_type` in `ApplyTypeLookup` by changing it slightly, to avoid duplicating map lookup code

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -85,7 +85,7 @@ struct AssociatedTyLookup {
 }
 
 enum ApplyTypeLookup<'k> {
-    Param(&'k chalk_ir::WithKind<ChalkIr, BoundVar>),
+    Parameter(&'k chalk_ir::WithKind<ChalkIr, BoundVar>),
     Adt(AdtId<ChalkIr>),
     FnDef(FnDefId<ChalkIr>),
     Closure(ClosureId<ChalkIr>),
@@ -117,7 +117,7 @@ impl Env<'_> {
         };
 
         match self.lookup_apply_type(name) {
-            Ok(ApplyTypeLookup::Param(p)) => {
+            Ok(ApplyTypeLookup::Parameter(p)) => {
                 let b = p.skip_kind();
                 match &p.kind {
                     chalk_ir::VariableKind::Ty(_) => Ok(chalk_ir::TyData::BoundVar(*b)
@@ -165,7 +165,7 @@ impl Env<'_> {
 
     fn lookup_apply_type(&self, name: &Identifier) -> LowerResult<ApplyTypeLookup> {
         if let Some(id) = self.parameter_map.get(&name.str) {
-            return Ok(ApplyTypeLookup::Param(id));
+            return Ok(ApplyTypeLookup::Parameter(id));
         }
 
         if let Some(id) = self.adt_ids.get(&name.str) {
@@ -1357,7 +1357,7 @@ impl LowerWithEnv for Ty {
 
             Ty::Apply { name, ref args } => {
                 let (apply_name, k) = match env.lookup_apply_type(&name)? {
-                    ApplyTypeLookup::Param(_) => {
+                    ApplyTypeLookup::Parameter(_) => {
                         return Err(RustIrError::CannotApplyTypeParameter(name.clone()))
                     }
 
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -692,7 +692,9 @@ trait LowerTypeKind {
 }
 
 trait LowerParameterMap {
-    fn synthetic_parameters(&self) -> Option<chalk_ir::WithKind<ChalkIr, Ident>>;
+    fn synthetic_parameters(&self) -> Option<chalk_ir::WithKind<ChalkIr, Ident>> {
+        None
+    }
     fn declared_parameters(&self) -> &[VariableKind];
     fn all_parameters(&self) -> Vec<chalk_ir::WithKind<ChalkIr, Ident>> {
         self.synthetic_parameters()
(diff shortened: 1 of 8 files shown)


chalk-integration/src/lowering.rs
```diff
@@ -96,110 +97,80 @@ enum ApplyTypeLookup {
...
-                    actual: 0,
-                });
-            } else {
-                return Ok(chalk_ir::TyData::Apply(chalk_ir::ApplicationTy {
-                    name: chalk_ir::TypeName::Closure(*id),
-                    // See note in `program`. Unlike rustc, we store upvars separately.
-                    substitution: chalk_ir::Substitution::empty(interner),
-                })
+        if let Ok(lookup) = self.lookup_apply_type(name) {
+            match lookup {
```
jackh726, line 122: Maybe use a single match for this? Then the arms be `Ok(ApplyTypeLookup::Param(p))`, `Ok(ApplyTypeLookup::Adt(id))`,..., `Err(_)`

chalk-integration/src/lowering.rs
```diff
@@ -86,7 +86,8 @@ struct AssociatedTyLookup {
     addl_variable_kinds: Vec<chalk_ir::VariableKind<ChalkIr>>,
 }
 
-enum ApplyTypeLookup {
+enum ApplyTypeLookup<'k> {
+    Param(&'k chalk_ir::WithKind<ChalkIr, BoundVar>),
```
jackh726, line 90: Hmm, so. I'm not super sure about this name, because I don't want it to be confused with the rustc `Param` type.

chalk-integration/src/lowering.rs
```diff
@@ -307,395 +278,404 @@ impl<'k> Env<'k> {
...
+    let mut opaque_ty_ids = BTreeMap::new();
+    let mut adt_kinds = BTreeMap::new();
+    let mut fn_def_kinds = BTreeMap::new();
+    let mut fn_def_abis = BTreeMap::new();
+    let mut closure_kinds = BTreeMap::new();
+    let mut trait_kinds = BTreeMap::new();
+    let mut opaque_ty_kinds = BTreeMap::new();
+    let mut object_safe_traits = HashSet::new();
+
+    macro_rules! lower_type_kind {
```
jackh726, line 346: I don't know if I prefer *this*, or using a trait where the impls are generated by a macro.

jackh726: I think I might prefer the latter. Since there's less cognitive complexity *here* (you still would have `defn.lower_type_kind()`. And the "impls" would be like `lower_type_kind!(AdtDefn, Adt)`. (You could make `LowerTypeKind` have a `Self: LowerParameterMap` bound I think so you can do `self.all_parameters()` in the default function.

jackh726: Well, actually no for `TraitDefn` and `OpaqueTyDefn`. But you should still be able to do some macro trickery.

chalk-integration/src/lowering.rs
```diff
@@ -726,114 +706,34 @@ trait LowerParameterMap {
...
+lower_param_map!(AssocTyDefn, None);
+lower_param_map!(AssocTyValue, None);
+lower_param_map!(Clause, None);
+lower_param_map!(
+    TraitDefn,
+    Some(chalk_ir::WithKind::new(
+        chalk_ir::VariableKind::Ty(TyKind::General),
+        Atom::from(SELF),
+    ))
+);
```
jackh726, line 736: Beautiful.

chalk-integration/src/lowering.rs
```diff
@@ -2082,84 +1736,46 @@ impl<'k> LowerGoal<Env<'k>> for Goal {
...
-            Self::Sized => rust_ir::WellKnownTrait::Sized,
-            Self::Copy => rust_ir::WellKnownTrait::Copy,
-            Self::Clone => rust_ir::WellKnownTrait::Clone,
-            Self::Drop => rust_ir::WellKnownTrait::Drop,
-            Self::FnOnce => rust_ir::WellKnownTrait::FnOnce,
-            Self::FnMut => rust_ir::WellKnownTrait::FnMut,
-            Self::Fn => rust_ir::WellKnownTrait::Fn,
-            Self::Unsize => rust_ir::WellKnownTrait::Unsize,
-        }
+fn apply_result<T>(result: LowerResult<Vec<T>>) -> Vec<LowerResult<T>> {
```
jackh726, line 1755: This is a weird function. Wonder if this could be refactored away.

jackh726 (changes requested):
Some nits. But I think I still would prefer the `lower_*` functions to just implement `Lower` but not use the environment.

chalk-integration/src/lowering.rs
detrumi: Hmm, not sure what else to go with, since `BoundVar` would be confusing as well.

chalk-integration/src/lowering.rs
detrumi: I'll just inline it for now, couldn't really simplify it (surprisingly, there's no transpose for iterator/vec, and I couldn't find a nicer way)

detrumi:
> Some nits. But I think I still would prefer the `lower_*` functions to just implement `Lower` but not use the environment.

The problem is that some functions, like `lower_fn_abi`, are called before we even construct the empty environment.

jackh726:
I don't think having two separate traits is all that bad. We're looking for the solution that best conveys what's going on in the lowering as a whole. Having one or two traits implemented for most things makes more sense that one trait for some and a bunch of functions for everything else. (I'm curious if we could refactor away the arguments in all of the functions, but I'm purposely that out from this PR, because I would like to land us somewhere that is definitely better than get us stuck on what ifs).

I think we can always just return a `LowerResult`, even if it's always Ok wrapped. For this bit of Chalk, readability is more important that really anything else. If `lower()` and `lower(env)` is too confusing (I don't think so), then we could name them differently. 

Another thing worth thinking about is which of the functions *must* be called before the environment is constructed. Then, these could be special and everything else takes an env. (This somewhat depends on how much there are and such)

detrumi:
> I don't think having two separate traits is all that bad

:+1:

> I think we can always just return a `LowerResult`, even if it's always Ok wrapped

This is awkward when lowering vecs, so let's go with this instead, even if it's a bit inconsistent:
```
trait Lower {
    type Lowered;

    fn lower(&self) -> Self::Lowered;
}
```

> Another thing worth thinking about is which of the functions _must_ be called before the environment is constructed. Then, these could be special and everything else takes an env. (This somewhat depends on how much there are and such)

Turns out that the `fn_def_abis` part wasn't even needed, so this isn't a problem after all. Entry points like `lower_goal` are a bit trickier (because it ends with `goal.lower(&env)`), but maybe that's something to fix in another PR.

chalk-integration/src/lowering.rs
jackh726: I guess it ends up serving a similar purpose to rustc's `Param`, so if we can't think of a better name, `Param` is probably acceptable. Another option: `Parameter`. Feel free to take it or leave it.

chalk-integration/src/lowering.rs
```diff
@@ -96,110 +95,77 @@ enum ApplyTypeLookup {
...
+            )
             .intern(interner)
-            .cast(interner));
-        }
-
-        if let Some(_) = self.trait_ids.get(&name.str) {
-            return Err(RustIrError::NotStruct(name.clone()));
+            .cast(interner)),
+            Err(_) => {
+                if let Some(id) = self.foreign_ty_ids.get(&name.str) {
```
jackh726, line 150: So, how did this end up on this path (I would expect it to be returned in `lookup_apply_type`)? Is it just because these can't have generics?

jackh726 (approved):
Just a couple more minor comments. But this looks great. Feel free to merge with/without comments addressed.

chalk-integration/src/lowering.rs
detrumi: I guess so, since there's no `foreign_ty_kinds`. Possibly `lookup_apply_type` and `lookup_generic_arg` should just use their own lookups, but it's kind of awkward to have all these separate maps.

detrumi:
Wait, that didn't merge it? Huh, I don't really get how bors works

jackh726:
> Wait, that didn't merge it? Huh, I don't really get how bors works

I think there's some bors weirdness happening right now.

PR #603, opened 2020-09-06
Extend well-known traits by Unpin

memoryleak47:
I read the RFC and didn't find any special rules concerning Unpin, so this one is rather straight-forward. Correct me if I missed something!
cc #363


jackh726 (changes requested):
I don't think this is right. I think `Unpin` needs to be implemented for all the built in types. Std has impls for all scalars, ref, raws; but we need builtin impls for function pointers, function defs, arrays, `str`, closures, slices(?).

jackh726:
Also, please update the well known traits table in the book.

memoryleak47:
I thought this was a consequence of defining it as an auto trait. I'll look into it.

jackh726 (approved):
LGTM! Pretty straightfoward until we get generators and auto traits for builtin types.

PR #606, opened 2020-09-08
More lowering improvements

detrumi:


chalk-integration/src/lowering/program_lowerer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+impl ProgramLowerer {
+    pub fn next_item_id(&mut self) -> RawId {
+        let index = self.next_item_index;
+        self.next_item_index += 1;
+        RawId { index }
+    }
+
+    /// Create ids for associated type declarations and values
+    pub fn find_associated_types(
```
jackh726, line 50: Naming could be better here. Maybe `extract_associated_types` or `insert_associated_types`.

chalk-integration/src/lowering/program_lowerer.rs
```diff
@@ -0,0 +1,428 @@
...
+                    }
+                }
+
+                _ => {}
+            }
+        }
+        Ok(())
+    }
+
+    pub fn lookup_ids(&mut self, program: &Program, raw_ids: &Vec<RawId>) -> LowerResult<()> {
```
jackh726, line 86: Naming again, `extract_ids`?

chalk-integration/src/lowering.rs
```diff
@@ -922,47 +268,50 @@ impl LowerWithEnv for LeafGoal {
...
-    adt_defn: &AdtDefn,
-    adt_id: chalk_ir::AdtId<ChalkIr>,
-    env: &Env,
-) -> LowerResult<rust_ir::AdtDatum<ChalkIr>> {
-    if adt_defn.flags.fundamental && adt_defn.all_parameters().len() != 1 {
-        Err(RustIrError::InvalidFundamentalTypesParameters(
-            adt_defn.name.clone(),
-        ))?;
-    }
+impl LowerWithEnv for (&AdtDefn, chalk_ir::AdtId<ChalkIr>) {
```
jackh726, line 271: I like this solution :)

chalk-integration/src/lowering.rs
```diff
@@ -1185,8 +529,7 @@ impl LowerWithEnv for QuantifiedInlineBound {
 
     fn lower(&self, env: &Env) -> LowerResult<Self::Lowered> {
         let variable_kinds = self.variable_kinds.iter().map(|k| k.lower());
-        let binders = env.in_binders(variable_kinds, |env| Ok(self.bound.lower(env)?))?;
-        Ok(binders)
+        env.in_binders(variable_kinds, |env| Ok(self.bound.lower(env)?))
```
jackh726, line 532: I think you can remove the `Ok` wrapping here.

chalk-integration/src/lowering/env.rs
```diff
@@ -0,0 +1,271 @@
...
+            Ok(ApplyTypeLookup::Opaque(id)) => Ok(chalk_ir::TyData::Alias(
+                chalk_ir::AliasTy::Opaque(chalk_ir::OpaqueTy {
+                    opaque_ty_id: id,
+                    substitution: chalk_ir::Substitution::empty(interner),
+                }),
+            )
+            .intern(interner)
+            .cast(interner)),
+            Err(_) => {
+                if let Some(id) = self.foreign_ty_ids.get(&name.str) {
```
jackh726, line 140: Is there anything stopping us from just making `lookup_apply_type` be able to return `Foreign`? It would clean this up and make it more uniform.

jackh726 (changes requested):
Some comments, but r=me after.

chalk-integration/src/lowering/env.rs
detrumi: What about the call in lowering `Ty` then? That doesn't need to match on `Foreign`

jackh726: I mean, we could either just `panic` there since we know we don't expect it.

chalk-integration/src/lowering/env.rs
```diff
@@ -0,0 +1,269 @@
...
+///          // addl_variable_kinds
+/// }
+/// ```
+#[derive(Debug, PartialEq, Eq)]
+pub struct AssociatedTyLookup {
+    pub id: chalk_ir::AssocTypeId<ChalkIr>,
+    pub addl_variable_kinds: Vec<chalk_ir::VariableKind<ChalkIr>>,
+}
+
+pub enum TypeLookup<'k> {
```
jackh726, line 73: I would argue this isn't *quite* `TypeLookup` since `Trait` isn't a `type`, but eh. This is probably fine.

jackh726 (approved):

jackh726:
@detrumi please r=jackh726 instead of merging directly

PR #607, opened 2020-09-12
Add CoerceUnsized builtin rules

Areredify:
So uhh either Im missing an obvious implementation option or we are missing a necessary abstraction to do things in a reasonable manner. The problem I ran into is that there is no way to extend sequent signature and clauses *and* query the solver about them. Consider this:
```
let forall_goals: Binders<Goal<I>> = ...;

gb.forall(forall_goals, |goals| {
    // here, the binders of `forall_goals` are "in environment", but to query the solver, you need to 
    // return from `GoalBuilder::forall` and call the solver with the result, you can't query the solver from inside.
    // So, if you want to branch depending on the result of solving the `forall_goal` inside the `GoalBuilder::forall`,
    // you need to return (popping binders from environment), call the solver, and then reenter binders again. 
    // At this point, it's easier to manipulate bound vars by hand (what I did in the pr).
})
```
I think an an abstraction similar to goal builder where you can call `forall` and `implies` (and these alone) to push new things to the environment, and then call solver with the goals about vars inside the environment, would be nice, (but Im not sure).


jackh726 (approved):
This LGTM! One small nit, take or leave. I'm not sure if there's any particular solve tests we need? I suppose not since this is just WF.

Areredify:
cc #363

PR #608, opened 2020-09-12
fix debug print of pointer types

Areredify:
Fixes confusing debug print of pointer types, previously `&'a T` it was printed as `Not<!0_0, T>` and `*const T` was printed as `Not<T>`, which was not ideal.

### DIFF
--- a/chalk-ir/src/debug.rs
+++ b/chalk-ir/src/debug.rs
@@ -186,8 +186,22 @@ impl<I: Interner> Debug for TypeName<I> {
             TypeName::OpaqueType(opaque_ty) => write!(fmt, "!{:?}", opaque_ty),
             TypeName::Slice => write!(fmt, "{{slice}}"),
             TypeName::FnDef(fn_def) => write!(fmt, "{:?}", fn_def),
-            TypeName::Raw(mutability) => write!(fmt, "{:?}", mutability),
-            TypeName::Ref(mutability) => write!(fmt, "{:?}", mutability),
+            TypeName::Ref(mutability) => write!(
+                fmt,
+                "{}",
+                match mutability {
+                    Mutability::Mut => "{{&mut}}",
+                    Mutability::Not => "{{&}}",
+                }
+            ),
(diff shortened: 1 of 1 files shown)


Areredify:
While we are at it, is there a specific reason for application types to be printed as ugly as they are now, or I can go ahead and change them to be rendered "properly"?

detrumi (changes requested):
I think we should add a `impl Debug for Mutability` instead, to avoid doing this for each place where it can occur.

> While we are at it, is there a specific reason for application types to be printed as ugly as they are now, or I can go ahead and change them to be rendered "properly"?

I'm assuming it's there because implementing Debug directly doesn't work, but if we don't need it then feel free to simplify it.

Areredify:
>I think we should add a impl Debug for Mutability
I am not sure how to do that, it renders to different things when in different places, i.e. `Mutability::Not` in `Ref` renders to `{{&}}` and to `{{*const}}` in `RawPtr`.

By ugly I didn't mean the implementation, I meant that printed types are in `{type name}{substitution}` format, instead of the natural way to render it, i.e. `{{array}}<u32, 3>` instead of `[u32; 3]`.

detrumi:
Ah right, did not notice that.
And yeah, improving the application ty formatting would be great

detrumi (approved):

jackh726:
So, yeah, debug printing of `Application` tys could be improved (see the [changes](https://github.com/rust-lang/rust/blob/bb0067c75ea82efa8a09213bfba0dc2583ac0a4c/compiler/rustc_middle/src/traits/chalk.rs#L105) I made in the rustc integration). The only problem is this can't be done *generically*, since it relies on unpacking the substitution *without* having access to an `interner` instance. But we can easily do this for the `chalk-integration` `Interner` impl.

jackh726:
@Areredify feel free to followup with this, if you want. Or just file an issue for someone else to followup.

In the meantime: @bors r=jackh726,detrumi

PR #609, opened 2020-09-13
Variance

daboross:
This contains @super-tuple and my work on rebasing @jackh726's variance work from #520 and implementing the generalizer on top of it.

As mentioned in Zulip, we've been trying to port rustc tests for the generalizer to chalk, in an effort to see what fails. Since we've been failing at this for so long, we decided to just try to write the generalizer instead! We don't fully understand it, but have enough of an idea to write something that hopefully works.

This fails 5 of the existing tests, and that's part of the motivation for opening this now - we aren't sure how to proceed on those. We have not written any new tests for variance nor the generalizer.

The last one we've debugged is `dyn_lifetime_bound`, which fails because when unifying our generalized value and the original ty value, the code tries to unify a inference var with universe 4, created freshly for our generalized value, with a placeholder with universe 6 - and that fails to pass the `OccursCheck`. Here's the debug logs we generated for that failed test: https://gist.github.com/daboross/24f1d2310bd86e1d542678c4ebd550e0

### DIFF
--- a/chalk-engine/src/normalize_deep.rs
+++ b/chalk-engine/src/normalize_deep.rs
@@ -107,6 +107,20 @@ mod test {
 
     const U0: UniverseIndex = UniverseIndex { counter: 0 };
 
+    // We just use a vec of 20 `Invariant`, since this is zipped and no substs are
+    // longer than this
+    #[derive(Debug)]
+    struct TestDatabase;
+    impl UnificationDatabase<ChalkIr> for TestDatabase {
+        fn fn_def_variance(&self, _fn_def_id: FnDefId<ChalkIr>) -> Variances<ChalkIr> {
+            Variances::from(&ChalkIr, [Variance::Invariant; 20].iter().copied())
+        }
+
+        fn adt_variance(&self, _adt_id: AdtId<ChalkIr>) -> Variances<ChalkIr> {
+            Variances::from(&ChalkIr, [Variance::Invariant; 20].iter().copied())
+        }
+    }
+
     #[test]
     fn infer() {
         let interner = &ChalkIr;
@@ -115,14 +129,28 @@ mod test {
         let a = table.new_variable(U0).to_ty(interner);
         let b = table.new_variable(U0).to_ty(interner);
         table
-            .unify(interner, &environment0, &a, &ty!(apply (item 0) (expr b)))
+            .relate(
+                interner,
+                &TestDatabase,
+                &environment0,
+                Variance::Invariant,
+                &a,
+                &ty!(apply (item 0) (expr b)),
+            )
             .unwrap();
         assert_eq!(
             DeepNormalizer::normalize_deep(&mut table, interner, &a),
             ty!(apply (item 0) (expr b))
         );
         table
-            .unify(interner, &environment0, &b, &ty!(apply (item 1)))
+            .relate(
+                interner,
+                &TestDatabase,
+                &environment0,
+                Variance::Invariant,
+                &b,
+                &ty!(apply (item 1)),
+            )
             .unwrap();
         assert_eq!(
             DeepNormalizer::normalize_deep(&mut table, interner, &a),
--- a/chalk-engine/src/normalize_deep.rs
+++ b/chalk-engine/src/normalize_deep.rs
@@ -138,10 +138,13 @@ mod test {
                 &ty!(apply (item 0) (expr b)),
             )
             .unwrap();
-        assert_eq!(
-            DeepNormalizer::normalize_deep(&mut table, interner, &a),
-            ty!(apply (item 0) (expr b))
-        );
+        // FIXME: can't just assert these are equal because the inference var gets set to a new var (?2)
+        // which is unified with b (?1). It might be good to pick a root to return. For now,
+        // this doesn't work as-is.
+        //assert_eq!(
+        //    DeepNormalizer::normalize_deep(&mut table, interner, &a),
+        //    &ty!(apply (item 1)),
+        //);
         table
             .relate(
                 interner,
--- a/chalk-engine/src/normalize_deep.rs
+++ b/chalk-engine/src/normalize_deep.rs
@@ -108,7 +108,6 @@ mod test {
     use super::*;
     use chalk_integration::interner::ChalkIr;
     use chalk_integration::{arg, ty};
-    use chalk_solve::logging::with_tracing_logs;
 
     const U0: UniverseIndex = UniverseIndex { counter: 0 };
 
--- a/chalk-engine/src/normalize_deep.rs
+++ b/chalk-engine/src/normalize_deep.rs
@@ -55,7 +55,11 @@ where
                 .assert_ty_ref(interner)
                 .fold_with(self, DebruijnIndex::INNERMOST)?
                 .shifted_in(interner)), // FIXME shift
-            None => Ok(var.to_ty(interner, kind)),
+            None => {
+                // Normalize all inference vars which have been unified into a
+                // single variable. Ena calls this the "root" variable.
+                Ok(self.table.inference_var_root(var).to_ty(interner, kind))
+            }
         }
     }
 
@@ -104,6 +108,7 @@ mod test {
     use super::*;
     use chalk_integration::interner::ChalkIr;
     use chalk_integration::{arg, ty};
+    use chalk_solve::logging::with_tracing_logs;
 
     const U0: UniverseIndex = UniverseIndex { counter: 0 };
 
@@ -138,13 +143,16 @@ mod test {
                 &ty!(apply (item 0) (expr b)),
             )
             .unwrap();
-        // FIXME: can't just assert these are equal because the inference var gets set to a new var (?2)
-        // which is unified with b (?1). It might be good to pick a root to return. For now,
-        // this doesn't work as-is.
-        //assert_eq!(
-        //    DeepNormalizer::normalize_deep(&mut table, interner, &a),
-        //    &ty!(apply (item 1)),
-        //);
+        // a is unified to Adt<#0>(c), where 'c' is a new inference var
+        // created by the generalizer to generalize 'b'. It then unifies 'b'
+        // and 'c', and when we normalize them, they'll both be output as
+        // the same "root" variable. However, there are no guarantees for
+        // _which_ of 'b' and 'c' becomes the root. We need to normalize
+        // "b" too, then, to ensure we get a consistent result.
+        assert_eq!(
+            DeepNormalizer::normalize_deep(&mut table, interner, &a),
+            ty!(apply (item 0) (expr DeepNormalizer::normalize_deep(&mut table, interner, &b))),
+        );
         table
             .relate(
                 interner,
--- a/chalk-solve/src/infer/unify.rs
+++ b/chalk-solve/src/infer/unify.rs
@@ -158,35 +158,15 @@ impl<'t, I: Interner> Unifier<'t, I> {
             (_, &TyKind::Alias(ref alias)) => self.relate_alias_ty(variance.invert(), alias, a),
             (&TyKind::Alias(ref alias), _) => self.relate_alias_ty(variance, alias, b),
 
-            // FIXME: needs to handle relating a var and ty; needs generalization
-            // Relating an inference variable with a non-inference variable.
-            (a_data @ &TyKind::InferenceVar(_, _), b_data @ _)
-            | (a_data @ _, b_data @ &TyKind::InferenceVar(_, _)) => {
-                // Correct variance when we flip a/b.
-                let (new_variance, var, kind, ty_data) = match (a_data, b_data) {
-                    (&TyKind::InferenceVar(var, kind), ty_data @ _) => {
-                        (variance, var, kind, ty_data)
-                    }
-                    (ty_data @ _, &TyKind::InferenceVar(var, kind)) => {
-                        // var is grabbed from 'b', but given in as the left
-                        // parameter to relate_var_ty, so we need to flip variance.
-                        (variance.invert(), var, kind, ty_data)
-                    }
-                    _ => unreachable!(),
-                };
-
+            (&TyKind::InferenceVar(var, kind), ty_data @ _) => {
                 let ty = ty_data.clone().intern(interner);
-                match (kind, ty.is_integer(interner), ty.is_float(interner)) {
-                    // General inference variables can unify with any type
-                    (TyVariableKind::General, _, _)
-                    // Integer inference variables can only unify with integer types
-                    | (TyVariableKind::Integer, true, _)
-                    // Float inference variables can only unify with float types
-                    | (TyVariableKind::Float, _, true) => {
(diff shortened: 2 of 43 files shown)


jackh726:
I added the ability to set the variances of adts and fn_defs in tests. E.g.
```
#[variance(Invariant, Covariant)]
struct Foo<A,B> {}
```

chalk-solve/src/infer/unify.rs
```diff
@@ -244,33 +271,116 @@ impl<'t, I: Interner> Unifier<'t, I> {
         {
```
matthewjasper, line 271: The above comment is not entirely correct with subtyping. It should be
```
for<'a...> T == for<'b...> U

if

for<'a...> exists<'b...> T :> U &&
for<'b...> exists<'a...> T <: U
```
and
```
for<'a...> T <: for<'b...> U

if

for<'b...> exists<'a...> T <: U
```

The code here should look like this

```rust
        if let Variance::Invariant | Variance::Contavariant = variance {
            let a_universal = self.table.instantiate_binders_universally(interner, a);
            let b_existential = self.table.instantiate_binders_existentially(interner, b);
            Zip::zip_with(self, Variance::Contavariant, &a_universal, &b_existential)?;
        }

        if let Variance::Invariant | Variance::Covariant = variance {
            let b_universal = self.table.instantiate_binders_universally(interner, b);
            let a_existential = self.table.instantiate_binders_existentially(interner, a);
            Zip::zip_with(self, Variance::Covariant, &a_existential, &b_universal)
        }
```

jackh726:
Did some work on this and rebased.

daboross:
Thank you for working on this! 

I've pushed a rebase onto 30f4a12b6ad945cd60701fd657d44a05e9219c3c, and currently figuring out everything that needs to be adjusted to rebase onto https://github.com/rust-lang/chalk/pull/629.

jackh726:
Rebased this :)

chalk-solve/src/infer/unify.rs
```diff
@@ -523,36 +1001,56 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
-        self.goals.push(InEnvironment::new(
-            self.environment,
-            WhereClause::LifetimeOutlives(LifetimeOutlives { a: b, b: a }).cast(self.interner),
-        ));
+    fn push_lifetime_eq_goals(&mut self, variance: Variance, a: Lifetime<I>, b: Lifetime<I>) {
+        debug!(
+            "pushing lifetime eq goals for a={:?} b={:?} with variance {:?}",
+            a, b, variance
+        );
+        if matches!(variance, Variance::Invariant | Variance::Covariant) {
```
matthewjasper, line 1009:
```suggestion
        if matches!(variance, Variance::Invariant | Variance::Contravariant) {
```
`'a <: 'b` is taken to mean `'b: 'a` (the other `matches` also needs changing)

nikomatsakis (commented):
OK, so, I have been reading the code locally. It seems good. I have to go now. I came here to comment on `push_lifetime_eq_goal` but I see that @matthewjasper has already made the comments I came here to make. Locally I have a few tests I was toying with adding and I may rebase and/or push those commits tomorrow.

chalk-engine/src/normalize_deep.rs
```diff
@@ -119,14 +133,31 @@ mod test {
...
-            DeepNormalizer::normalize_deep(&mut table, interner, &a),
-            ty!(apply (item 0) (expr b))
-        );
+        // FIXME: can't just assert these are equal because the inference var gets set to a new var (?2)
+        // which is unified with b (?1). It might be good to pick a root to return. For now,
+        // this doesn't work as-is.
+        //assert_eq!(
+        //    DeepNormalizer::normalize_deep(&mut table, interner, &a),
+        //    &ty!(apply (item 1)),
+        //);
```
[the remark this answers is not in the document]
daboross: This should be fixed! We've left a comment regarding the problem next to the new assertion.

tests/test/subtype.rs
```diff
@@ -121,6 +152,160 @@ fn multi_lifetime() {
...
+                exists<U> {
+                    Subtype(Foo<&'a u32>, Foo<U>),
+                    Subtype(Foo<&'b u32>, Foo<U>)
+                }
+            }
+        } yields {
+            // Because A is invariant, we require the lifetimes to be equal
+            "Unique; substitution [?0 := (&'!1_1 Uint(U32))], lifetime
+            constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_1 }, \
+            InEnvironment { environment: Env([]), goal: '!1_1: '!1_0 }]"
```
jackh726, line 242: Hmm, I wonder if it would be better to introduce a new variable `?U1` where both `'!1_0` and `!1_1` are equal. I guess, logically, they are the same; and this is what we get from short-circuiting the `Invariant` case.

tests/test/subtype.rs
```diff
@@ -121,6 +152,160 @@ fn multi_lifetime() {
...
+        } yields {
+            // Result should be identical to multi_lifetime result.
+            "Unique; for<?U1> { substitution [?0 := (&'^0.0 Uint(U32))], lifetime constraints [\
+            InEnvironment { environment: Env([]), goal: '!1_0: '^0.0  }, \
+            InEnvironment { environment: Env([]), goal: '!1_1: '^0.0  }] }"
+        }
+    }
+}
+
+/// Tests that the generalizer generalizes lifetimes within tuples correctly.
```
jackh726, line 293:
```diff
-Tests that the generalizer generalizes lifetimes within tuples correctly.
+Tests that the generalizer generalizes lifetimes within arrays correctly.
```

tests/test/subtype.rs
```diff
@@ -72,20 +72,49 @@ fn ref_lifetime_variance() {
...
+        goal {
+            Subtype(for<'a, 'b> fn(&'a u32, &'b u32) -> &'a u32, for<'a> fn(&'a u32, &'a u32) -> &'a u32)
+        } yields {
+            // TODO: are these results actually correct?
             "Unique; for<?U1,?U1> { substitution [], lifetime constraints [\
-            InEnvironment { environment: Env([]), goal: '!1_0: '^0.0 }, \
-            InEnvironment { environment: Env([]), goal: '!1_0: '^0.1 }] }"
+                InEnvironment { environment: Env([]), goal: '!1_0: '^0.0 }, \
+                InEnvironment { environment: Env([]), goal: '!1_0: '^0.1 }, \
+                InEnvironment { environment: Env([]), goal: '^0.0: '!1_0 }  \
```
jackh726: These are...weird. I would have expected something like: `'!1_0: '^0.0, '!1_1: '1_0, '!1_0: '1_1`.

The constraints here seem...backward?

tests/test/subtype.rs
```diff
@@ -72,20 +72,49 @@ fn ref_lifetime_variance() {
...
+            ]}"
+        }
+        goal {
+            Subtype(for<'a> fn(&'a u32, &'a u32) -> &'a u32, for<'a, 'b> fn(&'a u32, &'b u32) -> &'a u32)
+        } yields {
+            "Unique; for<?U1> { substitution [], lifetime constraints [\
+                InEnvironment { environment: Env([]), goal: '!1_0: '^0.0 }, \
+                InEnvironment { environment: Env([]), goal: '!1_1: '^0.0 }, \
+                InEnvironment { environment: Env([]), goal: '^0.0: '!1_0 }  \
```
jackh726, line 112: Similarly here; these seem backward.

tests/test/subtype.rs
daboross: :+1:

Didn't double-check these before pushing - there are a few mistakes in this test and the tuple test which we should push fixes for shortly.

tests/test/subtype.rs
matthewjasper: They looks correct to me, `fn` types are contravariant in their parameters.

tests/test/subtype.rs
```diff
@@ -0,0 +1,314 @@
...
+    test! {
+        program {
+        }
+
+        goal {
+            for<'a, 'b> fn(&'a u32, &'b u32) = for<'a> fn(&'a u32, &'a u32)
+        } yields {
+            "Unique;substitution [], lifetime constraints [\
+            InEnvironment { environment: Env([]), goal: '!1_0: '!1_1 }, \
+            InEnvironment { environment: Env([]), goal: '!1_1: '!1_0 }]"
```
matthewjasper, line 85: This doesn't look right, these constraints are not solvable.

chalk-solve/src/infer/unify.rs
```diff
@@ -274,38 +383,356 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
+        // for<'b...> exists<'a...> T <: U
+
         let interner = self.interner;
 
-        {
+        if let Variance::Invariant | Variance::Contravariant = variance {
             let a_universal = self.table.instantiate_binders_universally(interner, a);
             let b_existential = self.table.instantiate_binders_existentially(interner, b);
-            Zip::zip_with(self, &a_universal, &b_existential)?;
+            Zip::zip_with(self, variance, &a_universal, &b_existential)?;
```
matthewjasper, line 397:
```suggestion
            Zip::zip_with(self, Variance::Contravariant, &a_universal, &b_existential)?;
```
(And `Covariant` below). This is required for the tests I've commented on to give the correct results.

super-tuple:
Thank you for the comments!

We've found and fixed a bug where `relate_ty_ty` was incorrectly inverting the variance by re-ordering the parameters when passing them into `relate_var_ty`, and we've added additional tests to catch similar inversion bugs.

We haven't looked at the `FnPointer` problem yet.

tests/test/subtype.rs
jackh726: 🤦‍♂️

chalk-solve/src/infer/unify.rs
```diff
@@ -57,86 +62,129 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
                 a, b
             ),
 
             // Unifying an alias type with some other type `U`.
-            (_, &TyKind::Alias(ref alias)) => self.unify_alias_ty(alias, a),
-            (&TyKind::Alias(ref alias), _) => self.unify_alias_ty(alias, b),
+            (_, &TyKind::Alias(ref alias)) => self.relate_alias_ty(variance.invert(), alias, a),
+            (&TyKind::Alias(ref alias), _) => self.relate_alias_ty(variance, alias, b),
+
+            // FIXME: needs to handle relating a var and ty; needs generalization
```
jackh726, line 161: This FIXME can be removed now.

chalk-solve/src/infer/unify.rs
```diff
@@ -57,86 +62,129 @@ impl<'t, I: Interner> Unifier<'t, I> {
...
+                    (&TyKind::InferenceVar(var, kind), ty_data @ _) => {
+                        (variance, var, kind, ty_data)
+                    }
+                    (ty_data @ _, &TyKind::InferenceVar(var, kind)) => {
+                        // var is grabbed from 'b', but given in as the left
+                        // parameter to relate_var_ty, so we need to flip variance.
+                        (variance.invert(), var, kind, ty_data)
+                    }
+                    _ => unreachable!(),
+                };
```
jackh726, line 176: It would probably be cleaner to just have two different match arms here...

The code below can probably just be moved into `relate_var_ty` and make it return a `Result` so we can move the interner/float check into it

chalk-solve/src/infer/unify.rs
jackh726: *Is* `Contravariant` here right? I don't think so, since this also covers the `Invariant` case.

tests/test/subtype.rs
jackh726: But these are? This just says that `'a == 'b`

tests/test/subtype.rs
```diff
@@ -0,0 +1,686 @@
...
+                InEnvironment { environment: Env([]), goal: '!1_0: '!1_1 }, \
+                InEnvironment { environment: Env([]), goal: '!1_1: '!1_0 }  \
+            ]"
+        } yields[SolverChoice::slg_default()] {
+            // Because A is invariant, we require the lifetimes to be equal
+            "Unique; substitution [?0 := (&'!1_1 Uint(U32))], lifetime constraints [\
+                InEnvironment { environment: Env([]), goal: '!1_0: '!1_1 }, \
+                InEnvironment { environment: Env([]), goal: '!1_1: '!1_0 }  \
+            ]"
+        }
```
jackh726, line 312: Interesting that these are different between the solvers

super-tuple:
We found another bug, which was caused by inconsistent use of `push_lifetime_outlives_goals`. Specifically, the code that relates lifetimes in structs assumed the function behaved the opposite way to what it was doing. We updated the function to behave the way the struct code expected (which we believe is more intuitive), and then updated the code that relates references to work with the inverted behavior. In more detail, we updated `push_lifetime_outlives_goals` to say that `a` is a subtype of `b`, and therefore when called with `covariant`, it pushes constraints that `a` outlives `b`. We also had to fix an incorrect test case once we'd fixed the bug.

We also applied the code quality suggestions Jack Huey made, and removed some duplicate tests.

Additionally, we looked at the constraints emitted by `fn_lifetime_variance_with_return_type`, and I don't think any of them are _wrong_ although I don't know whether the output is meaningful. I think this PR is ready for review.

chalk-solve/src/infer/unify.rs
matthewjasper: In the invariant case this `if` uses `Contravariant` and the other one uses `Covariant`. This is required so that `T == U` is the same as `T <: U` and `U <: T`.

tests/test/subtype.rs
matthewjasper: These are placeholders, so relating them is an error. The requirements here should be `!1_0: '?0`, `!1_1: ?0`, `'!2_0: '?1`, `'!2_0: '?2` (the same as subtyping in both directions).

matthewjasper:
The lifetime variance was correct as it was. It is a bit strange that `Contravariant` is the "normal" variance for lifetimes, but it is what rustc is using.

tests/test/subtype.rs
jackh726: But we have allowed unifying placeholders? (On master, [here](https://github.com/rust-lang/chalk/blob/65daa08f07fcea1314c0bffd731e1fef1b76d619/chalk-solve/src/infer/unify.rs#L385) is unification, [here](https://github.com/rust-lang/chalk/blob/65daa08f07fcea1314c0bffd731e1fef1b76d619/tests/test/existential_types.rs#L310) an example where two lifetime placeholders are equated).

matthewjasper: yes, but region inference will error on such constraints.

jackh726:
I reverted the lifetime variance change and made `relate_binders` zip with `Contravariant` and `Covariant`

nikomatsakis (approved):

nikomatsakis:
r=me but it needs a rebase!

PR #611, opened 2020-09-17
Remove Context trait from chalk-engine

jackh726:
This is unneeded now that chalk-engine knows about chalk-solve. The only reason we might want this is for abstraction, but I don't think that really gets us very far here.

There's more that could be done here, for sure. But this has been sitting locally for a while and figured it's a good enough place as any to PR.


nikomatsakis (approved):

nikomatsakis:
This looks really good. I kinda' miss the old traits but I think concrete code will ultimately be much easier for folks to read.

nikomatsakis:
r=me to land @jackh726

jackh726:
> I kinda' miss the old traits but I think concrete code will ultimately be much easier for folks to read.

I definitely understand this sentiment. It's sort of nice to have an abstraction. But I think ultimately it isn't necessary anymore and makes it harder to understand.

@bors r=nikomatsakis

PR #616, opened 2020-09-30
Support fundamental types with multiple type parameters

### LINKED ISSUE #615: Implement clauses to allow multi-parameter fundamental types
See discussion here: https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/meeting.202020-09-29/near/211676604

Needed for rust-lang/rust#77187

For fundamental types like `Box<T, U>`, we should generate rules:

```
forall<T, U> { IsLocal(Box<T>) :- IsLocal(T) }
forall<T, U> { IsLocal(Box<T>) :- IsLocal(U) }
forall<T, U> { IsUpstream(Box<T, U>) :- IsUpstream(T), IsUpstream(U) }
forall<T, U> { DownstreamType(Box<T, U>) :- DownstreamType(T). }
forall<T, U> { DownstreamType(Box<T, U>) :- DownstreamType(U). }
```

nathanwhit:
Closes #615.

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -402,28 +402,17 @@ impl<I: Interner> ToProgramClauses<I> for AdtDatum<I> {
                 ($goal:ident) => {
                     // Fundamental types must always have at least one
                     // type parameter for this rule to make any
-                    // sense. We currently do not have have any
-                    // fundamental types with more than one type
-                    // parameter, nor do we know what the behaviour
-                    // for that should be. Thus, we are asserting here
-                    // that there is only a single type parameter
-                    // until the day when someone makes a decision
-                    // about how that should behave.
-                    assert_eq!(
-                        self_appl_ty.len_type_parameters(interner),
-                        1,
-                        "Only fundamental types with a single parameter are supported"
+                    // sense.
+                    assert!(
+                        self_appl_ty.len_type_parameters(interner) >= 1,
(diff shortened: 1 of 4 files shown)


chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -402,28 +402,17 @@ impl<I: Interner> ToProgramClauses<I> for AdtDatum<I> {
...
-                    builder.push_clause(
-                        DomainGoal::$goal(self_ty.clone()),
-                        Some(DomainGoal::$goal(
-                            // This unwrap is safe because we asserted
-                            // above for the presence of a type
-                            // parameter
-                            self_appl_ty.first_type_parameter(interner).unwrap(),
-                        )),
+                    // sense.
+                    assert!(
```
jackh726, line 406: Maybe this is preexisting. But more applicable now that we only use it twice. But I feel like the macro isn't super necessary. We can do an assert if fundamental. And just write the goal out in the two places. Would also reduce the asserts from two to one, though I'm not super worried about that.

jackh726 (approved):
Looks good, I think. Want a second look from @nikomatsakis though.

jackh726:
Also, my PR broke the links in the book. I'll make a PR for that separately.

jackh726:
This looks right and Niko didn't get to it this week. I don't want this issue to block the rustc PR, so I'm going to merge this so it gets into next release.

@bors r+

PR #617, opened 2020-10-03
Add static lifetime

jackh726:


tests/test/statics.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use super::*;
+
+
+#[test]
+fn static_lowering() {
+    lowering_success! {
+        program {
+            struct A<'a> where 'a: 'static {}
+            trait A<'a> where 'a: 'static {}
```
matthewjasper, line 10: Seeing a name collision here is somewhat surprising, even if it's allowed. Can the struct and trait have different names?

matthewjasper (commented):
Looks good. Just one a nit.

tests/test/statics.rs
jackh726: Whoops, that's unintended.

PR #620, opened 2020-10-04
Fix mdbook links after chalk-engine changes

detrumi:
This broke a while back in https://github.com/rust-lang/chalk/pull/611,
but only now broke the build because the hosted docs were updated.

CC @jackh726


detrumi:
@jackh726 The description for the [`context module`](https://rust-lang.github.io/chalk/chalk_engine/context/index.html) is outdated as well, should we just move `AnswerResult` and `AnswerStream` to `lib.rs`?

jackh726:
@detrumi chalk-engine probably needs a cleanup, yeah

r=me

Issue #614, opened 2020-09-28
Make reproducing test(s) for failing to generalize program clauses


nathanwhit:
If program clauses are not properly generalized it can cause issues in the recursive solver. This was the cause of the recent [breakage in rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/issues/6078). We should have a test that covers this to prevent future issues.

[Discussed briefly here](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/Recent.20breakage.20in.20recursive.20solver).

detrumi:
Minimal example:
```
fn main() {
    foo(|a: u32| {});
}

fn foo<F>(f: F)
where
    F: Send,
{
}
```

Trying to fit this in a test is tricky though, as the problem triggers with this goal:
```
Implemented({closure:ClosureId(0)}<for<0> Safe () [?0 := Uint(U32), ?1 := ^1.0]>: Send)
```
The syntax has changed slightly since, but this is a `TyKind::Closure` with the substitution being a `FnPointer` (`for<0> Safe () [?0 := Uint(U32), ?1 := ^1.0]`).

I don't think there's currently a way to trigger closure substitutions in chalk tests. My attempts so far:
```
test! {
        program {
            #[lang(fn_once)]
            trait FnOnce<Args> {
                type Output;
            }

            #[auto] trait Send { }

            fn foo<F>(f: F) -> ()
                where F: Send;

            closure bar(self, a: u32) {}
        }

        goal {
            <bar as FnOnce<(foo<(bar, u32)>)>>: Send // Not valid test syntax
        } yields {
            "Unique"
        }
    }
```

flodiebold:
Yes, I've had similar problems reproducing problems with closures from RA. Maybe the closure support in the test harness needs to be more flexible somehow. I've usually been able to reproduce the problems by just modeling the closure as a normal struct.

PR #624, opened 2020-10-09
Pass canonical goals to program clauses code in SLG solver

### LINKED ISSUE #568: Decide whether the program clauses code should work on canonical or instantiated goals
Currently, the SLG solver and the recursive solver disagree about what to pass to the program clauses code: The SLG solver instantiates the goal and then calls the program clauses code for it; the recursive solver calls it with the canonical goal. (This means that in the same goal, there will be either inference variables or bound variables depending on the solver.) There are advantages to both, but we should decide which way to go because it complicates the program clauses code to handle both.

nathanwhit:
It seemed like we were leaning towards this solution in the [design meeting](https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/meeting.202020-09-22), and it turned out (assuming these changes are right) to be pretty straightforward.

Closes #568.

This should stop seeing issues related to failing to generalize program clauses pop up, but I'm only going to cc #614 since I'd like some tests maybe.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -269,12 +269,7 @@ impl<I: Interner> Forest<I> {
                     },
                     universes: goal.universes,
                 };
-                let clauses = program_clauses_for_goal(
-                    program,
-                    &canon_domain_goal.canonical.value.environment,
-                    &canon_domain_goal.canonical.value.goal,
-                    &canon_domain_goal.canonical.binders,
-                );
+                let clauses = program_clauses_for_goal(program, &canon_domain_goal);
                 let (infer, subst, InEnvironment { environment, goal }) =
                     chalk_solve::infer::InferenceTable::from_canonical(
                         context.program().interner(),
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -12,8 +12,8 @@ use crate::{
 
 use chalk_ir::interner::Interner;
 use chalk_ir::{
-    AnswerSubst, Canonical, CanonicalVarKinds, ConstrainedSubst, FallibleOrFloundered, Floundered,
-    Goal, GoalData, InEnvironment, NoSolution, Substitution, UCanonical, UniverseMap,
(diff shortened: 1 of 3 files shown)


jackh726 (commented):
I'm on mobile, so not going to be a perfect response here. But this looks...not that great. I mean, it looks *correct*, but we went from instantiating in one place to two places, plus that rewrapping of the domain goal in the canonical binders and environment.

I guess it would nice if we could clean that up somehow. If you don't want to do that in this PR, I would like a FIXME at least and then r=me.

chalk-engine/src/logic.rs
```diff
@@ -303,6 +309,13 @@ impl<I: Interner> Forest<I> {
...
 
             _ => {
+                let (infer, subst, InEnvironment { environment, goal }) =
+                    chalk_solve::infer::InferenceTable::from_canonical(
+                        context.program().interner(),
+                        goal.universes,
+                        &goal.canonical,
+                    );
+                let mut infer = TruncatingInferenceTable::new(context.max_size(), infer);
                 // `canonical_goal` is a goal. We can simplify it
```
nikomatsakis, line 319: pre-existing, but "canonical goal is a goal" is a pretty mystifying sentence. I guess what I meant is "canonical goal is a meta-goal" or something like that. In other words, it's a goal whose handling is built-in to the solver, versus a domain goal that is defined by rules. I forget the term for this, I think metagoal is right but...

nikomatsakis (commented):
One thing that is also missing from this PR is that we could cleanup the `program_clauses_from_goal` code to error out if it encounters inference variables. We're also going to have to *extend* that code once we start canonicalizing universal variables.

nikomatsakis:
Overall this seemed pretty nice to me. I still think that it's the right setup -- I guess my reasoning is basically that I think we ought to be caching the results of those calls, so passing in canonical things seems correct.

jackh726:
Review comments have been addressed and this should be good to merge.

nikomatsakis (approved):

PR #625, opened 2020-10-12
Don't add wf clauses for function args

### LINKED ISSUE #556: Incorrect elaboration of `WellFormed` clauses for functions
For `Wellformed` goals for functions, we push a `WellFormed` clause for each of the functions arguments here: https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/clauses.rs#L548

This incorrectly handles bound vars. Take the following goal as an example:
```
WellFormed(for<3> [?0 := Not<'^0.0, ^1.0>])
```

The following clauses would get pushed:
```
forall<> WellFormed(for<3> [?0 := Not<'^0.0, ?0>])
forall<> WellFormed(Not<'^`1.0, ?0>)
```

I don't even think we *need* to push the `Wellformed` clauses for the parameters.

jackh726:
Closes #556

We might want to add this back in *correctly* at some point, but it does need some design work anyways (in terms of what WF functions mean). For now, this is broken and not needed for any tests. So remove it.

### DIFF
--- a/chalk-solve/src/clauses.rs
+++ b/chalk-solve/src/clauses.rs
@@ -755,12 +755,6 @@ fn match_ty<I: Interner>(
             .to_program_clauses(builder, environment),
         TyData::Function(quantified_ty) => {
             builder.push_fact(WellFormed::Ty(ty.clone()));
-            quantified_ty
-                .substitution
-                .iter(interner)
-                .map(|p| p.assert_ty_ref(interner))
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):

nikomatsakis:
As we said in the chalk meeting last week, I believe, we can't really descend through binders this way, and I think that for backwards compatibility and other reasons we don't really *want* to.

PR #626, opened 2020-10-12
Handle cached answers correctly in any_future_answer

matthewjasper:
This is required for `1 + 2` to typeck correctly.


tests/test/numerics.rs
```diff
@@ -225,3 +225,26 @@ fn integers_are_sized() {
         }
     }
 }
+
+/// Simplified version of a goal that needs to be solved for typ checking
```
nikomatsakis, line 229:
```suggestion
/// Simplified version of a goal that needs to be solved for type checking
```

chalk-engine/src/logic.rs
```diff
@@ -117,12 +117,15 @@ impl<I: Interner> Forest<I> {
     pub(super) fn any_future_answer(
         &self,
         table: TableIndex,
-        answer: AnswerIndex,
+        mut answer_index: AnswerIndex,
         mut test: impl FnMut(&Substitution<I>) -> bool,
     ) -> bool {
-        if let Some(answer) = self.tables[table].answer(answer) {
+        while let Some(answer) = self.tables[table].answer(answer_index) {
```
nikomatsakis, line 123: We should write a better comment here, I think. Maybe something like:

First, check the cached answers, starting from `answer_index`. Then check whether the partial values for strands, as they may produce future cached answers.

nikomatsakis (approved):
Left a nit but looks good to me

jackh726:
@matthewjasper can you `@bors r+` instead, in the future? (Not sure if you have the perms, but we should probably add you to the wg-traits team *anyways*)

PR #630, opened 2020-10-22
Document notation

### LINKED ISSUE #561: Add a "notation" section to the glossary of the book
While we talk about bound vars, inference vars, and placeholders in different parts of the book and sometimes introduce notation, we should have a single consolidated section in the glossary to specifically mention some notation, such as:
```
?0 - Inference Var
^0, ^1.0 - Bound var
!0 - Placeholder
A :- B - A if B
```

Maybe others

camelid:
Fixes #561.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -21,7 +21,7 @@
     - [Opaque types (impl Trait)](./clauses/opaque_types.md)
     - [Well known traits](./clauses/well_known_traits.md)
     - [Well-formedness checking](./clauses/wf.md)
-    - [Coherence](./coherence.md)
+    - [Coherence](./clauses/coherence.md)
 - [Canonical queries](./canonical_queries.md)
     - [Canonicalization](./canonical_queries/canonicalization.md)
 - [Chalk engine](./engine.md)
--- a/book/src/glossary.md
+++ b/book/src/glossary.md
@@ -6,12 +6,12 @@ This is a glossary of terminology (possibly) used in the chalk crate.
 
 ### Basic notation
 
-| Notation     | Meaning                          |
-|--------------|----------------------------------|
-| `?0`         | [Type inference variable]        |
-| `^0`, `^1.0` | [Bound variable]                 |
-| `!0`         | [Placeholder]                    |
-| `A :- B`     | [Clause]; A is true if B is true |
+| Notation     | Meaning                                 |
+|--------------|-----------------------------------------|
+| `?0`         | [Type inference variable]               |
+| `^0`, `^1.0` | [Bound variable]; bound in a [`forall`] |
+| `!0`, `!1.0` | [Placeholder]                           |
+| `A :- B`     | [Clause]; A is true if B is true        |
 
 ### Rules
 
@@ -26,6 +26,7 @@ This is a glossary of terminology (possibly) used in the chalk crate.
 
 [Type inference variable]: ../types/rust_types.md#inference-variables
 [Bound variable]: ../types/rust_types.md#bound-variables
+[`forall`]: #debruijn-index
 [Placeholder]: ../types/rust_types.md#placeholders
 [Clause]: ../clauses/goals_and_clauses.md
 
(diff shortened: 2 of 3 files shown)


book/src/glossary.md
```diff
@@ -2,6 +2,33 @@
...
 
+## Notation
+
+### Basic notation
+
+| Notation     | Meaning                          |
+|--------------|----------------------------------|
+| `?0`         | [Type inference variable]        |
+| `^0`, `^1.0` | [Bound variable]                 |
+| `!0`         | [Placeholder]                    |
```
camelid, line 13: I have to say that this notation confuses me a bit; I just copied this part from the issue, so I would appreciate if someone more knowledgeable could double-check this :)

camelid:
Not sure why CI is failing...

book/src/SUMMARY.md
```diff
@@ -21,7 +21,6 @@
     - [Opaque types (impl Trait)](./clauses/opaque_types.md)
     - [Well known traits](./clauses/well_known_traits.md)
     - [Well-formedness checking](./clauses/wf.md)
-    - [Coherence](./coherence.md)
```
jackh726, line 24: Why remove this?

jackh726 (approved):
These LGTM, other than the deletion of coherence?

book/src/glossary.md
jackh726: These look right. If these confuse you though, they most certainly confuse someone else. So, what confuses you?

book/src/SUMMARY.md
camelid: Because the page doesn't exist; it's auto-filled by mdBook to this:

image

See rust-lang/mdbook#1364.

jackh726: Ah oops. It should be `./clauses/coherence.md`. That's an error from #623

book/src/glossary.md
camelid: Well, for one the syntax confuses me: what does `1.0` mean? It looks a floating point number, which I'm not used to seeing to represent a variable. I *think* it's a de Bruijn index, but I still don't fully understand it.

Also, I don't understand the difference between a placeholder and a bound variable/inference variable; the docs for it don't explicitly describe the difference.

book/src/SUMMARY.md
camelid: Fixed.

camelid:
Looks like `rust,ignore` is not working for some reason: it's still reporting doc test failures.

book/src/glossary.md
jackh726: Yeah, so maybe it's worth *describing* the syntax here a bit. I'll describe a little bit here. I'm not quite sure the best place to put this though. (Or how much might be duplicated from other places in the book)

Inference variables have the format `?0`. This means that in the current inference context, this is variable `0`.

Bound variables can take the format of `^1.0` or `^0`. In the former case, the `1` represents the Debruijn index of the variable. The `0` represents the index in that level of binding. For example, in `forall<X,Y> { forall <Z> { X } }`, `X` could be represented as `^1.0`. If the Debruijn index is `0`, then it may sometimes be omitted, leading the the second format `^0`.

Placeholders take the format `!1_0` or `!0`. For the former, the `1` is universe index of the placeholder whereas the `0` is the index in that universe. If the universe is `0`, then it can be omitted, leading to the second format.


I'm not sure if this is the right place to explain placeholders vs bound or inference vars. I would have to think of a good explanation anyways. Let's just focus on notation here, for now.

jackh726:
So, looks like a bunch of stuff in `coherence.md` isn't correct. Feel free to either 1) fix the errors with `notrust` or 2) just change the SUMMARY back and ignore the coherence doc for now.

I think the `rust,ignore` is working for the glossary.

book/src/glossary.md
camelid: Thanks for the description! I'll add some of that to the docs.

camelid: What's an example of a rule where a placeholder would be created?

camelid:
Okay, CI should pass now! (except for linkcheck)

jackh726 (approved):
This looks nice! Thanks!

PR #633, opened 2020-10-23
Add `DiscriminantKind` builtin trait

Areredify:


tests/test/wf_lowering.rs
```diff
@@ -1438,3 +1438,21 @@ fn coerce_unsized_struct() {
         }
     }
 }
+
+#[test]
+fn no_discriminant_kind_impls() {
```
jackh726, line 1443: Can we move this to `discriminant_kind` test file? Would like to start putting related tests together.

tests/test/discriminant_kind.rs
```diff
@@ -0,0 +1,138 @@
+use super::*;
```
jackh726, line 1: Can you write some comments here for what each of the test goals are supposed to be for?

jackh726 (changes requested):
I think this looks good overall. Book will need updating, and I had a couple comments on the tests.

Unfortunately, since I don't know a ton about this, I'd like a second set of eyes. @nikomatsakis?

Areredify:
cc #363

chalk-solve/src/clauses/builtin_traits/discriminant_kind.rs
```diff
@@ -0,0 +1,69 @@
...
+        | TyKind::Alias(_)
+        | TyKind::BoundVar(_)
+        | TyKind::Placeholder(_)
+        | TyKind::AssociatedType(..)
+        | TyKind::Error
+        | TyKind::InferenceVar(..) => false,
+    };
+
+    if !can_determine_discriminant {
+        return;
```
nikomatsakis, line 41: I'm not sure if this is the right rule here. This will mean that we *can't* prove the `DiscriminantKind`, but I feel like we may want to return an "ambiguous" result instead.  It's not that there *is no* discriminant kind, it's that we don't know what it is.

nikomatsakis: I guess the thing to do is to make it .. uh .. flounder for these sorts of self types.

nikomatsakis (changes requested):

jackh726:
Review comment has been addressed so this should be good to review

nikomatsakis (approved):

PR #638, opened 2020-10-27
Clauses cleanup

matthewjasper:
- Add basic clauses for trait objects being wf
- Push clauses for auto traits on `TyKind::OpaqueType`
- Use a single match on the `TyKind` for Implemented goals
- Normalize `TyKind::Alias` self types


jackh726 (approved):
This LGTM, I think. I'm a bit surprised these cases weren't covered in some form before. r=me or maybe @nikomatsakis might want a look

PR #644, opened 2020-10-30
Change Fold and Folder traits to take values

### LINKED ISSUE #642: Refactor Fold to take by value
Currently, the [`Folder` trait](https://github.com/rust-lang/chalk/blob/80747d9147a77dae4e16b7abcfa2e169b036cb85/chalk-ir/src/fold.rs#L66) takes folded values by reference. As part of the [shared type library plan](https://github.com/rust-lang/wg-traits/issues/16), the fold traits should take their folded values by value, to match [the `TypeFolder`](https://github.com/rust-lang/rust/blob/ffe52882ed79be67344dd6085559e308241e7f60/compiler/rustc_middle/src/ty/fold.rs#L45) trait in rustc.

For example, this
```
fn fold_ty(&mut self, ty: &Ty<I>, outer_binder: DebruijnIndex) -> Fallible<Ty<TI>>
```
should be changed to this:
```
fn fold_ty(&mut self, ty: Ty<I>, outer_binder: DebruijnIndex) -> Fallible<Ty<TI>>
```
Same thing with the other folding functions.
Lastly, code that calls folding functions should pass the values by value, instead of by reference.

lnicola:
Fixes #642

### DIFF
--- a/chalk-ir/src/fold/boring_impls.rs
+++ b/chalk-ir/src/fold/boring_impls.rs
@@ -12,22 +12,23 @@ use std::sync::Arc;
 impl<'a, T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for &'a T {
     type Result = T::Result;
     fn fold_with<'i>(
-        &self,
-        folder: &mut dyn Folder<'i, I, TI>,
-        outer_binder: DebruijnIndex,
+        self,
+        _folder: &mut dyn Folder<'i, I, TI>,
+        _outer_binder: DebruijnIndex,
     ) -> Fallible<Self::Result>
     where
         I: 'i,
         TI: 'i,
     {
-        (**self).fold_with(folder, outer_binder)
+        unreachable!();
+        // (*self).fold_with(folder, outer_binder)
     }
 }
 
 impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Vec<T> {
     type Result = Vec<T::Result>;
     fn fold_with<'i>(
-        &self,
+        self,
         folder: &mut dyn Folder<'i, I, TI>,
         outer_binder: DebruijnIndex,
     ) -> Fallible<Self::Result>
@@ -44,7 +45,7 @@ impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Vec<T> 
 impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Box<T> {
     type Result = Box<T::Result>;
     fn fold_with<'i>(
-        &self,
+        self,
         folder: &mut dyn Folder<'i, I, TI>,
         outer_binder: DebruijnIndex,
     ) -> Fallible<Self::Result>
@@ -52,14 +53,14 @@ impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Box<T> 
         I: 'i,
         TI: 'i,
     {
-        Ok(Box::new((**self).fold_with(folder, outer_binder)?))
+        Ok(Box::new((*self).fold_with(folder, outer_binder)?))
     }
 }
 
 impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Arc<T> {
     type Result = Arc<T::Result>;
     fn fold_with<'i>(
-        &self,
+        self,
         folder: &mut dyn Folder<'i, I, TI>,
         outer_binder: DebruijnIndex,
     ) -> Fallible<Self::Result>
@@ -67,7 +68,7 @@ impl<T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for Arc<T> 
         I: 'i,
         TI: 'i,
     {
-        Ok(Arc::new((**self).fold_with(folder, outer_binder)?))
+        Ok(Arc::new(self.as_ref().fold_with(folder, outer_binder)?))
     }
 }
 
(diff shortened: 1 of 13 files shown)


chalk-ir/src/fold/boring_impls.rs
```diff
@@ -12,22 +12,23 @@ use std::sync::Arc;
...
+        self,
         folder: &mut dyn Folder<'i, I, TI>,
         outer_binder: DebruijnIndex,
     ) -> Fallible<Self::Result>
     where
         I: 'i,
         TI: 'i,
     {
-        (**self).fold_with(folder, outer_binder)
+        unimplemented!();
```
lnicola, line 23: Heads up: I think implementation should be removed, but I haven't quite figured how to do that.

chalk-ir/src/fold/boring_impls.rs
```diff
@@ -322,15 +323,15 @@ impl<I: Interner, TI: TargetInterner<I>> SuperFold<I, TI> for ProgramClauseData<
...
+        self,
         folder: &mut dyn Folder<'i, I, TI>,
         outer_binder: DebruijnIndex,
     ) -> ::chalk_ir::Fallible<Self::Result>
     where
         I: 'i,
         TI: 'i,
     {
-        let clause = self.data(folder.interner());
+        let clause = self.data(folder.interner()).clone();
```
lnicola, line 334: Heads up: `clone()` here from borrowed data.

chalk-solve/src/infer/ucanonicalize.rs
```diff
@@ -80,7 +81,7 @@ pub trait UniverseMapExt {
     fn add(&mut self, universe: UniverseIndex);
     fn map_universe_to_canonical(&self, universe: UniverseIndex) -> Option<UniverseIndex>;
     fn map_universe_from_canonical(&self, universe: UniverseIndex) -> UniverseIndex;
-    fn map_from_canonical<T, I>(
+    fn map_from_canonical<T: Clone, I>(
```
lnicola, line 84: Heads up: new `Clone` bounds.

detrumi: Yeah, no way around this

chalk-ir/src/fold/boring_impls.rs
detrumi: So this is a bit of a weird case, since `self.fold_with()` would call this impl recursively.
This should be `self.clone.fold_with()` instead.

lnicola: I think `(*self).clone()`, but it fails to build because of some missing `Clone` bounds.

chalk-ir/src/fold/boring_impls.rs
```diff
@@ -9,25 +9,25 @@ use crate::*;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
-impl<'a, T: Fold<I, TI>, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for &'a T {
+impl<'a, T: Fold<I, TI> + Clone, I: Interner, TI: TargetInterner<I>> Fold<I, TI> for &'a T {
```
jackh726, line 12: Where is this impl actually used?

chalk-recursive/src/fulfill.rs
```diff
@@ -84,7 +84,7 @@ pub(super) trait RecursiveInferenceTable<I: Interner> {
         T: Fold<I>,
         T::Result: HasInterner<Interner = I>;
 
-    fn u_canonicalize<T>(
+    fn u_canonicalize<T: Clone>(
```
jackh726, line 87: Instead, should we just take `value0` by value?

chalk-solve/src/infer/ucanonicalize.rs
jackh726: Well, if we take `canonical_value` by value...

chalk-ir/src/fold/boring_impls.rs
jackh726: Hmm.

chalk-ir/src/fold/boring_impls.rs
jackh726: This might be hiding a lot of cloning.

lnicola: It does, which is why I was thinking about removing it. I'm not sure where it's used, but replacing the code with `unreachable!()` makes most of the tests fail.

chalk-solve/src/infer/ucanonicalize.rs
lnicola: Good poing, updated, but there are still two `clone()`s needed in `merge_answer_into_strand`.

chalk-ir/src/fold/boring_impls.rs
lnicola:
```
   1: chalk_ir::fold::boring_impls::<impl chalk_ir::fold::Fold<I,TI> for &T>::fold_with
             at ./chalk-ir/src/fold/boring_impls.rs:23:9
   2: <T as chalk_ir::fold::shift::Shift<I>>::shifted_in_from
             at ./chalk-ir/src/fold/shift.rs:32:9
   3: <T as chalk_ir::fold::shift::Shift<I>>::shifted_in
             at ./chalk-ir/src/fold/shift.rs:28:9
   4: chalk_solve::coherence::solve::<impl chalk_solve::coherence::CoherenceSolver<I>>::disjoint::{{closure}}
             at ./chalk-solve/src/coherence/solve.rs:104:26
```

lnicola:
Closing this for now. I'm still not sure how to make it work, and I haven't had much time for it lately :disappointed:.

jackh726:
@lnicola thanks for trying regardless

lnicola:
I dropped the changes adding extra `Clone` bounds and rebased, just in case someone wants to take it from here.

PR #647, opened 2020-11-01
Make max goal size for recursive solver configurable

flodiebold:


jackh726:
@bors r+

This is going to make us have to rebase the variance PR, but oh well.

PR #649, opened 2020-11-06
Fix "var_universe invoked on bound variable" crash

flodiebold:
This should fix rust-analyzer/rust-analyzer#5495.

`unify_ty_ty` calls `normalize_ty_shallow` on both types and then assumes that they can't be bound variables. With integer type variables, this assumption was broken, because you could have a general type variable resolving to an integer type variable resolving to i32. To fix this, `normalize_ty_shallow` probes twice, to make sure to fully resolve the variable.

### DIFF
--- a/chalk-solve/src/infer.rs
+++ b/chalk-solve/src/infer.rs
@@ -119,6 +119,15 @@ impl<I: Interner> InferenceTable<I> {
     }
 
     pub fn normalize_ty_shallow(&mut self, interner: &I, leaf: &Ty<I>) -> Option<Ty<I>> {
+        // An integer/float type variable will never normalize to another
+        // variable; but a general type variable might normalize to an
+        // integer/float variable. So we potentially need to normalize twice to
+        // get at the actual value.
+        self.normalize_ty_shallow_inner(interner, leaf)
+            .map(|ty| self.normalize_ty_shallow_inner(interner, &ty).unwrap_or(ty))
+    }
+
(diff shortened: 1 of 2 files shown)


jackh726:
So, I think this problem (or a related one) came up in the variance PR. (See [this commit](https://github.com/rust-lang/chalk/pull/609/commits/2d82a433fd63a7f242a20764456ad862ff9b64e1) as a fix).

I guess, I'm a little bit unsure on if the "root" variable also corresponds to the resolved type. In that commit, we're worried about always getting back the same *inference var*. For this PR, I guess I'm somewhat worried about the arbitrary normalize *twice*.

This is definitely something I want @nikomatsakis to comment on, because I don't know enough about ena and the semantics of probing for variables.

flodiebold:
That's a different problem. The case in the commit you reference is that multiple inference variables of the same kind (i.e. general, or integer, or float) that don't have a value yet are unified. Here, it's about when e.g. an integer variable is unified with a general variable. That works differently: variables of different kind aren't simply unified as variables, but rather the general type variable's value is set to the integer variable (just like if we were unifying with a concrete type instead of an integer variable). That's why a general type variable can never resolve to another general type variable, but it can resolve to an integer or float variable.

[Here's](https://github.com/rust-lang/rust/blob/f2ea2f648e117013b0217f001088ae89e0f163ca/compiler/rustc_infer/src/infer/mod.rs#L1526-L1539) the same code in rustc (it just recurses instead of normalizing twice, but the effect should be the same).

jackh726:
Thanks for the clarification!

@bors r+

PR #655, opened 2020-11-20
Make Variances Debug

flodiebold:
Necessary for e.g. putting them into a Salsa DB.


flodiebold:
Of course this isn't that easy :grimacing:

flodiebold:
Ok, this should be better.

PR #659, opened 2020-11-27
Always relate with Invariant to non-General inference vars

### LINKED ISSUE #658: Regression with unification within fn (?)
See https://github.com/rust-analyzer/rust-analyzer/issues/6628.

I've got this down to the following Chalk test:
```rust
#[test]
fn ambiguous_unification_in_fn() {
    test! {
        program {
            trait FnOnce<Args> {
                type Output;
            }

            struct MyClosure<T> {}
            impl<T> FnOnce<(T,)> for MyClosure<fn(T) -> ()> {
                type Output = ();
            }
        }
        goal {
            exists<int T, U> {
                MyClosure<fn(&'static U) -> ()>: FnOnce<(&'static T,)>
            }
        } yields {
            "Unique"
        }
    }
}
```

This succeeded before #609 and now fails with an ambiguous result.

CC @jackh726  @daboross

jackh726:
Fixes #658 

An alternative way to approach this is to always unify inference variables as `Invariant`, which is how it's implemented in rustc right now. We should think about if we want that long term though (and possibly come up with tests that hit that).

### DIFF
--- a/chalk-engine/src/simplify.rs
+++ b/chalk-engine/src/simplify.rs
@@ -79,11 +79,6 @@ impl<I: Interner> Forest<I> {
                     Err(_) => return FallibleOrFloundered::NoSolution,
                 },
                 GoalData::SubtypeGoal(goal) => {
-                    if goal.a.inference_var(context.program().interner()).is_some()
-                        && goal.b.inference_var(context.program().interner()).is_some()
-                    {
-                        return FallibleOrFloundered::Floundered;
-                    }
                     match infer.relate_tys_into_ex_clause(
                         context.program().interner(),
                         context.unification_database(),
@@ -93,8 +88,13 @@ impl<I: Interner> Forest<I> {
                         &goal.b,
                         &mut ex_clause,
                     ) {
-                        Ok(()) => {}
-                        Err(_) => return FallibleOrFloundered::NoSolution,
+                        FallibleOrFloundered::Ok(_) => {}
+                        FallibleOrFloundered::Floundered => {
+                            return FallibleOrFloundered::Floundered
+                        }
+                        FallibleOrFloundered::NoSolution => {
+                            return FallibleOrFloundered::NoSolution
+                        }
                     }
                 }
                 GoalData::DomainGoal(domain_goal) => {
--- a/chalk-engine/src/slg.rs
+++ b/chalk-engine/src/slg.rs
@@ -213,7 +213,7 @@ pub trait UnificationOps<I: Interner> {
         a: &Ty<I>,
         b: &Ty<I>,
         ex_clause: &mut ExClause<I>,
-    ) -> Fallible<()>;
+    ) -> FallibleOrFloundered<()>;
 }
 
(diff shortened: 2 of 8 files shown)


matthewjasper:
Rustc doesn't unify inference variables with subtype bounds:

https://github.com/rust-lang/rust/blob/b776d1c3e3db8befabb123ebb1e46c3531eaed46/compiler/rustc_infer/src/infer/sub.rs#L87-L112

Looking at unification, the only difference here with rustc in this area is that `{int_var} <: {ty_var}` is turned into `{int_var} = {ty_var}` (and likewise with floats, and with the subtype relation reversed).

jackh726:
> Looking at unification, the only difference here with rustc in this area is that `{int_var} <: {ty_var}` is turned into `{int_var} = {ty_var}` (and likewise with floats, and with the subtype relation reversed).

Ugh yeah, that's what I was missing and what the proper change is.

jackh726:
Actually, that alone is not enough to make the test pass :/

matthewjasper:
I'm not sure if the test case with two arbitrary variables needs to pass. One of the variables will eventually be resolved if the program is going to compile.

jackh726:
I wondered about that. I did test with the original test from @flodiebold with the `int T` and that didn't pass.

nikomatsakis:
I don't love this solution. Like @matthewjasper, I'm not sure that the test with two *arbitrary variables* needs to pass. @jackh726 can you share what you tried with integer variables? In rustc, we definitely consider integer variables "more specific" than normal variables (so e.g. a general variable would be "assigned" the value of an integer variable).

jackh726:
Yeah, I updated.

jackh726:
So, to elaborate on why only always unifying non-general inference vars doesn't work:
First, we start off trying to unify our goal `Implemented(MyClosure<"rust" for<0> [?0 := (&'static ?0), ?1 := 0]>: FnOnce<1<(&'static ?1i)>>)` with our (instantiated) clause `Implemented(MyClosure<"rust" for<0> [?0 := ?2, ?1 := 0]>: FnOnce<1<?2>>)`.

From there, we try to unify `MyClosure<"rust" for<0> [?0 := (&'static ?0), ?1 := 0]>` and `MyClosure<"rust" for<0> [?0 := ?2, ?1 := 0]>` first. When `?2` tries to get unified with `(&'static ?0)`, it is generalized and set to `(&'?3 ?4)`. From the next unification, we get `?4 < ?0` (and the reverse too, since both `Covariant` and `Contravariant`, since we're in `Invariant`).

By this point, when we unify `1<(&'static ?1i)>` and `1<?2>`, it's too late, and we've already added the subtype goals.

jackh726:
Okay, so we probably just need to actually resolve inference vars when trying to simplify a subtype goal...

jackh726:
Okay so that worked.

nikomatsakis (approved):

nikomatsakis:
oops, I forgot, we use bors now don't we... d'oh

PR #660, opened 2020-12-04
Fold by value, not by reference

### LINKED ISSUE #642: Refactor Fold to take by value

ecstatic-morse:
Resolves #642.

This might need some more work. I suspect there are unnecessary clones. Without familiarity with the code base I found it hard to choose between changing a function signature versus cloning the argument in some places. It should be pretty close though.

### DIFF
--- a/chalk-derive/src/lib.rs
+++ b/chalk-derive/src/lib.rs
@@ -296,14 +296,14 @@ fn derive_fold(mut s: synstructure::Structure) -> TokenStream {
             type Result = #result;
 
             fn fold_with<'i>(
-                &self,
+                self,
                 folder: &mut dyn ::chalk_ir::fold::Folder < 'i, #interner >,
                 outer_binder: ::chalk_ir::DebruijnIndex,
             ) -> ::chalk_ir::Fallible<Self::Result>
             where
                 #interner: 'i,
             {
-                Ok(match *self { #body })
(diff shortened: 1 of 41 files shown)


jackh726 (approved):
♥ Thanks for this. This looks good to me generally, and I definitely am not going to particular about clones. Many of these should be "free" when `I::InternedTy: Copy`.

PR #662, opened 2020-12-08
Optimize `Fold` impls for types on the heap

ecstatic-morse:
Resolves a FIXME in #660.

This PR implements the same optimization used by [`rustc`](https://github.com/rust-lang/rust/blob/5be3f9f10e9fd59ea03816840a6051413fbdefae/compiler/rustc_data_structures/src/functor.rs#L13) to avoid unnecessary heap allocations in `Fold` impls for types like `Box` and `Vec`. Chalk's `Fold` trait is more customizable than the one in `rustc`: it allows one type to be `folded` into a different one as well as for early return via a `Result`. Accordingly, types are checked for layout compatibility (size and alignment) before the optimization is enabled. Types with different layouts (as well as zero-sized types) use the original `Fold` impl. Also, care is taken to avoid leaking resources when failure occurs while folding a `Vec`, since this may happen not only due to panicking but also due to an early return.

AFAICT, Chalk doesn't have benchmarks, but the new code executes about 350,000 times while running the test suite, so I suspect there will be some benefit for e.g. `rust-analyzer`. That said, I'm not opposed to waiting for some benchmarking before merging this, since it adds some non-trivial (albeit nicely encapsulated) unsafe code.

### DIFF
--- a/chalk-ir/src/fold.rs
+++ b/chalk-ir/src/fold.rs
@@ -5,6 +5,7 @@ use std::fmt::Debug;
 
 mod binder_impls;
 mod boring_impls;
+mod in_place;
 pub mod shift;
 mod subst;
 
--- a/chalk-ir/src/fold/boring_impls.rs
+++ b/chalk-ir/src/fold/boring_impls.rs
@@ -4,12 +4,9 @@
 //!
 //! The more interesting impls of `Fold` remain in the `fold` module.
 
+use super::in_place;
 use crate::*;
 use std::marker::PhantomData;
-
-// FIXME: Reuse memory owned by `Box`, `Vec` and others when `T` has the same layout as
-// `T::Result`.
-// See <https://github.com/rust-lang/rust/blob/5be3f9f10e9fd59ea03816840a6051413fbdefae/compiler/rustc_data_structures/src/functor.rs#L13>
 
 impl<T: Fold<I>, I: Interner> Fold<I> for Vec<T> {
     type Result = Vec<T::Result>;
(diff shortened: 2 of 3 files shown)


jackh726 (commented):
This looks good in general to me, but I have a few thoughts.
1) I don't have a ton of experience with unsafe guidelines, so I don't feel entirely comfortable signing off on this myself. I do think the unsafe blocks and functions need a bit better documentation on the safety, though.
2) I saw before that the rustc version is in `rustc_data_structures`. This is probably good since that crate (or probably some subset) will probably end up being a dependency of chalk-ir when the shared type library becomes a thing officially. Comparing these though, they differ a bit. Most notably this deals with different types whereas the rustc version is same type, as you said. But also, the vec impl is quite different. They should probably be synced at some point (well, they will be by the time we get to shared library, but I mean before that).
3) I don't really mind landing this prior to benchmarks, since this *should* be an obvious optimization (and that the same thing is done in rustc already). I don't remember if @lcnr did a perf run with/without this optimization and observed a noticeable difference. I would like to see at least some basic tests for these though just to confirm that nothing disastrously weird goes on. (It's a bit unfortunate there aren't tests on the rustc side 😕)

lcnr:
> It's a bit unfortunate there aren't tests on the rustc side

yeah :sweat_smile:

ecstatic-morse:
The additional complexity is there to solve [this shortcoming](https://github.com/rust-lang/rust/blob/5be3f9f10e9fd59ea03816840a6051413fbdefae/compiler/rustc_data_structures/src/functor.rs#L44-L45) in `rustc`. If a  call to`map` fails, `rustc` leaks all elements in the `Vec`. This doesn't matter in practice because `map` can only fail if it panics and no effort is made to catch panicking `Fold` impls. In Chalk, however, `Fold` can fail if `Err` is returned when folding one of the elements, and we shouldn't leak all the other elements in this case.

jackh726 (approved):
Thanks for adding the tests :)

Given that this matches the rustc code pretty well, has tests, and review capacity is a bit low right now, I'm going to go ahead and r+ this.

PR #663, opened 2020-12-10
Update mermaid.min.js to fix mermaid graphs

firefighterduck:
Apparently, mermaid broke some time ago and the graphs in [what_is_chalk.md](https://rust-lang.github.io/chalk/book/what_is_chalk.html) and [search_graph.md](https://rust-lang.github.io/chalk/book/recursive/search_graph.html) are not correctly rendered.
The first graph in each file is not rendered due to an error that was reproducible from different machines with both Firefox and different chromium-based browsers. The error is:
image

All other graphs on the same page are than depicted as text only (cf. [here](https://rust-lang.github.io/chalk/book/recursive/search_graph.html#goal-lifecycle)). For the missing graph see below or directly [in the current version of the book](https://rust-lang.github.io/chalk/book/what_is_chalk.html):
image

With the most current version 8.8.4 of the mermaid.min.js file directly from the [repository](https://github.com/mermaid-js/mermaid) all graphs are rendered correctly (although in other modes than light mode some edges are only faintly visible).
Here is the same passage as above built with the updated mermaid locally on my machine:
image

Everything else seems to work with this version of mermaid.js as well. I used the same versions of mdbook and all plugins as in the CI pipeline for the build. Unfortunately, I do not know whether I could test the build by actually calling the CI locally as well.


Mcat12 (approved):
Tested, it does fix the graphs

jackh726:
Thanks!
@bors r+

PR #664, opened 2020-12-11
Remove dead code and re-enable functions tests

Mcat12:
This code is not attached to the crate's module tree and has since been moved to chalk-recursive.

I also noticed that `tests/test/functions.rs` is not attached to the module tree, but I'm not sure if that is intentional.


jackh726:
> I also noticed that `tests/test/functions.rs` is not attached to the module tree, but I'm not sure if that is intentional.

Not intentional. Can you fix that here?

Mcat12:
Do you want `functions.rs` to be removed or re-added to the module tree?

jackh726:
> Do you want `functions.rs` to be removed or re-added to the module tree?

Just added to the module tree.

tests/test/functions.rs
```diff
@@ -61,6 +62,9 @@ fn function_implement_fn_traits() {
             struct Ty { }
 
             trait Clone { }
+
+            impl Clone for Ty { }
+
             opaque type MyOpaque: Clone = Ty;
```
Mcat12, line 68: The last two tests in this set are supposed to flounder, but instead return "no more solutions". Maybe the floundering behavior has purposefully changed since this test was written?

Mcat12: I think something like the following should be added here https://github.com/rust-lang/chalk/blob/0b391f8eed657ac23abad15e5d6ae9a5c7852f11/chalk-solve/src/clauses.rs#L423-L426
```rust
if trait_datum.is_non_enumerable_trait() && /* trait does not appear in opaque type bound */ {
    return Err(Floundered);
}
```

jackh726:
Well, I had *really* hoped that the tests wouldn't fail when we readded `functions.rs`. Seems to have been accidently removed in https://github.com/rust-lang/chalk/commit/f63aee58290c7b54b900182882f206f60f3b7aad# and the failing test was moved *back* into that file in https://github.com/rust-lang/chalk/commit/039d0baa53c3f1aa2f1f8d1afe2cf7d057cb795a#. These tests must have been broken at or after that PR. Let me think about the right solution.

jackh726:
@Mcat12 So this ends up being a bit of a mess. See https://github.com/rust-lang/chalk/compare/master...jackh726:functions for a diff to fix the tests. Though, when doing that, I found that there are some weirdness to discuss.
So, two options:
1) Go back to just removing the dead code and I can just merge that. Do the functions changes in another PR (you or I).
2) Apply the diff I have. We'll have to wait on some input/discussion from @nikomatsakis and/or @detrumi wrt. the opaque type bits in `function_implement_fn_traits`. This means we probably won't merge until next year, since holidays and such.

Mcat12:
I'm fine with working on the real fix for this, especially since the original change was just removing dead code (not blocking anything functional).

tests/test/functions.rs
matthewjasper: New output looks correct here

jackh726:
> I'm fine with working on the real fix for this, especially since the original change was just removing dead code (not blocking anything functional).
> 
> Edit: Hm, the tests fail on this branch (based on master) but pass on your branch (based on an older commit). I don't see any obvious issues from looking at the diff between the branches.

Ah oops, didn't realize I was using an outdated base. The problem is that we should also be floundering for general `BoundVar`s too

jackh726 (approved):
This LGTM, but I essentially wrote most of this, I'd like someone else to review. @matthewjasper does this look good to you?

PR #668, opened 2020-12-23
Update occurences of outdated ProjectionEq predicate in Chalk book

firefighterduck:
`ProjectionEq` was remodelled to be only one variant of `AliasEq` some time ago. This change to the source code (mostly in chalk-solve and chalk-ir, chalk-parse still uses the old name) is now also reflected in the according sections of the Chalk book.
In addition, some new cross-references regarding aliases are added to allow for simpler navigation through these related sections.


jackh726:
Thanks!

@bors r+

PR #671, opened 2020-12-28
Subst canonical environment clauses

### LINKED ISSUE #670: Env elaborater doesn't handle canonically-bound vars correctly
```
#[test]
fn env_bound_vars() {
    test! {
        program {}
        goal {
            exists<'a> {
                if (WellFormed(&'a ())) {
                    WellFormed(&'a ())
                }
            }
        } yields {
            "Unique"
        }
        goal {
            exists<'a> {
                if (FromEnv(&'a ())) {
                    WellFormed(&'a ())
                }
            }
        } yields {
            "Unique"
        }
    }
}
````
currently gives this error:
```
thread 'test::misc::env_bound_vars' panicked at 'assertion failed: `(left == right)`
  left: `^1`,
 right: `^0`', <::std::macros::panic macros>:5:6
```

### PR DESCRIPTION
**Pull request by jackh726**
Fixes #670 

I'm not sure if this is the right approach. Recursive solver gets substed later because clauses get instantiated later. There's also a `clone` there which would be nice to get rid of.

### DIFF
--- a/chalk-engine/src/logic.rs
+++ b/chalk-engine/src/logic.rs
@@ -9,12 +9,13 @@ use crate::{
     TimeStamp,
 };
 
+use chalk_ir::could_match::CouldMatch;
 use chalk_ir::interner::Interner;
 use chalk_ir::{
     AnswerSubst, Canonical, ConstrainedSubst, Constraints, FallibleOrFloundered, Floundered, Goal,
-    GoalData, InEnvironment, NoSolution, Substitution, UCanonical, UniverseMap,
+    GoalData, InEnvironment, NoSolution, ProgramClause, Substitution, UCanonical, UniverseMap,
 };
-use chalk_solve::clauses::program_clauses_for_goal;
+use chalk_solve::clauses::program_clauses_that_could_match;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by jackh726** [comment]
Actually, there's more wrong here that the test case doesn't cover.

**Comment by jackh726** [comment]
Better repro:
```
#[test]
fn env_bound_vars() {
    test! {
        program {}
        goal {
            exists<'a> {
                if (FromEnv(&'a ())) {
                    WellFormed(&'a ())
                }
            }
        } yields[SolverChoice::slg_default()] {
            "Unique"
        }
    }
}
```

jackh726:
Okay this is good to go now

jackh726:
Ugh this still isn't completely correct. This fails when the environment with bound vars from the `Canonical` end up in constraints.

jackh726:
Alright so, this is kind of a difficult design problem: what are the environments in `Constraints` supposed to mean. I've opted for now to just always pass an empty env to constraints, since it's unused currently in both rust-analyzer and rustc (and is blocking rustc integration upgrade). But this is a short-term solution.

nikomatsakis (approved):

PR #672, opened 2021-01-04
Update mdbook

ehuss:
Changelog to 0.4.5: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-045

I removed mdbook-toc instead of updating it, because it does not appear to be used anywhere.

mdbook-mermaid is no longer publishing pre-compiled binaries, so it needs to be built locally.

Newer versions of mdbook require the `optional` flag for missing backends, so I added it for mdbook-linkcheck.  This means that if mdbook-linkcheck is not installed, it will just display a warning instead of failing.

I could add some caching for mdbook-mermaid if you want.  Caching can be fussy, so I'm not sure if you actually want it, but I think it would look something like this (untested):

```yml
      - uses: actions/cache@v2
        with:
          path: |
            ~/.cargo/registry
            ~/.cargo/bin/mdbook-mermaid
            ~/.cargo/.crates.toml
            ~/.cargo/.crates2.json
          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock', '.github/workflows/*.yml') }}
          restore-keys: ${{ runner.os }}-cargo-
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,15 +42,12 @@ jobs:
       - name: Install mdbook
         run: |
           cd book
-          curl -L https://github.com/rust-lang/mdBook/releases/download/v0.3.7/mdbook-v0.3.7-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.5/mdbook-v0.4.5-x86_64-unknown-linux-gnu.tar.gz | tar xz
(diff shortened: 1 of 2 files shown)


jackh726:
Given that this is a security-related update, I'm going to accept this.

I'm not sure if we want caching. It could be useful, but I guess we'll see how CI times do. I'll also note down to bring this up in a wg-traits meeting.

Thanks!

@bors r+

PR #673, opened 2021-01-05
Generalize with variance

### LINKED ISSUE #667: Endless loop in recursive solver
```rust
#[test]
fn endless_loop() {
    test! {
        disable_coherence;
        program {
            trait FnOnce {
                type Output;
            }

            struct MyClosure<F> {}
            impl<T> FnOnce for MyClosure<fn() -> T> {
                type Output = T;
            }
        }
        goal {
            exists<T> {
                <MyClosure<fn() -> T> as FnOnce>::Output = T
            }
        } yields[SolverChoice::recursive_default()] {
            "Unique; for<?U0> { substitution [?0 := ^0.0], lifetime constraints [] }"
        }
    }
}
```
This hangs forever in the recursive solver; in the SLG solver, it results in `Ambiguous; no inference guidance`, which seems wrong as well? Replacing `fn() -> T` by `T` makes the test succeed (in both solvers).

CC rust-analyzer/rust-analyzer#6945.

jackh726:
Closes #667

### DIFF
--- a/chalk-solve/src/infer/unify.rs
+++ b/chalk-solve/src/infer/unify.rs
@@ -764,8 +764,12 @@ impl<'t, I: Interner> Unifier<'t, I> {
                 return lifetime.clone();
             }
             _ => {
-                let ena_var = self.table.new_variable(universe_index);
-                ena_var.to_lifetime(interner)
+                if matches!(variance, Variance::Invariant) {
+                    lifetime.clone()
+                } else {
+                    let ena_var = self.table.new_variable(universe_index);
+                    ena_var.to_lifetime(interner)
+                }
             }
         }
     }
--- a/chalk-solve/src/infer/unify.rs
+++ b/chalk-solve/src/infer/unify.rs
@@ -74,7 +74,24 @@ impl<'t, I: Interner> Unifier<'t, I> {
         T: ?Sized + Zip<I>,
     {
         Zip::zip_with(&mut self, variance, a, b)?;
-        Ok(RelationResult { goals: self.goals })
+        let interner = self.interner();
+        let mut goals = self.goals;
+        let table = self.table;
+        // Sometimes we'll produce a lifetime outlives goal which we later solve by unification
+        // Technically, these *will* get canonicalized to the same bound var and so that will end up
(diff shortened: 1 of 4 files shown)


chalk-solve/src/infer/unify.rs
```diff
@@ -660,7 +740,9 @@ impl<'t, I: Interner> Unifier<'t, I> {
                 if matches!(kind, TyVariableKind::Integer | TyVariableKind::Float) {
                     ty.clone()
                 } else if let Some(ty) = self.table.normalize_ty_shallow(interner, ty) {
-                    self.generalize_ty(&ty, universe_index)
+                    self.generalize_ty(&ty, universe_index, variance)
+                } else if matches!(variance, Variance::Invariant) {
```
matthewjasper, line 744: The rustc code here also checks that this variable can name `universe_index` here, although that's maybe only for error messages.

chalk-solve/src/infer/unify.rs
```diff
@@ -674,6 +756,7 @@ impl<'t, I: Interner> Unifier<'t, I> {
         &mut self,
         lifetime: &Lifetime<I>,
         universe_index: UniverseIndex,
+        variance: Variance,
     ) -> Lifetime<I> {
```
matthewjasper, line 760: Lifetimes are also cloned in the `Invariant` case.

chalk-solve/src/infer/unify.rs
jackh726: Hmm. So should I change this here?

chalk-solve/src/infer/unify.rs
jackh726: Updated.

chalk-solve/src/infer/unify.rs
matthewjasper: I think that we can keep the simpler code for now.

lnicola:
If it's not too much of a hassle, would it be possible to publish a new version to crates.io?

jackh726:
@lnicola just kicked off a publish

lnicola:
Note to self: we should do `chalk` upgrades on Monday.

PR #674, opened 2021-01-08
Deduplicate solution enum in chalk-recursive

firefighterduck:
Removes the duplicate definition of the Solution, Guidance and SolutionDisplay enums and struct in chalk-recursive. Lifts the extended functionality that was implemented for these versions the general enum impl, too.


jackh726:
@bors r+

thanks!

PR #675, opened 2021-01-12
Update WellFormed-AssocTy rule in book

firefighterduck:
Update the description of the WellFormed-AssocTy rule in the Chalk book to match the corresponding implementation [here](https://github.com/rust-lang/chalk/blob/23887fda56ff2f471d521bdef666aac8475781eb/chalk-solve/src/clauses/program_clauses.rs#L826) (it also defined with the same name [here](https://github.com/rust-lang/chalk/blob/23887fda56ff2f471d521bdef666aac8475781eb/chalk-solve/src/clauses/program_clauses.rs#L764)).


jackh726:
Thanks :)

@bors r+

Issue #677, opened 2021-01-30
Opaque type doesn't find super traits impls


detrumi:
Opaque types currently don't find impls for super traits.
Originally found here: https://github.com/rust-analyzer/rust-analyzer/issues/7273

This test should succeed, but currently doesn't find a solution:
```
program {
    trait Base {}
    trait Super where Self: Base {}
    impl Base for () {}
    impl Super for () {}

    opaque type Opaque: Super = ();
}

goal {
    Opaque: Base
} yields {
    "Unique"
}
```

I'm assuming we need to generate `Implemented` clauses for all super traits, like we do for `dyn Trait` types.

detrumi:
Fixed by #677

PR #681, opened 2021-01-31
Generate clauses for super traits of opaque bounds

detrumi:
Implements #677

Turns out that push_dyn_ty_impl_clauses can also be used for opaque types.


jackh726:
I keep meaning to get to reviewing this, and have read through it a couple times. But still haven't thought completely about it. I'll try to get to it this weekend.

detrumi:
@jackh726 Ping, could you do a quick review? The essence of the change is in the first commit, the rest is just moving some code around to avoid duplicating this code for both opaque types and dyn types.

chalk-solve/src/clauses/super_traits.rs
```diff
@@ -0,0 +1,130 @@
+use rustc_hash::FxHashSet;
+
+use super::builder::ClauseBuilder;
+use crate::RustIrDatabase;
+use chalk_ir::{
+    fold::shift::Shift, interner::Interner, Binders, BoundVar, DebruijnIndex, TraitId, TraitRef,
+    WhereClause,
+};
+
+pub(super) fn push_where_clause_binders<I: Interner>(
```
jackh726, line 10: There's something that bugs me about this function. I'm not quite sure why exactly. I *think* this is too "generalized"; i.e it's a bit unclear what this is supposed to be used for. (Maybe documentation here would help. Maybe it's just poorly named? Not sure.)

Can we just inline this into the two places it's used.

chalk-solve/src/clauses/super_traits.rs
```diff
@@ -0,0 +1,130 @@
...
+    builder: &mut ClauseBuilder<'_, I>,
+    binders: Binders<WhereClause<I>>,
+) {
+    builder.push_binders(binders, |builder, bound| match &bound {
+        // For the implemented traits, we need to elaborate super traits and add where clauses from the trait
+        WhereClause::Implemented(trait_ref) => {
+            push_trait_super_clauses(builder.db, builder, trait_ref.clone())
+        }
+        WhereClause::AliasEq(_) => builder.push_fact(bound),
```
jackh726, line 19: So, I'm of mixed minds here. The fact that there is the "Associated item bindings are just taken as facts (?)" comment where this bit of code came from tells me that this bit might be underspecified. Can you leave a FIXME instead in the two places when you re-inline? For a bit better visibility.

chalk-solve/src/clauses/super_traits.rs
```diff
@@ -0,0 +1,130 @@
...
+fn push_trait_super_clauses<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    trait_ref: TraitRef<I>,
+) {
+    let interner = db.interner();
+    // We have some `dyn Trait`, and some `trait SuperTrait: WC`
+    // which is a super trait of `Trait` (including actually
+    // just being the same trait); then we want to push
+    // `Implemented(dyn Trait: SuperTrait) :- WC`.
```
jackh726, line 37: It would be nice here to generalize the comments here for not just `dyn` tys.

jackh726 (approved):
A couple nits; take them or leave them

r=me either way

PR #682, opened 2021-02-02
Compute flags in interner

### LINKED ISSUE #666: Compute flags only when interning
This is probably a large enough change to be in another PR, but I would expect the interner to compute the flags, so that in the case it is actually interning it only has to compute flags when it's allocating.

_Originally posted by @matthewjasper in https://github.com/rust-lang/chalk/pull/639#discussion_r514588031_

detrumi:
Closes #666


flodiebold:
Would this mean we'd have to reimplement (or copy) all that code in rust-analyzer?

detrumi:
Very sharp! I think we can solve that by putting `compute_flags` on `TyKind` directly, so that only this would have to be copied:
```
    fn intern_ty(&self, kind: TyKind<ChalkIr>) -> Arc<TyData<ChalkIr>> {
        Arc::new(TyData { kind, flags: kind.compute_flags(self) })
    }
```

PR #683, opened 2021-02-04
Prevent false positives in coinductive cycles for recursive solver 

firefighterduck:
This PR is meant to address the issue of false positives in coinductive cycles as described in #399 . It also fixes the two coinduction tests related to that issue (i.e. `coinductive_unsound1` and `coinductive_multicycle4`).

The PR adds a simple mechanisms to delay caching of results inside the coinductive cycle and to store only valid results that can not be false positives. This is done by a simple over-approximation that could potentially increase runtime. The exact mechanism is also described in the corresponding book chapter. In general, it follows the idea from #399 to start with a positive result instead of the negative one but delays its effects on other goals until it is know whether this assumption was correct.

Albeit the mechanism comes with two new tests and was also rather carefully designed, there could be some edge cases that I missed where the delayed caching introduces new problems. Thus the draft status. I'd be thankful for any comments and thoughts on this.

### DIFF
--- a/chalk-recursive/src/search_graph.rs
+++ b/chalk-recursive/src/search_graph.rs
@@ -37,6 +37,11 @@ pub(super) struct Node<I: Interner> {
     /// from the stack, it contains the DFN of the minimal ancestor
     /// that the table reached (or MAX if no cycle was encountered).
     pub(crate) links: Minimums,
+
+    /// If this is true, the node is the start of coinductive cycle.
+    /// Thus, some cleanup has to be done before its result can be
+    /// cached to rule out false positives.
+    pub(crate) coinductive_start: bool,
 }
 
 impl<I: Interner> SearchGraph<I> {
@@ -71,6 +76,7 @@ impl<I: Interner> SearchGraph<I> {
             solution_priority: ClausePriority::High,
             stack_depth: Some(stack_depth),
             links: Minimums { positive: dfn },
+            coinductive_start: false,
         };
         self.nodes.push(node);
         let previous_index = self.indices.insert(goal.clone(), dfn);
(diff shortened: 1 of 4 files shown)


chalk-recursive/src/search_graph.rs
```diff
@@ -101,6 +107,32 @@ impl<I: Interner> SearchGraph<I> {
...
+    /// have a solution as they might be false positives due to
+    /// coinductive reasoning.
+    #[instrument(level = "debug", skip(self))]
+    pub(crate) fn remove_false_positives_after(&mut self, dfn: DepthFirstNumber) {
+        let mut false_positive_indices = vec![];
+
+        // Find all possible false positives in the graph below the
+        // start of the coinductive cycle
+        for (index, node) in self.nodes[dfn.index + 1..].iter().enumerate() {
+            if node.solution.is_ok() {
```
jackh726, line 121: Hmm. So, one consideration here comes with the idea of negative coinduction cycles. Currently, we don't allow them (in SLG; in recursive, they might still be allowed?). But the assumption is here is that an invalid solution is okay because it can never be used to "prove" coinduction. Is this correct?

jackh726 (commented):
I *still* don't know enough about the recursive solver to be able to comment on this in any meaningful way.

Going to cc @nikomatsakis and @flodiebold since they might have better feedback.

chalk-recursive/src/search_graph.rs
firefighterduck: Yes, that's correct. It might be a false assumption on my side, yet I'm still looking for examples where the approach doesn't work as intended. Regarding negative cycles in general, I did not really think about them as #526 seems to indicate that they are currently not supported/needed in the recursive solver.

firefighterduck:
While thinking about this rather "hacky" solution, I had a more elaborate idea:
The solver could have another dedicated cache for solutions that depend on the initial assumption of a coinductive cycle. With this it should be simple to delay the actual caching until it is known whether they would hold. The only problem with this approach is that the solver has to somehow track which solution depends on the assumption and forward this information through several layers. This might need some greater restructuring of the entire recursive solver. 
I'll try to sketch this out in an other branch.

firefighterduck:
So, I thought a bit more about both approaches: 

1. The "hacky" one: this dos not work with negations (simple fix is possible) and does most likely not work if there are other cycles in parallel to the coinductive one. Thus, it improves the soundness a small bit with minimal changes.
2. The two level cache: this does not work for nested coinductive cycles where only the inner fails. This reduces the likelihood of false positives relatively good and thus improves the soundness quite a bit. Unfortunately, it requires a whole lot of changes to the structure of the recursive solver. Adding an own cache for each nested coinductive cycle or tracking which cycle influenced a result would both be possible solutions to improving soundness but would most likely improve complexity even further.

I'm not sure whether either one of these approaches would be acceptable for Chalk but it might help finding an even better solution while improving soundness at least a bit.
I'll probably try implementing the mentioned improvements for the second approach but can't really say whether these are enough to actually solve the whole issue.

firefighterduck:
Closed in favour of #690 .

PR #691, opened 2021-03-10
fix typo

basavesh:
A simple typo fix, found while reading the book.


detrumi (approved):
Good catch, thanks for fixing that!

@bors r+

basavesh:
@detrumi  I see similar stuff in https://rust-lang.github.io/chalk/book/types.html too. Should I create an another PR?

> The AST. This is used purely for writing test cases with a Rust-like syntax. This is consumed by lowering code, which takes AST and products Rust IR (the next bullet point).
> The Rust IR. This is a "HIR-like" notation that defines the interesting properties of things like traits, impls, and structs. It is an input to the rules code, which produces

detrumi:
Hm, why didn't bors merge yet?

@bors r+

PR #692, opened 2021-03-11
fix typo types.md

basavesh:
minor fix


basavesh:
Does it have it be "which produces logical predicates" instead of "which produces **Chalk IR** (the next bullet point)."?

jackh726:
This works :) Thank you!

@bors r+

PR #694, opened 2021-03-25
Add binders validator

flodiebold:
This is a visitor which checks that all bound vars refer to an existing binder of the right type. This has been quite useful for me: we somewhat regularly have bugs in the Chalk transformation code (or now with the chalk_ir move earlier) where we don't shift the bound variables correctly, which then causes a crash some time down the line when something tries to substitute them. At that point it's usually harder to find the actual place the error comes from; also you don't always hit the error. For the latest of these bugs, I added this validator, and immediately had failing tests that pointed me to the right location.

The disadvantage is that I needed to add some rather specific methods to `Visitor`. I didn't find a better way to model this (and this is also the reason I couldn't just implement it in rust-analyzer).

### DIFF
--- a/chalk-ir/src/visit.rs
+++ b/chalk-ir/src/visit.rs
@@ -228,6 +228,15 @@ where
         }
     }
 
+    /// Invoked for each occurence of a `Binders`, before visiting them.
+    fn before_binders(&mut self, _kinds: &crate::VariableKinds<I>) {}
+    /// Invoked for each occurence of a `Canonical`, before visiting them.
+    fn before_canonical(&mut self, _kinds: &crate::CanonicalVarKinds<I>) {}
+    /// Invoked for each occurence of a `FnPointer`, before visiting them.
+    fn before_fn_pointer_substs(&mut self, _number: usize) {}
+    /// Invoked for each occurence of a type introducing binders, after visiting them.
+    fn after_any_binders(&mut self) {}
+
     /// Gets the visitor's interner.
     fn interner(&self) -> &'i I;
 }
--- a/chalk-ir/src/visit/binder_impls.rs
+++ b/chalk-ir/src/visit/binder_impls.rs
@@ -15,14 +15,18 @@ impl<I: Interner> Visit<I> for FnPointer<I> {
     where
         I: 'i,
     {
-        self.substitution
-            .visit_with(visitor, outer_binder.shifted_in())
+        visitor.before_fn_pointer_substs(self.num_binders);
+        let result = self
+            .substitution
+            .visit_with(visitor, outer_binder.shifted_in());
+        visitor.after_any_binders();
+        result
     }
 }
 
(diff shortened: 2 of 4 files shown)


nikomatsakis (commented):
No objection, if it's useful.

chalk-ir/src/visit.rs
```diff
@@ -228,6 +228,15 @@ where
...
 
+    /// Invoked for each occurence of a `Binders`, before visiting them.
+    fn before_binders(&mut self, _kinds: &crate::VariableKinds<I>) {}
+    /// Invoked for each occurence of a `Canonical`, before visiting them.
+    fn before_canonical(&mut self, _kinds: &crate::CanonicalVarKinds<I>) {}
+    /// Invoked for each occurence of a `FnPointer`, before visiting them.
+    fn before_fn_pointer_substs(&mut self, _number: usize) {}
+    /// Invoked for each occurence of a type introducing binders, after visiting them.
+    fn after_any_binders(&mut self) {}
+
```
jackh726, line 239: I wonder if it's worth just adding a `visit_binders` function like in rustc? (Probably) Is it all helpful to have these split out? I don't know if rustc calls `visit_binder` for fn pointer substs or canonical items, but we should.

Unfortunately, I think that our `visit_binders` would have to look like `fn visit_binder<T: SuperVisit<I>>(&mut self, binders: crate::VariableKinds<I>, value: T) -> ControlFlow<Self::BreakTy>);`, but that's probably fine.

jackh726 (commented):
The only other thought I have is how useful it is to validate all bound vars in a "type" (or whatever we're checking), versus just checking for bound vars at the *current* level?

chalk-ir/src/visit.rs
jackh726: Alternatively, we could just change this to only have `before_binders` and do the "into `VariableKinds` conversion" happen before calling it.

jackh726:
I'd like to see the change/cleanup from my comment, but if that doesn't work out for whatever reason, I'm fine with the current state if things can't be cleaned up.

So, r=me preferring the requested changes but either way I'm fine.

chalk-ir/src/visit.rs
flodiebold: I tried something like that first, the problem is that it makes the `Visitor` trait not object safe.

We could do the `before_binders`, but it'd mean doing the conversion in every visitor instead of just the one where it's necessary.

flodiebold:
> The only other thought I have is how useful it is to validate all bound vars in a "type" (or whatever we're checking), versus just checking for bound vars at the current level?

Hmm. I guess it would actually be possible to validate bound vars in each `Binders`/`Canonical`/`FnPointer` as we're creating them...

chalk-ir/src/visit.rs
jackh726: Well, I mean, you only need to do the conversion in the `Visit` impls, right?

PR #696, opened 2021-03-26
Properly link Rustc's TyKind

spastorino:
[`TyKind`] clashes with the existing link for Chalk's `TyKind`


detrumi (approved):
The book is getting better and better today

PR #698, opened 2021-03-30
Coinduction handling for recursive solver

firefighterduck:
This PR is meant to address the issue of invalid result in coinductive cycles as described in #399 . It also fixes the two coinduction tests related to that issue (i.e. coinductive_unsound1 and coinductive_multicycle4). As such it is an improved version of #683 as it uses "solution0" described [here](https://hackmd.io/2nm3xPJ1TTGc2r4iiWi4Lg) and discussed [here](https://zulip-archive.rust-lang.org/144729wgtraits/71232coinduction.html) to handle more complicated coinductive cycles.

### DIFF
--- a/book/src/recursive/coinduction.md
+++ b/book/src/recursive/coinduction.md
@@ -2,52 +2,74 @@
 
 This sub-chapter is meant to describe the current handling of coinductive goals in the recursive solver rather than providing an extensive overview over the theoretical backgrounds and ideas.
 It follows the description in [this GitHub comment](https://github.com/rust-lang/chalk/issues/399#issuecomment-643420016) and the Zulip topic linked there.
+In general, coinductive cycles can arise for well-formedness checking and autotraits.
+Therefore, correctly handling coinductive cycles is necessary to model the Rust trait system in its entirety.
 
 ## General Idea
-The general idea for the handling of coinductive cycles in the recursive solver is to start by assuming the goal is provable and then try to find evidence that it is not.
-This search for a disproof is done by the standard recursive solving process described in the sub-chapters before.
+Coinductive cycles can be handled the same way as inductive cycles described [before](./inductive_cycles.md).
+The only difference is the start value for coinductive goals.
+Whereas inductive goals start with a negative result and are iterated until a least fixed-point is found, coinductive goals start with a positive result (i.e. a unique solution with identity substitution).
+This negative result is then iterated until a greatest fixed-point is reached.
 
-Albeit this approach would allow for the handling of mixed inductive/co-inductive cycles, these are actually handled as errors to prevent propagation of the assumed provability outside of the coinductive cycle.
-This propagation of the assumed solution might also happen in pure coinductive cycles and can potentially lead to invalid results.
+## Mixed co-inductive and inductive Cycles
+As described above, the handling of inductive and coindutive cycles differs only in the start value from which the computation begins.
+Thus, it might seem reasonable to have mixed inductive and coinductive cycles as all goals inside these cycles would be handled the same way anyway.
+Unfortunately, this is not possible for the kind of logic that Chalk is based on (i.e. essentially an extension of co-LP for Hereditary Harrop clauses, cf. [this paper][co-LP]).
+
+There is fundamental difference between results for inductive cycles and results for coinductive cycles of goals.
+An inductive goal is provable if and only if there exists a proof for it consisting of a finite chain of derivations from axioms that are members of the least-fixed point of the underlying logic program.
+On the other hand, coinductive goals are provable if there exists an at most infinite derivation starting from the axioms that proves it (this includes in particular all finite derivations).
+This infinite derivation is then part of the greatest fixed-point of the logic program.
+As infinite derivations are not feasible to compute, it is enough to show that such a derivation contains no contradiction.
+
+A simple example `X :- X.` (with `X` a free variable) is thus not provable by inductive reasoning (the least solution/lfp for this is the empty solution, a failure) but it is provable by coinductive reasoning (the greatest solution/gfp is the universe, i.e. all values).
+
+This difference between inductive and coinductive results becomes a problem when combined in a single cycle.
+Consider a coinductive goal `CG` and an inductive goal `IG`. Now consider the simplest possible mixed cycle:
+```notrust
+CG :- IG
+IG :- CG
+```
+It is apparent, that there can not exist a solution for `IG` as the cyclic dependency prevents a finite proof derivation.
+In contrast to that, `CG` could potentially be provable as the derivation *`CG` if `IG` if `CG` if `IG` ...* is infinite and based only on the two axioms.
+As a result, `CG` would hold whereas `IG` would not hold, creating a contradiction.
+
+The simplest solution to this problem, proposed by Simon et al. in [their paper about co-LP][co-LP], is to disallow mixed inductive and coinductive cycles.
+This approach is also used by Chalk.
 
 ## Prevention of Invalid Results
 The problem of invalid results propagated outside of the coinductive cycle is also described in the [Coinduction chapter](../engine/logic/coinduction.md) for the SLG solver alongside the rather complex handling used with it.
+Whereas the SLG solver introduces [special constructs](../engine/logic/coinduction.html#nikos-proposed-solution) to handle coinduction, it is sufficient for the recursive solver to use the same logic for inductive and coinductive cycles.
+The following is a description of how this works in more detail.
 
 ### The Problem
-The problem arises if a solution that is purely based on the positive starting value for the coinductive cycle is cached and as such propagated to other goals that are possibly reliant on this. An example may look like this (cf. the test case `coinduction::coinductive_unsound`):
+The problem arises if a solution that is purely based on the positive starting value for the coinductive cycle is cached (or tabled in logic programming terms) and as such propagated to other goals that are possibly reliant on this. An example where all clause goals are assumedly coinductive may look like this (cf. the test case `coinduction::coinductive_unsound1`):
 
 ```notrust
 C :- C1.
-C :- C2
+C :- C2.
 C1 :- C2, C3.
 C2 :- C1.
 ```
-
-Here `C` may be proved by either showing `C1` or `C2`.
-Assuming the solver starts evaluating the branch with `C1` first, it then recursively tries to prove `C2` and `C3`.
-For proving `C2` it needs to show `C1` again, the coinductive cycle becomes evident.
-Therefore, `C1` is assumed to be provable and the solver proves `C2` with this information.
-Assuming, the solve does not handle this case specifically, the solution for `C2` is cached. 
-Now it tries solving `C3` but fails due to the lack of information about it.
-As such, `C1` can also not be proven for this program.
-The recursive solver will now attempt to prove the initial goal `C` by solving `C2`.
-Unfortunately, it finds the invalidly cached solution and returns it as proof for `C`.
-
-By visualizing this path of computation, it becomes evident, where the problem lies:
-* Start proving `C` with `C1`:
-    * For `C1` prove `C2` and `C3`:
-        * For `C2` prove `C1`:
-            * This is a coinductive cycle. Assume that `C1` holds.
-        * Thus `C2` also holds. Store this result about `C2` in the cache.
-        * There is no way to prove `C3`. Lift this failure up.
-    * Due to the failure of `C3` there is also no solution for `C1`.
-* Try proving `C` with `C2`:
-    * Find the cached result that `C2` has a solution and return it as the solution for `C`.
-* Stop with the invalid result for `C`.
+The following is a computation to find out whether there exists a type that implements `C`.
+Here the implementation of `C` may be proved by either showing that the type implements `C1` or `C2`.
+* Start proving `C` by trying to prove `C1`:
+    * For `C1` try to prove `C2` and `C3`:
+        * Start with `C2`. For `C2` we need to prove `C1`:
+            * This is a (coinductive) cycle. Assume that `C1` holds, i.e. use the positive start value.
+        * Based on this `C2` also holds. If this case is not handled specifically, the solution for `C2` is cached without a reference to the solution for `C1` on which it depends.
+        * Now try to prove `C3`:
+            * Find that there is no way do so from the given axioms.
+        * Thus, there exists no solution for `C3` and the computation fails. This valid result is cached and lifted back up.
+    * Due to the failure of `C3` there is also no solution for `C1`. This failure is also cached correctly and lifted back up. The cached solution for `C2` has now become invalid as it depends on a positive result for `C1`.
+* As a result of the failure for `C1`, `C` can not be proved from `C1`. Try proving `C` from `C2` instead:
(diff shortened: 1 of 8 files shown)


chalk-recursive/src/recursive.rs
```diff
@@ -148,13 +148,28 @@ impl<'me, I: Interner> Solver<'me, I> {
...
+                self.context.search_graph.rollback_to(dfn + 1);
+                self.context.search_graph[dfn].solution = current_answer;
+
+                // Restart the cycle.
+                minimums.positive = DepthFirstNumber::MAX;
+                let (new_answer, new_prio) = self.solve_iteration(&canonical_goal, minimums);
+                current_answer = new_answer;
+                current_prio = new_prio;
+            }
+
```
nikomatsakis, line 172: I'm not clear on why special case code is needed here

chalk-recursive/src/recursive.rs
```diff
@@ -224,20 +239,40 @@ impl<'me, I: Interner> SolveDatabase<I> for Solver<'me, I> {
...
-                // so we can return normally. Note that this return is
-                // not tabled.
-                //
-                // XXX how does caching with coinduction work?
+                // so we can return and set the minimum to its DFN.
+                // Note that this return is not tabled. And so are
+                // all other solutions in the cycle until the cycle
+                // start is finished. This avoids prematurely cached
+                // false positives.
                 if self.context.stack.coinductive_cycle_from(depth) {
```
nikomatsakis, line 247: I think this should be a bit different. I think ideally it would be like:

```rust
 if let Some(depth) = self.context.search_graph[dfn].stack_depth {
    self.context.stack[depth].flag_cycle();
    if self.context.stack.mixed_inductive_coinductive_cycle_from(depth) {
        return Error();
    }
}
```

(I don't know whether flagging the cycle is needed or not in the case of error, but I don't see how it can hurt)

nikomatsakis: note that the coinductive case (like the inductive case) just falls through

chalk-recursive/src/search_graph.rs
```diff
@@ -37,6 +37,10 @@ pub(super) struct Node<I: Interner> {
     /// from the stack, it contains the DFN of the minimal ancestor
     /// that the table reached (or MAX if no cycle was encountered).
     pub(crate) links: Minimums,
+
+    /// If this is true, the node is the start of coinductive cycle.
+    /// Thus, its solution is assumed to be positive at first.
+    pub(crate) coinductive_start: bool,
```
nikomatsakis, line 43: don't think we need this

chalk-recursive/src/recursive.rs
```diff
@@ -224,20 +239,40 @@ impl<'me, I: Interner> SolveDatabase<I> for Solver<'me, I> {
...
+
+                        // Mark the start of the coinductive cycle and assume the positive result.
+                        self.context.search_graph[dfn].coinductive_start = true;
+                        self.context.search_graph[dfn].solution = solution.clone();
+
+                        return solution;
+                    }
                 }
 
                 self.context.stack[depth].flag_cycle();
```
nikomatsakis, line 278: It's not part of the diff, but I think we want to rewrite this code:

https://github.com/rust-lang/chalk/blob/a0eac73f60ad0ab1de2075def346bd5a5f2352cd/chalk-recursive/src/recursive.rs#L257-L262

so that the `self.context.search_graph.insert` doesn't always set the "initial result" to `Err`. Instead, when `coinductive_goal` is true, the initial result should be `Ok()` with an identity substitution.

nikomatsakis (changes requested):
This looks close, but I think it can get simpler still! See comments below.

chalk-recursive/src/recursive.rs
firefighterduck: I'd suspect that this introduces a bit of unnecessary overhead for some goals but it also allows to greatly simplify most of the code. I'm not sure whether and if how performance could be tested for this.

nikomatsakis: Why would it introduce overhead?

firefighterduck: The generation of a solution with identity substitution is more computation expansive than the `Err`. Necessary is this start value only for the start goals of coinductive cycles as far as I understand. Thus, the changes might introduce unnecessary overhead for some goals. On the other hand, the explicit handling of coinductive cycles would probably introduce even more overhead so I would expect no problem in that regard.

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,73 @@
...
-TBD
\ No newline at end of file
+This sub-chapter is meant to describe the current handling of coinductive goals in the recursive solver rather than providing an extensive overview over the theoretical backgrounds and ideas.
+It follows the description in [this GitHub comment](https://github.com/rust-lang/chalk/issues/399#issuecomment-643420016) and the Zulip topic linked there.
+
+## General Idea
+The general idea for the handling of coinductive cycles in the recursive solver is to start by assuming the goal is provable and then try to find evidence that it is not.
+This search for a disproof is done by the standard recursive solving process described in the sub-chapters before.
+
+Albeit this approach would allow for the handling of mixed inductive/co-inductive cycles, these are actually handled as errors to prevent propagation of the assumed provability outside of the coinductive cycle.
```
nikomatsakis, line 10: See my explanation below. Maybe we can add a subsection like

## Mixed co-inductive and inductive cycles

and then reference that section from the code? My comment seems to long to include in the code itself.

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,73 @@
...
+This propagation of the assumed solution might also happen in pure coinductive cycles and can potentially lead to invalid results.
+
+## Prevention of Invalid Results
+The problem of invalid results propagated outside of the coinductive cycle is also described in the [Coinduction chapter](../engine/logic/coinduction.md) for the SLG solver alongside the rather complex handling used with it.
```
nikomatsakis, line 14: It's worth noting that, in the recursive solver, no special cases are required. The logic for handling inductive cycles suffices.

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,73 @@
...
+The problem arises if a solution that is purely based on the positive starting value for the coinductive cycle is cached and as such propagated to other goals that are possibly reliant on this. An example may look like this (cf. the test case `coinduction::coinductive_unsound`):
+
+```notrust
+C :- C1.
+C :- C2
+C1 :- C2, C3.
+C2 :- C1.
+```
+
+Here `C` may be proved by either showing `C1` or `C2`.
```
nikomatsakis, line 26: I suspect this would be more readable if formatted as a bulleted list

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,73 @@
...
+Assuming the solver starts evaluating the branch with `C1` first, it then recursively tries to prove `C2` and `C3`.
+For proving `C2` it needs to show `C1` again, the coinductive cycle becomes evident.
+Therefore, `C1` is assumed to be provable and the solver proves `C2` with this information.
+Assuming, the solve does not handle this case specifically, the solution for `C2` is cached. 
+Now it tries solving `C3` but fails due to the lack of information about it.
+As such, `C1` can also not be proven for this program.
+The recursive solver will now attempt to prove the initial goal `C` by solving `C2`.
+Unfortunately, it finds the invalidly cached solution and returns it as proof for `C`.
+
+By visualizing this path of computation, it becomes evident, where the problem lies:
```
nikomatsakis, line 36: I guess we sort of did that here... maybe the paragraph above and this can be combined? There's a bit of extra info in the paragraph above (e.g., "assuming the solver does not handle this case specifically, the solution for `C2` is cached" is more informative than "Store this result about `C2` in the cache")

chalk-recursive/src/recursive.rs
nikomatsakis: I see. I think it's worth it. We could specialize the representation of identity substitution if needed (e.g., make an enum like `enum Solution { Identity, Subst(...) }`

chalk-recursive/src/recursive.rs
```diff
@@ -163,18 +163,6 @@ impl<'me, I: Interner> Solver<'me, I> {
                 return *minimums;
             }
 
-            let old_answer = &self.context.search_graph[dfn].solution;
```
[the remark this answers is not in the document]
firefighterduck: It might have been a mistake to remove this part. My primary motivation to do so was that your comment [here ](https://github.com/rust-lang/chalk/issues/399#issuecomment-643420016) described it as necessary. In addition, without removing it most tests simply failed. Despite this, I share your concerns about infinite recursion and will try to look into how to integrate it with the other changes (i.e. I suspect that the change of assumed start values would need some adjustment to the combination logic).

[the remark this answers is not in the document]
flodiebold: There's a discussion about this [here](https://github.com/rust-lang/chalk/pull/526#discussion_r440336092), but I already couldn't remember the specific case that made this necessary back then :grimacing:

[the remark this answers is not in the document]
nikomatsakis: I think in principle it should be unnecessary -- but I think you had a legit concern about something that made it sort of non-obvious that it was unnecessary. Grr.

chalk-recursive/src/recursive.rs
```diff
@@ -223,24 +211,15 @@ impl<'me, I: Interner> SolveDatabase<I> for Solver<'me, I> {
...
-                    };
-                    debug!("applying coinductive semantics");
-                    return Ok(Solution::Unique(Canonical {
-                        value,
-                        binders: goal.canonical.binders,
-                    }));
-                }
-
                 self.context.stack[depth].flag_cycle();
+                // Mixed cycles are not allowed.
```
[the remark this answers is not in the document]
firefighterduck: I'm currently skimming through [this paper](https://link.springer.com/chapter/10.1007%2F978-3-540-73420-8_42) which is based on the one you mentioned but concentrates on combining inductive and coinductive semantics. Combined with your explanation it should help me formulating the section for the Chalk book.

chalk-recursive/src/recursive.rs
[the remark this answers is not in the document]
flodiebold: FWIW, I compiled rust-analyzer with this PR, ran it on a few examples and didn't observe any difference in results or performance, so that points to it indeed being unnecessary.

chalk-recursive/src/recursive.rs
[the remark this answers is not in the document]
firefighterduck: I did a bit of research yesterday and it seems that there exist ways to handle mixed cycles by extending the logic (cf. [this paper](https://arxiv.org/abs/2008.02140)). Though, I'm not sure yet, whether this would be possible for Chalk.

book/src/recursive/coinduction.md
firefighterduck: I tried adding a comprehensive subsection about mixed cycles. I hope that it is understandable.

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,97 @@
...
+It is apparent, that there can not exist a solution for `IG` as the cyclic dependency prevents a finite proof derivation.
+In contrast to that, `CG` could potentially be provable as the derivation *`CG` if `IG` if `CG` if `IG` ...* is infinite and based only on the two axioms.
+As a result, `CG` would hold whereas `IG` would not hold, creating a contradiction.
+
+The simplest solution to this problem, proposed by Simon et al. in [their paper about co-LP][co-LP], is to disallow mixed inductive and coinductive cycles.
+This approach is also used by Chalk.
+
+## Prevention of Invalid Results
+The problem of invalid results propagated outside of the coinductive cycle is also described in the [Coinduction chapter](../engine/logic/coinduction.md) for the SLG solver alongside the rather complex handling used with it.
+Whereas the SLG solver introduces a new kind of clauses (so called ExClauses) to handle coinduction, it is sufficient for the recursive solver to use the same logic for inductive and coinductive cycles.
```
nikomatsakis, line 41:
```suggestion
Whereas the SLG solver introduces [special constructs](../engine/logic/coinduction.html#nikos-proposed-solution) to handle coinduction, it is sufficient for the recursive solver to use the same logic for inductive and coinductive cycles.
```

nikomatsakis (changes requested):
Code looks great! Here are some nits on the write-up. I might like to read it over one more time, or maybe @jackh726 can read it and see if he understands it.

book/src/recursive/coinduction.md
```diff
@@ -1,3 +1,97 @@
...
+This negative result is then iterated until a greatest fixed-point is reached.
+
+## Mixed co-inductive and inductive Cycles
+As described above, the handling of inductive and coindutive cycles differs only in the start value from which the computation begins.
+Thus, it might seem reasonable to have mixed inductive and coinductive cycles as all goals inside these cycles would be handled the same way anyway.
+Unfortunately, this is not possible for the kind of logic that Chalk is based on (i.e. essentially an extension of co-LP for Hereditary Harrop clauses, cf. [this paper][co-LP]).
+
+There is fundamental difference between results for inductive cycles and results for coinductive cycles of goals.
+An inductive goal is provable if and only if there exists a proof for it consisting of a finite chain of derivations from axioms that are members of the least-fixed point of the underlying logic program.
+On the other hand, coinductive goals are provable if and only if there exists an infinite derivation starting from the axioms that proves it.
```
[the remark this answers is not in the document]
firefighterduck: I adjusted this and the other suggested changes. 

More of a meta question: am I supposed to resolve the review comments that were fixed already or is this something the reviewer does? It's currently a bit unclear to me which of these are still relevant.

chalk-recursive/src/recursive.rs
```diff
@@ -223,24 +211,17 @@ impl<'me, I: Interner> SolveDatabase<I> for Solver<'me, I> {
...
-                    };
-                    debug!("applying coinductive semantics");
-                    return Ok(Solution::Unique(Canonical {
-                        value,
-                        binders: goal.canonical.binders,
-                    }));
-                }
-
                 self.context.stack[depth].flag_cycle();
+                // Mixed cycles are not allowed. For mor information about this
```
jackh726, line 215: More*

jackh726 (approved):
Bravo. Reading through this, I realize I have so little knowledge about the recursive solver and how it works. But the documentation here is splendid. I'm not sure I can fully grok the code changes and their implications, but nothing sticks out as wrong to me.

book/src/recursive/coinduction.md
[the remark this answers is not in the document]
nikomatsakis: I usually resolve if it's a nit but otherwise leave it up to the reviewer.

chalk-recursive/src/recursive.rs
nikomatsakis:
```suggestion
                // Mixed cycles are not allowed. For more information about this
```

nikomatsakis (approved):

nikomatsakis:
r=me once tests are done! Nice work @firefighterduck!

nikomatsakis:
Forgot we use bors now

PR #708, opened 2021-05-19
rework recursive solver for better integration into an expanded version of salsa

nikomatsakis:
The overall plan is [described here](https://hackmd.io/FCrUiW27TnKw3MTvRtfPjQ). I should probably move that into a tracking issue or project board or *something*. =)

This branch so far contains:

* [x] Separate out the logic to iterate until a fixed point is reached and to manage the caching during that iteration from the logic to search program clauses.


chalk-recursive/src/fixed_point.rs
```diff
@@ -0,0 +1,215 @@
...
+    search_graph: SearchGraph<K, V>,
+
+    /// The "cache" stores results for goals that we have completely solved.
+    /// Things are added to the cache when we have completely processed their
+    /// result.
+    cache: Option<Cache<K, V>>,
+
+    /// The maximum size for goals.
+    max_size: usize,
+}
```
jackh726, line 32: Where is this used? This doesn't seem like the right place for this logically...

jackh726 (approved):
Sorry it took me so long to look at this.

A couple thoughts but overall LGTM and r=me

chalk-recursive/src/fixed_point.rs
```diff
@@ -0,0 +1,237 @@
...
+
+pub(super) trait SolverStuff<K, V>: Copy
```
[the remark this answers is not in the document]
nikomatsakis: lol, yes, I plan to rename this eventually =) I wasn't 100% sure what this interface was going to look like though so I left it for now.

chalk-recursive/src/fixed_point.rs
nikomatsakis: The field is used to truncate goals and things.

Hmm, you're not wrong. We should eventually factor `RecursiveContext` into "local state" and "global state" -- the cache and max-size are both "global state". The "stack" and "search graph" are "local state" (there could be multiple copies of them).

jackh726: Well, truncation doesn't feel like logic that the "fixed state" portion of the solver should be doing. More so, where we get answers and such.

nikomatsakis: Truncation *is* part of the fixed state -- it is a "deep property". i.e., if you are solving goal A, and it has a bunch of sub-goals, and some of them get truncated, the results you get for A are affected.

jackh726:
Going to merge this. Better to not let this bitrot.

@bors r+

PR #710, opened 2021-05-27
Automate the changelog

### LINKED ISSUE #665: RELEASES.md has no information on the automated releases
Since adding automated releases, each new entry in [RELEASES.md] is empty. Can we put links to the commits or PRs merged?

[RELEASES.md]: https://github.com/rust-lang/chalk/blob/master/RELEASES.md

AzureMarker:
Older versions without tags use "best-effort" commit hashes.

The choice of changelog generation tool can be changed if requested.

Fixes #665


nikomatsakis:
Does this imply that we should follow particular conventions in our commit messages? If so, can we document those in the chalk book? I like the general idea, I've been too lazy to adopt a convention like this but it makes sense to me.

AzureMarker:
I don't think there's a particular convention required. They way I set it up, the changelog items are just the PRs that were merged.

AzureMarker:
This is ready for review, whenever someone has time.

nikomatsakis:
Seems good! We may want to have some documentation in the chalk book about how the changelogs are generated (actually, it'd be good to document the whole release process!) but that can be done in a follow-up PR. @AzureMarker are you up for that?

.github/workflows/publish.yml
```diff
@@ -28,16 +28,27 @@ jobs:
...
+          auto-changelog --output RELEASES.md \
+                         --starting-version v0.11.0 \
+                         --merge-pattern 'Auto merge of #(\d+) - .+\n\n(.+)' \
+                         --template releases-template.hbs
+
           git config --global user.email "runner@gha.local"
           git config --global user.name "Github Action"
           cargo workspaces -v version -ay --force '*' --include-merged-tags --no-git-commit --exact patch
           export VERSION=$(cargo pkgid | sed -E 's/.*#(.*)/\1/g')
-          sed -E -i '' $'s/(# Unreleased)/\\1\\\n\\\n# Release '"$VERSION/" RELEASES.md
```
pksunkara, line 40: How do you actually determine the next version?

AzureMarker: Ah, right, the changelog generation should probably be after the tag is created :man_facepalming:. I'll make a PR to fix this soon.

The auto-changelog command looks at the git tags, so the new tag should be created before the changelog is generated.

AzureMarker:
> Seems good! We may want to have some documentation in the chalk book about how the changelogs are generated (actually, it'd be good to document the whole release process!) but that can be done in a follow-up PR. @AzureMarker are you up for that?

Sure, I can look into this. Where would you expect this info to be located? (releases.md, contributing.md, the book, etc)

.github/workflows/publish.yml
AzureMarker: This should be fixed by #713

PR #711, opened 2021-06-06
Skip crate publishing if there have been no changes

AzureMarker:
The last few crate releases have not introduced any changes, but the weekly publishing job has still pumped out new versions:
image

Now, using the [`cargo workspaces changed`](https://github.com/pksunkara/cargo-workspaces#changed) command, new crate versions are only published if there has been a change in one of the crates.

One interesting effect of using this method is that changes to sections of the project such as the book or main readme (which do not affect the crates) do not trigger crate releases. I don't expect that this will be an issue, but it's important to note.

Edit: I've updated the code to skip the empty release check if the job is manually triggered. This should be a sufficient workaround.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -40,7 +40,7 @@ jobs:
         run: |
           # Check if we can skip releasing a new version
           # (there are no changes and the job was not manually triggered)
-          export CHANGED=$(cargo workspaces changed --ignore-changes "**/Cargo.toml")
+          export CHANGED=$(cargo workspaces changed --include-merged-tags --ignore-changes "**/Cargo.toml")
           if [[ -z "$CHANGED" && "$GITHUB_EVENT_NAME" != "workflow_dispatch" ]]; then
             # Nothing has changed, so don't publish a new version
             exit 0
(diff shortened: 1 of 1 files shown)


AzureMarker:
When the dev version is committed (ex. v0.1.0-dev) it affects the Cargo.toml files and would trigger a new publish cycle. Unfortunately this means that Cargo.toml-only changes do not trigger a publish, but I don't expect that scenario would be common.

pksunkara (approved):

.github/workflows/publish.yml
```diff
@@ -38,6 +38,14 @@ jobs:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         shell: bash
         run: |
+          # Check if we can skip releasing a new version
+          # (there are no changes and the job was not manually triggered)
+          export CHANGED=$(cargo workspaces changed --ignore-changes "**/Cargo.toml")
```
pksunkara, line 43: You might need `--include-merged-tags` here.

jackh726:
r=me after you've looked at @pksunkara's comment (you should have r+ rights)

.github/workflows/publish.yml
AzureMarker: I'll include it, but I don't think it's necessary since the publish job always executes on the master branch (this flag adds into consideration tags from merged branches).

PR #720, opened 2021-07-26
Install auto-changelog globally to fix command not found

AzureMarker:
Installing auto-changelog globally will add it to the path.

I ran this on my repo to verify (note: it eventually fails because I didn't set it up with actual publishing keys):
https://github.com/AzureMarker/chalk/runs/3166751455?check_suite_focus=true#step:6:102

Fixes #719

PR #722, opened 2021-07-29
Switch runner for publish job to Ubuntu from Mac

### LINKED ISSUE #721: CI Publish job using Mac runner
I noticed while putting together #720 that the publish job is run under `macos-latest`. @pksunkara is this intentional? I don't see why we can't use a Linux runner, since those are more plentiful/standard.

AzureMarker:
Closes #721 

I ran this on my repo to verify (note: it eventually fails because I didn't set it up with actual publishing keys, and I added in the fix from #720 so it wouldn't immediately fail):
https://github.com/AzureMarker/chalk/runs/3187634249

Note: Merge #720 first so the book lint check passes.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   publish:
     name: Publish
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout repository


jackh726:
Oh and ICE in CI...

AzureMarker:
I'll rebase this so the lint doesn't fail.

jackh726:
Don't worry about it.

@bors r+

PR #723, opened 2021-08-16
Make the Generator trait well-known

nrc:
r? @jackh726 There's quite a lot of cargo culting of code from function trait support which I don't fully understand, so I'd appreciate a close review.

As discussed on Zulip, this adds initial support for the Generator trait as a well-known trait. I'll work on `DispatchFromDyn` next, but want to check I'm doing the right things first.

### DIFF
--- a/chalk-solve/src/clauses/builtin_traits.rs
+++ b/chalk-solve/src/clauses/builtin_traits.rs
@@ -6,6 +6,7 @@ mod clone;
 mod copy;
 mod discriminant_kind;
 mod fn_family;
+mod generator;
 mod sized;
 mod unsize;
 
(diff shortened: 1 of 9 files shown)


chalk-solve/src/clauses/builtin_traits.rs
```diff
@@ -73,6 +77,13 @@ pub fn add_builtin_assoc_program_clauses<I: Interner>(
         WellKnownTrait::DiscriminantKind => {
             discriminant_kind::add_discriminant_clauses(db, builder, self_ty)
         }
+        WellKnownTrait::Generator => {
+            let generalized = generalize::Generalize::apply(db.interner(), self_ty);
```
jackh726, line 81: Why do we need this? `trait_ref` (from which `self_ty` is from) is already generalized?

chalk-solve/src/clauses/builtin_traits/generator.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    match self_ty.kind(interner) {
+        TyKind::Generator(id, substitution) => {
+            let generator_datum = db.generator_datum(*id);
+            let generator_io_datum = generator_datum
+                .input_output
+                .clone()
+                .substitute(interner, &substitution);
+
+            let trait_id = db.well_known_trait_id(WellKnownTrait::Generator).unwrap();
```
jackh726, line 24: Side comment (don't need to change here, since it's done elsewhere), but we probably want to just not add clauses if this isn't defined rather than panicking.

jackh726 (commented):
A question, but otherwise this LGTM

chalk-solve/src/clauses/builtin_traits.rs
nrc: We don't have `trait_ref` here (it's in `add_builtin_program_clauses` but not `add_builtin_assoc_program_clauses` and I don't think we can pass it in, unless I'm missing something)

jackh726: Oh, whoops.

PR #724, opened 2021-08-28
Improve compile times

matklad:
By default, tracing includes `attributes` feature, which enables proc
macros. This makes any crate using tracing to depend on `syn` being
fully compiled, which reduces opportunities for parallelism.

cc https://github.com/tokio-rs/tracing/issues/1526


matklad:
Ah, that's not really possible, as chalk does use instrumentation. The reason why this works is becaues attributes is enabled by something else, probably tracing-tree. Sorry for the noise then!

PR #726, opened 2021-09-04
Add NormalizeFn domain goal that associates projections of functions on traits with the matching implementation

### LINKED ISSUE #716: identifying which method will run for a trait
As part of rust-analyzer/rust-analyzer#4558, we would like to extend chalk so that it can identify the fn that will run. The basic idea is to model this in a similar fashion to how associated type normalization works. 

## Summary of the idea

Given:

```rust
impl<A> Foo for Bar<A> {
    fn method(&self) { }
}
```

where `method` has the type `FnDef(X, [_])` we will generate a program clause like

```
forall<A> {
    NormalizeFn(<Foo as Bar<A>>::method => FnDef(X, [A]))
}
```

Rust-analyzer can then do a query like `NormalizeFn(<Foo as Bar<u32>>::method => ?X)` to find the `FnDef` type for the method that will run (which includes its unique id and the values of all its relevant type parameters).

lf-:
Fixes #716
Related to https://github.com/rust-analyzer/rust-analyzer/issues/4558

This adds a NormalizeFn domain goal that's analogous to Normalize for types, which can be used to infer the related impl functions.

Along with this, I modify the chalk integration testing system's parser to add syntax naming impls: `impl SomeTrait@MyImpl<T> for SomeType<T>`. Impl functions can be referenced with `@MyImpl::some_func<P1, P2, .....>`. The parser also now supports functions in traits and impls.

The current state of this patch is ready for others to break it ;-) I don't have enough edge cases of traits in my head to think of all the ways to find more bugs in it.

Here is one of the tests from the PR, for context on what it does:

```rust
#[test]
fn impl_function_basic_generics() {
    test! {
        program {
            trait Trait<T> {
                fn a();
            }

            struct A<T> {}
            struct B<T> {}
            struct C {}

            impl@Impl<T> Trait<T> for A<T> {
                fn a();
            }
            impl@NotImpl<T> Trait<T> for B<T> {
                fn a();
            }
        }

        goal {
            exists <F> {
                NormalizeFn(<A<C> as Trait<C>>::a -> F)
            }
        } yields {
            "Unique; substitution [?0 := {impl @Impl}::a<C>], lifetime constraints []"
        }
    }
}
```

This may be generalizable to constants also? I didn't explore this yet. If it's desired to explore it for this PR, I will have to take a break for a while before getting to it.

### DIFF
--- a/chalk-solve/src/split.rs
+++ b/chalk-solve/src/split.rs
@@ -5,8 +5,8 @@ use chalk_ir::*;
 use std::sync::Arc;
 use tracing::{debug, instrument};
 
-/// Methods for splitting up the projections for associated types from
-/// the surrounding context.
+/// Methods for splitting up the projections for associated types and
+/// functions from the surrounding context.
 pub trait Split<I: Interner>: RustIrDatabase<I> {
     /// Given a projection of an associated type, split the type
     /// parameters into those that come from the *trait* and those
@@ -182,7 +182,7 @@ pub trait Split<I: Interner>: RustIrDatabase<I> {
     /// Returns the tuple of:
     ///
     /// * the parameters for the impl (`[Y]`, in our example)
-    /// * the parameters for the associated type value (`['a]`, in our example)
+    /// * the parameters for the associated type value (`['x]`, in our example)
     fn split_associated_ty_parameters<'p, P>(
         &self,
         parameters: &'p [P],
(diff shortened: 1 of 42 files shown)


lf-:
I created a thread on Zulip here for questions: https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/Chalk.20PR.20to.20identify.20which.20impl.20provides.20a.20function

lf-:
tried rebasing, see if that fixes the spurious CI failure due to lalrpop failing to build

lf-:
*sigh* added a `cargo update` commit

chalk-solve/src/split.rs
```diff
@@ -194,6 +194,146 @@ pub trait Split<I: Interner>: RustIrDatabase<I> {
         let (other_params, trait_params) = parameters.split_at(split_point);
         (trait_params, other_params)
     }
+
+    /// Given the full set of parameters (or binders) for an
```
nikomatsakis, line 198: Hmm, this duplication is a bit unfortunate. I wonder if there's a good way to refactor things to avoid it. I guess it's a consequence of trying to preserve the split between associated types and functions.

tests/test/impls.rs
```diff
@@ -677,3 +677,121 @@ fn unify_types_in_impl() {
...
+                NormalizeFn(<A as Trait>::a -> F)
+            }
+        } yields {
+            "Unique; substitution [?0 := {impl @Impl}::a], lifetime constraints []"
+        }
+    }
+}
+
+#[test]
+fn impl_function_basic_generics() {
```
nikomatsakis, line 712: Can we have a variant of this test that includes method generics?

e.g.,

```rust
trait Trait<T> {
    fn a<U>(self);
}
```

tests/test/impls.rs
```diff
@@ -677,3 +677,121 @@ fn unify_types_in_impl() {
...
+
+            struct A<T> {}
+            struct B<T> {}
+            struct C {}
+
+            impl@Impl<T> Trait<T> for A<T> {
+                fn a(v: T);
+            }
+            impl@NotImpl<T> Trait<T> for B<T> {
+                fn a(v: T);
```
nikomatsakis, line 785: Can we have a test where the choice between impls is driven by the where clauses?

e.g.

```rust
impl<T> Trait<T> for T: Copy { fn a(); }
impl<T> Trait<T> for Vec<T> { fn a(); }
```

normalizing `Vec<u32>` should give the second impl.

tests/test/impls.rs
```diff
@@ -677,3 +677,121 @@ fn unify_types_in_impl() {
...
+            }
+        }
```
nikomatsakis, line 787: Can we have tests where the normalization *fails*:

1. `NormalizeFn` for some type that *doesn't* implement the trait.
2. for some type that does but the where clauses on impl are not satisfied
3. for some type that does but the where clauses on the *method* are not satisfied

nikomatsakis (changes requested):
This looks very good! I left some requests for more tests.

lf-:
> This looks very good! I left some requests for more tests.

So I've been working a little on implementing this in rust analyzer and I feel like I might really want to just refactor the whole thing to work on any kind of associated item, which is a refactoring I'm slightly dreading doing because it feels overwhelming (it shouldn't actually be that bad but still). I'm increasingly convinced it needs to happen both for this use case and for the compiler :/

It should not actually be that terrible because hardly any of this is actually function-specific logic, really, I just have to try it. 

I'll see about writing those tests because they'll be useful regardless.

nikomatsakis:
@lf- I don't understand what refactoring you have in mind exactly. Do you mean having just one kind of domain goal, used for both associated types and for associated functions?

lf-:
> @lf- I don't understand what refactoring you have in mind exactly. Do you mean having just one kind of domain goal, used for both associated types and for associated functions?

Not quite: I want one domain goal for identifying the appropriate definition in impls, for all item types including constants and associated types, as proposed on the original r-a thread. Normalize would be kept as is and NormalizeFn would be changed.

tests/test/impls.rs
lf-: This necessitates overlapping impls, does it not?

lf-:
I found a bug I probably wrote in chalk-integration that I can't solve right now, with "duplicate or shadowed parameters" in the case of function generics, where the generic is introduced in the binders in the outer scope and also in `LowerFnDefn`. Need to do more work to figure this out.

nikomatsakis:
@lf- I've been talking with @tmandry about the idea of exposing an associated type for every associated function. I think that is very likely to happen. That suggests that I was wrong to encourage you to split associated functions and associated types, and it would be better to combine the two. We should definitely find some time to sync up on the best path forward for this PR!

tests/test/impls.rs
Aloso: `T: Copy` and `Vec<T>` do not overlap, or do you mean something else?

nikomatsakis:
@lf- I'm going to close this PR, because I think you've kind of run out of time to pursue it, but I would like to pick it up again.

PR #728, opened 2021-10-02
More clause fixes

matthewjasper:
- Makes all traits with built-in implementations effectively non-enumerable
- Handle clause generation for trait objects with escaping canonical vars
- Check more things for some of the clauses for `WellFormed`
- Make const generics have type `usize` in chalk-integration

### DIFF
--- a/chalk-integration/src/lowering.rs
+++ b/chalk-integration/src/lowering.rs
@@ -107,8 +107,8 @@ lower_param_map!(
     ))
 );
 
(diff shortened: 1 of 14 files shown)


jackh726:
Overall this looks quite good to me. I was trying to look through history a bit here to see why the WF clauses haven't always been this way.

Anyways, I want to review this once more when I'm not on mobile before I merge, but this is quite appreciated!

Issue #688, opened 2021-02-27
Solving closure goal keeps expanding lifetimes


detrumi:
Found in [rust-analyzer#7796](https://github.com/rust-analyzer/rust-analyzer/issues/7796):
When trying to solve this goal:
```
Implemented({closure:ClosureId(3)}<[?0 := () for<0> [?0 := (&'static 2<[?0 := ^1.0, ?1 := (&'static ^1.1)]>), ?1 := (&'static 2<[?0 := ^1.2, ?1 := ^1.1]>), ?2 := Ordering<[]>]]>
    : FnMut<2<[?0 := (&'static 2<[?0 := ^0.3, ?1 := ^0.4]>), ?1 := (&'static 2<[?0 := ^0.3, ?1 := ^0.4]>)]>>)
```

Chalk keeps expanding the query when trying to normalize this:
```
Normalize(<{closure:ClosureId(3)}<[?0 := () for<0> [?0 := (&'static 2<[?0 := ^1.0, ?1 := EXPAND_1]>), ?1 := (&'static 2<[?0 := ^1.36, ?1 := EXPAND_2]>), ?2 := Ordering<[]>]]> as FnOnce<2<[?0 := (&'static 2<[?0 := ^0.38, ?1 := EXPAND_3]>), ?1 := (&'static 2<[?0 := ^0.38, ?1 := EXPAND_3)]>)]>>>::Output -> Ordering<[]>)

EXPAND_1 = (&'^1.1 (&'^1.2 (&'^1.3 ...)))
EXPAND_2 = (&'^1.2 (&'^1.3 (&'^1.4 ...)))
EXPAND_3 = (&'static (&'static ... (&'static ^0.39)))
```

detrumi:
[rust-analyzer#7680](https://github.com/rust-analyzer/rust-analyzer/issues/7680) might also be related, as that keeps expanding nested queries with dyn/AliasEq, again inside a closure substitution.

jackh726:
I'm still having trouble creating a chalk test case that repros this. @detrumi do you think you could capture a `CHALK_DEBUG` log up to a couple rounds of expansion?

I have a feeling this might be caused by the difference in representation between the chalk-integration closure substs and rust-analyzer/rustc closure substs

detrumi:
@jackh726 Sure! It's a lot of output, but I think this shows the first few repeats:

<details>

```
[...]

PR #733, opened 2021-11-15
Allow tests to override the default closure impls

JakobDegen:
In tests, closures can now have explicit `Fn*` impls, in which case the default impls will no longer be emitted for that closure. These types of impls are restricted to being non-generic over the type. This allows reproducing #688 in a chalk test.

The mechanism for this is an absolute hack; if there's any `impl Fn* for MyClosure` then the parser will generate a custom `LocalImplAllowed` clause which is later searched for during WF checking of impls and before emitting default closure impls for that type. The upside of this is that we can re-use the plumbing on the chalk-solve side, especially the WF checking. This probably breaks coherence checking for `Fn*` traits in tests, but that doesn't seem like an issue.

### DIFF
--- a/chalk-solve/src/clauses/builtin_traits/fn_family.rs
+++ b/chalk-solve/src/clauses/builtin_traits/fn_family.rs
@@ -3,8 +3,8 @@ use crate::rust_ir::{ClosureKind, FnDefInputsAndOutputDatum, WellKnownTrait};
 use crate::{Interner, RustIrDatabase, TraitRef};
 use chalk_ir::cast::Cast;
 use chalk_ir::{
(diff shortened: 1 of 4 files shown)


chalk-solve/src/clauses/builtin_traits/fn_family.rs
```diff
@@ -110,6 +110,21 @@ pub fn add_fn_trait_program_clauses<I: Interner>(
             Ok(())
         }
         TyKind::Closure(closure_id, substitution) => {
+            for custom in db.custom_clauses() {
+                if let DomainGoal::LocalImplAllowed(tr) =
```
nikomatsakis, line 114: Wait, why do we have to add this `LocalImplAllowed`? I'm fine with funky testing frameworks, but this seems to be a bit more invasive to the "lower layers" than I would like.

PR #740, opened 2021-12-17
Remove `Ord` from `chalk_ir::interner::DefId`

pierwill:
This change is required for https://github.com/rust-lang/rust/pull/90749.

For background on the initiative of removing ordering traits from items like `DefId` in `rustc`, see https://github.com/rust-lang/rust/issues/90317.


chalk-solve/src/coherence.rs
```diff
@@ -106,18 +116,24 @@ where
...
+
+        let node_impls: Vec<ImplId<_>> = forest.raw_nodes().iter().map(|x| x.weight).collect();
 
         // Find all specializations (implemented in coherence/solve)
         // Record them in the forest by adding an edge from the less special
         // to the more special.
         self.visit_specializations_of_trait(|less_special, more_special| {
-            forest.add_edge(less_special, more_special, ());
+            // Check so that we never add multiple nodes with the same ImplId.
+            if !node_impls.contains(&less_special) && !node_impls.contains(&more_special) {
```
pierwill, line 128: I'm not sure this covers all the relevant cases.

pierwill: See https://github.com/rust-lang/chalk/pull/740/files#diff-ba814a8e084488bce444bd626693c6c99786b7be83fc1f0f05c51853e0fb96deL63-R74

jackh726:
Interesting...I forgot that auto traits are sorted in `chalk-integration`, where the `DefId` *is* `Ord`. I can't think of any other place that needs `Ord`..

chalk-solve/src/coherence.rs
```diff
@@ -60,8 +61,7 @@ impl<I: Interner> SpecializationPriorities<I> {
     /// Store the priority of an impl (used during construction).
     /// Panics if we have already stored the priority for this impl.
     fn insert(&mut self, impl_id: ImplId<I>, p: SpecializationPriority) {
-        let old_value = self.map.insert(impl_id, p);
-        assert!(old_value.is_none());
```
jackh726, line 64: So, why is this assert removed?

chalk-solve/src/coherence.rs
```diff
@@ -106,18 +106,24 @@ where
...
+
+        let node_impls: Vec<ImplId<_>> = forest.raw_nodes().iter().map(|x| x.weight).collect();
 
         // Find all specializations (implemented in coherence/solve)
         // Record them in the forest by adding an edge from the less special
         // to the more special.
         self.visit_specializations_of_trait(|less_special, more_special| {
-            forest.add_edge(less_special, more_special, ());
+            // Check so that we never add multiple nodes with the same ImplId.
+            if !node_impls.contains(&less_special) && !node_impls.contains(&more_special) {
```
jackh726, line 118: This seems wrong. Theoretically, we can have these two nodes, but not connected to each other.

Isn't [`update_edge`](https://docs.rs/petgraph/0.6.0/petgraph/graph/struct.Graph.html#method.update_edge) the right function to call? "Note: Graph allows adding parallel (“duplicate”) edges. If you want to avoid this, use .update_edge(a, b, weight) instead."

jackh726 (changes requested):

chalk-solve/src/coherence.rs
pierwill: Testing this is https://github.com/rust-lang/chalk/pull/740/commits/c89e4b79465bc9fe7e17b6f9ed42e08c79137e08.

chalk-solve/src/coherence.rs
pierwill: This assertion panics in https://github.com/rust-lang/chalk/commit/0c6d80689e4bf5cdb6015e968f9ba9c3d9482e5f. I have a feeling it is okay/correct to remove it.

chalk-solve/src/coherence.rs
```diff
@@ -106,18 +106,20 @@ where
...
-        let mut forest = DiGraphMap::new();
+        let mut forest = DiGraph::new();
 
         // Find all specializations (implemented in coherence/solve)
         // Record them in the forest by adding an edge from the less special
         // to the more special.
         self.visit_specializations_of_trait(|less_special, more_special| {
-            forest.add_edge(less_special, more_special, ());
+            // Check so that we never add multiple nodes with the same ImplId.
+            let l = forest.add_node(less_special);
```
jackh726, line 116: If we didn't add the nodes before, why do we need to add them now? update_edge doesn't do anything different versus add_edge with respect to nodes.

chalk-solve/src/coherence.rs
jackh726: I don't really know this bit of code, so I'm weary of removing assertions without knowing why they were there, why they suddenly are failing, and why that's okay

chalk-solve/src/coherence.rs
pierwill: The nodes must be added because [`petgraph::graphmap::GraphMap::add_edge`](https://docs.rs/petgraph/latest/petgraph/graphmap/struct.GraphMap.html#method.add_edge) can take a generic type `N` (that must implement `Ord`), where [`petgraph::graph::Graph::update_edge`](https://docs.rs/petgraph/latest/petgraph/graph/struct.Graph.html#method.update_edge) takes a [`NodeIndex`](https://docs.rs/petgraph/latest/petgraph/graph/struct.NodeIndex.html).

chalk-solve/src/coherence.rs
pierwill: > I don't really know this bit of code, so I'm weary of removing assertions without knowing why they were there, why they suddenly are failing, and why that's okay

Agreed. I'll keep looking into this. Hopefully we can get to a good understanding of this behavior.

chalk-solve/src/coherence.rs
jackh726: Ah, I was looking at Graph::add_edge.

Can this create duplicate nodes?

pierwill: It's not entirely clear from the docs whether Graph::add_node will create "duplicate" nodes... https://docs.rs/petgraph/latest/petgraph/graph/struct.Graph.html#method.add_node

pierwill: Graph::add_node does create multiple nodes given the same value:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=fe37070d0747054726917b2ba2cd2e96

pierwill: I'm going to see if I can handle this logic with a match statement: 

```
let node_impls: Vec<ImplId<_>> = forest.raw_nodes().iter().map(|x| x.weight).collect();
self.visit_specializations_of_trait(|less_special, more_special| {
    match (
        node_impls.contains(&less_special),
        node_impls.contains(&more_special),
    ) {
        (true, true) => todo!(),
        (true, false) => todo!(),
        (false, true) => todo!(),
        (false, false) => todo!(),
    }
}
```

pierwill: BTW, [Graph::update_edge](https://docs.rs/petgraph/0.6.0/petgraph/graph/struct.Graph.html#method.update_edge) is only used to update the data for an edge. In this case, the edges are all `()`, so this doesn't help us.

jackh726: I see "Add or update an edge from a to b." - So it's basically `add_edge` without allowing duplicates.

jackh726: Seems like it might make sense to keep around a `Map<ImplId, NodeIndex>`

jackh726: Before `visit_specialization_of_trait`, `forest` will be empty...

pierwill: I'm trying to reason this out. The logic below seems alright, but I'm not sure how to get the indices for existing nodes...

```
// Find all specializations. (This is implemented in coherence/solve.)
// Record them in the forest by adding an edge from the less special
// to the more special.
self.visit_specializations_of_trait(|less_special, more_special| {
    match (
        node_impls.contains(&less_special),
        node_impls.contains(&more_special),
    ) {
        (true, true) => {
            // but how do we get indices for l and m?
            forest.add_edge(l, m, ());
        }
        (true, false) => {
            let m = forest.add_node(more_special);
            // but how do we get index for l?
            forest.add_edge(l, m, ());
        }
        (false, true) => {
            let l = forest.add_node(less_special);
            // but how do we get index for m?
            forest.add_edge(l, m, ());
        }
        (false, false) => {
            // add L and M
            let l = forest.add_node(less_special);
            let m = forest.add_node(more_special);
            forest.add_edge(l, m, ());
        }
    }
})?;
```

pierwill: Can this even be done correctly without GraphMap? 😬

jackh726: I'm thinking something like
```rust
let mut forest = DiGraph::new();
let mut node_map = FxHashMap::new();
self.visit_specializations_of_trait(|less_special, more_special| {
    let less_special_node = *node_map.entry(less_special).or_insert_with(|| forest.add_node(less_special));
    let more_special_node = *node_map.entry(more_special).or_insert_with(|| forest.add_node(more_special));
    forest.update_edge(less_special_noe, more_special_node, ());
}
```

chalk-solve/src/display/state.rs
```diff
@@ -1,4 +1,3 @@
-//! Persistent state passed down between writers.
```
jackh726, line 1: Why remove this?

chalk-solve/src/display/state.rs
```diff
@@ -36,31 +36,31 @@ impl Display for InvertedBoundVar {
...
         IdAliasStore {
-            aliases: BTreeMap::default(),
+            aliases: IndexMap::default(),
             next_unused_for_name: BTreeMap::default(),
         }
     }
 }
 
-impl<T: Copy + Ord> IdAliasStore<T> {
+impl<T: Copy + Eq + core::hash::Hash> IdAliasStore<T> {
```
jackh726, line 63: Can you just import this?

chalk-solve/src/coherence.rs
```diff
@@ -106,18 +107,22 @@ where
 
     // Build the forest of specialization relationships.
     fn build_specialization_forest(&self) -> Result<Graph<ImplId<I>, ()>, CoherenceError<I>> {
-        // The forest is returned as a graph but built as a GraphMap; this is
-        // so that we never add multiple nodes with the same ItemId.
-        let mut forest = DiGraphMap::new();
+        let mut forest = DiGraph::new();
+        let mut node_map = rustc_hash::FxHashMap::default();
```
jackh726, line 111: Can you import this instead?

jackh726 (commented):
Nits but LGTM otherwise

chalk-solve/src/display/state.rs
pierwill: Looks like an accident. Will fix.

chalk-solve/src/coherence.rs
```diff
@@ -1,13 +1,15 @@
...
 
 use crate::solve::Solver;
 use crate::RustIrDatabase;
 use chalk_ir::interner::Interner;
 use chalk_ir::{self, ImplId, TraitId};
-use std::collections::BTreeMap;
 use std::fmt;
 use std::sync::Arc;
 
+use indexmap::IndexMap;
```
pierwill, line 11: maybe this should be up with the other external deps...

PR #745, opened 2022-02-08
Add ProjectionTerm

JulianKnodt:
This starts adding terms into chalk so that associated consts can be equated.

### DIFF
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -807,8 +807,8 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyDatum<I> {
              }| {
                 let substitution = builder.substitution_in_scope();
 
-                let projection = ProjectionTy {
-                    associated_ty_id: self.id,
+                let projection = ProjectionTerm {
+                    associated_term_id: self.id,
                     substitution: substitution.clone(),
                 };
                 let projection_ty = AliasTy::Projection(projection.clone()).intern(interner);
--- a/chalk-solve/src/clauses/program_clauses.rs
+++ b/chalk-solve/src/clauses/program_clauses.rs
@@ -127,8 +127,8 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTermValue<I> {
                 Normalize {
                     alias: AliasTy::Projection(projection.clone()),
                     term: match assoc_ty_value {
-                      AssociatedTermValueBound::Ty(ty) => Term::Ty(ty),
-                      AssociatedTermValueBound::Const(ct) => Term::Const(ct),
+                        AssociatedTermValueBound::Ty(ty) => Term::Ty(ty),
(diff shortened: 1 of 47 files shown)


compiler-errors:
howdy @JulianKnodt, what's the status of this?

JulianKnodt:
@compiler-errors Started on it, been caught up with some other deadlines so have not had a chance to progress, will get to it eventualllyyyy but I'm unsure when I'll be able to.

HKalbasi:
Is this PR in the right direction? I think chalk should not get its hand dirty with const evaluation, since const eval needs handling and interpreting whole rust syntax, and doing that in chalk is against modularization. I imagine that the user of chalk should provide const eval infrastructure, and evaluate const generics so chalk has them in the concrete shape.

In this way, chalk doesn't need to normalize associated constants (it can't normalize completely without full const eval), it just needs to provide a function that computes an `ImplId + substitution ` from a `<Type as Trait>`, so that the caller can find associated constants and evaluate them (and other associated items for other reasons).

JulianKnodt:
@HKalbasi It's been a minute since I started this, and it's a good reminder for me to go back and actually implement this (I'm starting to finally get more time on my hands), but the context for this is that it would make sense to support const equality in the context of trait bounds [(see here)](https://github.com/rust-lang/rust/issues/92827). While I don't know whether Chalk needs to support evaluating them, my understanding from a few months ago that chalk needs to support consts in a new position which it previously did not.

If any of this is wrong, please let me know, I haven't thought about this in a minute and should get back to it.

HKalbasi:
I thought it is going to add a `NormalizeConst` (which is essentially const eval) similar to `NormalizeFn` in #726 and handles associated constants in `<Type as Trait>::ASSOC_CONST` position, but it is about associated constants in `Trait<ASSOC_CONST=?>` position, so my comment above is irrelevant. Sorry for noise.

JulianKnodt:
@compiler-errors Do you know who I can ask about Chalk? I'm starting to run into things I am unsure about, and am unsure where to look.

jackh726:
Generally I or @nikomatsakis should have the most knowledge of Chalk things. Easiest to ask in Zulip.

There is very little infrastructure for consts in Chalk. I think most/all of it is in `chalk-ir`, for variable kinds and such.

chalk-solve/src/clauses/program_clauses.rs
```diff
@@ -123,13 +123,20 @@ impl<I: Interner> ToProgramClauses<I> for AssociatedTyValue<I> {
...
+                    builder.push_clause(
+                        Normalize {
+                            alias: AliasTy::Projection(projection.clone()),
+                            term: Term::Ty(ty),
+                        },
+                        impl_where_clauses.chain(assoc_term_where_clauses),
+                    );
+                }
+                AssociatedTermValueBound::Const(ct) => {
+                    todo!();
```
JulianKnodt, line 137: Am unsure why this is not causing anything to break

PR #755, opened 2022-03-12
Solve auto traits for closures (issue #734)

### LINKED ISSUE #734: Auto traits are not handled for generic closures
The following test fails with both solvers (returning ambiguous).

```rust
program {
      #[auto]
      trait Send { }

      closure with_ty<T>(self,) { T }
}

goal {
    forall<T> { if (T: Send) { with_ty<T>: Send } }
} yields {
    "Unique"
}
 ```
 
The empty substs here seem suspicious.
 https://github.com/rust-lang/chalk/blob/e0c94db8a497eee4ed1f49534fb0714801cb90e8/chalk-solve/src/clauses.rs#L167

shamatar:
Fixes #734 

Behavior for resolution for auto trait should not be different from requirements for any other trait, even though it's placed in special location - one should require that every bound type should implement some (auto) trait for a closure to automatically implement trait.


jackh726:
Agh forgot about this!

@bors r+

PR #772, opened 2022-06-22
Split fallible infallible folding

eggyal:
For folding operations that cannot fail, the existing fallible folding trait is a little unergonomic.  This PR introduces an infallible folder trait that can be used as a more ergonomic alternative in such cases, reflecting the status quo in rustc.

Infallible folders must however also implement the fallible trait (it's a supertrait of the infallible one), but such implementation should merely delegate to the infallible trait.  Coherence rules unfortunately do not permit this to be provided by blanket implementation, however delegating implementations can be simply derived on infallible folders via the `chalk_derive::FallibleTypeFolder` macro.

r? @jackh726

### DIFF
--- a/chalk-derive/src/lib.rs
+++ b/chalk-derive/src/lib.rs
@@ -117,6 +117,7 @@ enum DeriveKind {
     FromInterner,
 }
 
+decl_derive!([FallibleTypeFolder, attributes(has_interner)] => derive_fallible_type_folder);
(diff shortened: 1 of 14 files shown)


jackh726:
Let's merge this

@bors r+

eggyal:
Looking again at this now, I'm wondering whether implementing infallible folders in the type library is a mistake, as becoming fallible in the future will now be a breaking change?

PR #778, opened 2022-10-11
Fix `Self` type retrieval from `ProjectionTy`

lowr:
As I understand it, a `ProjectionTy`'s substitution is in the order of the args for the associated type, and then the args for its parent (i.e. impl or trait). `ProjectionTy::self_type_parameter()` has been returning the first type argument in the substitution, which is not the `Self` type its callers would expect when the associated type has its own type parameters.

The added test shows the problem: it fails in current master as the recursive solver reports ambiguity after it gets wrong `Self` type and flounders in [here](https://github.com/rust-lang/chalk/blob/c8bc9aae199d5c424044f6608eb1e553648bce2c/chalk-solve/src/clauses.rs#L615-L622). It hasn't been uncovered because all the traits in tests for GATs are not `#[non_enumerable]` (and rust-analyzer hasn't implemented GATs yet).

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -1643,17 +1643,6 @@ pub struct ProjectionTy<I: Interner> {
 
 impl<I: Interner> Copy for ProjectionTy<I> where I::InternedSubstitution: Copy {}
 
-impl<I: Interner> ProjectionTy<I> {
-    /// Gets the type parameters of the `Self` type in this alias type.
-    pub fn self_type_parameter(&self, interner: I) -> Ty<I> {
-        self.substitution
-            .iter(interner)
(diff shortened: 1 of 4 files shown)


lowr:
The failure for Book lint check seems to be spurious (network failure). As for Test (nightly), it's failing on [this `compile_fail`](https://github.com/rust-lang/chalk/blob/c8bc9aae199d5c424044f6608eb1e553648bce2c/chalk-solve/src/rust_ir.rs#L220) as GATs now compile without feature gate in nightly I assume. Is there a way to apply `compile_fail` conditionally?

jackh726:
For that one doc test, can you just ignore it with comment to unignore when GATs stable?

lowr:
Replaced `compile_fail` with `ignore` and added comment about it.

PR #788, opened 2023-01-16
add test for bug with mixed cycles

lcnr:


tests/test/cycle.rs
```diff
@@ -302,3 +302,83 @@ fn inductive_canonical_cycle() {
...
+            #[coinductive]
+            trait A<T> {}
+            #[coinductive]
+            trait B<T> {}
+            trait C<T> {}
+
+            impl<T> A<T> for ()
+            where
+                (): C<T>,
+                (): B<T>,
```
lcnr, line 358: apparently nested obligations are checked in reverse order '^^, first looking at `(): B<T>` and then `(): C<T>`.

PR #793, opened 2023-03-27
update syn to 2.0

fee1-dead:


compiler-errors:
Dependencies that introduce syn:

- tracing v0.1.29
- salsa-macros v0.16.0 (via salsa v0.16.0)
- string_cache v0.8.2 (via serde-derive v1.0.130, via serde v1.0.130)

so those will need to be upgraded

davidbarsky:
> tracing v0.1.29

Yeah, I'm (pretty sure) we're on it.

PR #794, opened 2023-04-07
Add support for `Pointee` trait

HKalbasi:
cc https://github.com/rust-lang/rust-analyzer/issues/13992

### DIFF
--- a/chalk-solve/src/clauses/builtin_traits/pointee.rs
+++ b/chalk-solve/src/clauses/builtin_traits/pointee.rs
@@ -1,0 +1,149 @@
+use crate::clauses::ClauseBuilder;
+use crate::rust_ir::WellKnownTrait;
+use crate::{Interner, RustIrDatabase, TraitRef};
+use chalk_ir::{
+    AliasTy, Floundered, Normalize, ProjectionTy, Substitution, Ty, TyKind, TyVariableKind,
+};
+
+use super::last_field_of_struct;
+
+fn push_clauses<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    self_ty: Ty<I>,
+    metadata: Ty<I>,
+) {
+    let interner = db.interner();
+    let trait_id = db.well_known_trait_id(WellKnownTrait::Pointee).unwrap();
+    let substitution = Substitution::from1(interner, self_ty);
+    let trait_datum = db.trait_datum(trait_id);
+    assert_eq!(
+        trait_datum.associated_ty_ids.len(),
+        1,
+        "Pointee trait should have exactly one associated type, found {:?}",
+        trait_datum.associated_ty_ids
+    );
+    let metadata_id = trait_datum.associated_ty_ids[0];
+    let alias = AliasTy::Projection(ProjectionTy {
+        associated_ty_id: metadata_id,
+        substitution,
+    });
+    builder.push_fact(Normalize {
+        alias,
+        ty: metadata,
+    });
+}
+
+/// Add implicit impl for the `Pointee` trait for all types
+pub fn add_pointee_program_clauses<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    self_ty: Ty<I>,
+) -> Result<(), Floundered> {
+    let interner = db.interner();
+    let trait_id = db.well_known_trait_id(WellKnownTrait::Pointee).unwrap();
+    let substitution = Substitution::from1(interner, self_ty.clone());
+    builder.push_fact(TraitRef {
+        trait_id,
+        substitution: substitution.clone(),
+    });
+    match self_ty.kind(interner) {
+        TyKind::Str | TyKind::Slice(_) => push_clauses(
+            db,
+            builder,
+            self_ty.clone(),
+            TyKind::Scalar(chalk_ir::Scalar::Uint(chalk_ir::UintTy::Usize)).intern(interner),
+        ),
+        TyKind::Array(_, _)
+        | TyKind::Never
+        | TyKind::Closure(_, _)
+        | TyKind::FnDef(_, _)
+        | TyKind::Scalar(_)
+        | TyKind::Raw(_, _)
+        | TyKind::Function(_)
+        | TyKind::InferenceVar(_, TyVariableKind::Float)
+        | TyKind::InferenceVar(_, TyVariableKind::Integer)
+        | TyKind::Generator(_, _)
+        | TyKind::GeneratorWitness(_, _)
+        | TyKind::Ref(_, _, _) => push_clauses(
+            db,
+            builder,
+            self_ty,
+            TyKind::Tuple(0, Substitution::empty(interner)).intern(interner),
+        ),
+        TyKind::Adt(id, subst) => {
+            if let Some(last_field_ty) = last_field_of_struct(db, *id, subst) {
+                push_for_last_field(last_field_ty, db, builder, self_ty);
+            } else {
+                push_clauses(
+                    db,
+                    builder,
+                    self_ty,
+                    TyKind::Tuple(0, Substitution::empty(interner)).intern(interner),
+                );
+            }
+        }
+        TyKind::Tuple(_, subst) => {
+            let last_field_ty = subst
+                .iter(interner)
+                .rev()
+                .next()
+                .and_then(|x| x.ty(interner))
+                .cloned();
+            if let Some(last_field_ty) = last_field_ty {
+                push_for_last_field(last_field_ty, db, builder, self_ty);
+            } else {
+                push_clauses(
+                    db,
+                    builder,
+                    self_ty,
+                    TyKind::Tuple(0, Substitution::empty(interner)).intern(interner),
+                );
+            }
+        }
+        TyKind::BoundVar(_)
+        | TyKind::AssociatedType(_, _)
+        | TyKind::OpaqueType(_, _)
+        | TyKind::Foreign(_)
+        | TyKind::Error
+        | TyKind::Placeholder(_)
+        | TyKind::Alias(_) => (),
+        TyKind::Dyn(_) => {
+            // FIXME: We should add a `Normalize(<dyn Trait as Pointee>::Metadata -> DynMetadata<dyn Trait>)` here, but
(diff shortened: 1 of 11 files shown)


HKalbasi:
It doesn't support `dyn Trait`, since I didn't find out how declare lang item structs for chalk. If it is easy, I can add it with some mentoring, otherwise I think it can be deferred to another PR.

HKalbasi:
@jackh726 Can you please review this?

HKalbasi:
> Is this affecting users in r-a?

It will, if they use this unstable feature, but I'm adding this now since it is used in internal implementation of the standard library. I'm currently experimenting with interpreting unit tests using r-a without compiling code, and I'm constantly hitting this.

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,107 @@
...
+    });
+}
+
+fn metadata_for_ty<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    self_ty: Ty<I>,
+) -> Result<Option<Ty<I>>, Floundered> {
+    let interner = db.interner();
+    let unit = TyKind::Tuple(0, Substitution::empty(interner)).intern(interner);
+    let usize_ty = TyKind::Scalar(chalk_ir::Scalar::Uint(chalk_ir::UintTy::Usize)).intern(interner);
```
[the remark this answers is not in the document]
HKalbasi: I just did it for code reuse (they are used in multiple branches). I don't think it is important for performance, as unit and usize are always interned in any realistic usage, but I can fix it if needed.

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,107 @@
...
+            metadata_for_ty(db, last_field_ty)?
+        }
+        TyKind::BoundVar(_)
+        | TyKind::AssociatedType(_, _)
+        | TyKind::OpaqueType(_, _)
+        | TyKind::Foreign(_)
+        | TyKind::Error
+        | TyKind::Placeholder(_)
+        | TyKind::Dyn(_)
+        | TyKind::Alias(_) => None,
```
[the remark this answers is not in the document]
HKalbasi: I guess instead of the current recursive code I should add a `forall<T> Normalize(<LastField as Pointee>::Metadata -> T) :- Normalize(<Adt as Pointee>::Metadata -> T)`, right?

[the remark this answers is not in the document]
HKalbasi: It didn't work. It returns `No possible solution` for
```
goal {
    Normalize(<(i32, str) as Pointee>::Metadata -> usize)
} yields {
    expect![["Unique"]]
}
```
The added clauses are:
```
for<> Implemented(Str: Pointee),
for<> Normalize(<Str as Pointee>::Metadata -> Uint(Usize)),
for<> Implemented(2<Int(I32), Str>: Pointee),
for<type> Normalize(<Str as Pointee>::Metadata -> ^0.0) :- Normalize(<2<Int(I32), Str> as Pointee>::Metadata -> ^0.0),
```
Which looks reasonable to me. Is chalk solver incapable of solving higher ranked clauses, or I missed something?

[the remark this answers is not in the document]
HKalbasi: I added them in the reverse direction 🤦 I also found out that it doesn't need qualifier and I can add `Normalize(<Adt as Pointee>::Metadata -> <LastField as Pointee>::Metadata) directly. It is now working.

chalk-solve/src/clauses/builtin_traits/pointee.rs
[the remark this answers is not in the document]
HKalbasi: I ended up removing them, since after the changes they were not reused much.

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+        TyKind::Adt(id, subst) => {
+            let adt_datum = db.adt_datum(*id);
+            if adt_datum.kind != AdtKind::Struct {
+                push(TyKind::Tuple(0, Substitution::empty(interner)).intern(interner));
+                return Ok(());
+            }
+            let last_field_ty = adt_datum
+                .binders
+                .map_ref(|b| b.variants.clone())
```
compiler-errors, line 73: Why don't you just map to the last variant here?

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/// Add implicit impl for the `Pointee` trait for all types
+pub fn add_pointee_program_clauses<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    self_ty: Ty<I>,
+) -> Result<(), Floundered> {
+    let interner = db.interner();
+    let mut push = |m| push_clauses(db, builder, self_ty.clone(), m);
```
compiler-errors, line 46: Inline this please, it's not adding much value IMO.

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+                .binders
+                .map_ref(|b| b.variants.clone())
+                .substitute(interner, subst)
+                .into_iter()
+                .take(1) // We have a struct so we're guaranteed one variant
+                .flat_map(|mut v| v.fields.pop())
+                .next();
+            push_for_last_field(last_field_ty, db, builder, self_ty);
+        }
+        TyKind::Tuple(_, subst) => {
```
compiler-errors, line 81: I'd probably just break out the unit tuple and unit adt cases separately rather than hiding that logic in `push_for_last_field`

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+        let trait_id = db.well_known_trait_id(WellKnownTrait::Pointee).unwrap();
+        let trait_datum = db.trait_datum(trait_id);
+        assert_eq!(
+            trait_datum.associated_ty_ids.len(),
+            1,
+            "Pointee trait should have exactly one associated type, found {:?}",
+            trait_datum.associated_ty_ids
+        );
+        let metadata_id = trait_datum.associated_ty_ids[0];
+        builder.push_fact(TraitRef {
```
compiler-errors, line 121: Perhaps we should just push this trait clause at the top of `add_pointee_program_clauses` instead of in `push_clauses` and here. We still want trait predicates like `dyn Trait: Pointee` to hold, even if we can't normalize the associated type to anything useful currently?

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+                .cloned();
+            push_for_last_field(last_field_ty, db, builder, self_ty);
+        }
+        TyKind::BoundVar(_)
+        | TyKind::AssociatedType(_, _)
+        | TyKind::OpaqueType(_, _)
+        | TyKind::Foreign(_)
+        | TyKind::Error
+        | TyKind::Placeholder(_)
+        | TyKind::Dyn(_)
```
compiler-errors, line 96: add a FIXME for the `DynMetadata` case?

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+                push(TyKind::Tuple(0, Substitution::empty(interner)).intern(interner));
+                return Ok(());
+            }
+            let last_field_ty = adt_datum
+                .binders
+                .map_ref(|b| b.variants.clone())
+                .substitute(interner, subst)
+                .into_iter()
+                .take(1) // We have a struct so we're guaranteed one variant
+                .flat_map(|mut v| v.fields.pop())
```
compiler-errors, line 77: personally prefer the readability of `.last().cloned()` over a mutable operation here..

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,145 @@
...
+            push_for_last_field(last_field_ty, db, builder, self_ty);
+        }
+        TyKind::BoundVar(_)
+        | TyKind::AssociatedType(_, _)
+        | TyKind::OpaqueType(_, _)
+        | TyKind::Foreign(_)
+        | TyKind::Error
+        | TyKind::Placeholder(_)
+        | TyKind::Dyn(_)
+        | TyKind::Alias(_) => (),
```
compiler-errors, line 97: Also, noted below, but we should still be assembling a `SelfTy: Pointee` for *all* types?

tests/test/pointee.rs
```diff
@@ -0,0 +1,112 @@
+use super::*;
+
+#[test]
+fn last_field_metadata() {
```
compiler-errors, line 4: Can you add some negative tests? Wrong metadata, something that can't be projected, etc?

chalk-solve/src/clauses/builtin_traits/pointee.rs
HKalbasi: This was from a copy pasted code. Made that a function and fixed these problems there.

HKalbasi:
@compiler-errors Thanks for review so far. I think I have addressed your comments. Can you please take another look?

chalk-solve/src/clauses/builtin_traits/pointee.rs
```diff
@@ -0,0 +1,153 @@
...
+fn push_clauses<I: Interner>(
+    db: &dyn RustIrDatabase<I>,
+    builder: &mut ClauseBuilder<'_, I>,
+    self_ty: Ty<I>,
+    metadata: Ty<I>,
+) {
+    let interner = db.interner();
+    let trait_id = db.well_known_trait_id(WellKnownTrait::Pointee).unwrap();
+    let substitution = Substitution::from1(interner, self_ty);
+    builder.push_fact(TraitRef {
```
compiler-errors, line 19: aren't we pushing this trait goal twice?

HKalbasi: image
image

compiler-errors (approved):

PR #797, opened 2023-06-23
fix: generalize types before generating built-in `Normalize` clauses

lowr:
When we generate built-in `Normalize` clauses, `self_ty` in the original goal may contain bound variables that prevent us from using it as-is. We need to universally quantify over the bound variables before passing it to trait-specific clause generators, which we've been failing to do for `DiscriminantKind`.

cc rust-lang/rust-analyzer#15113


compiler-errors:
ha, I knew that we needed to generalize `DiscriminantKind` too. I think I pointed that out in some other PR 🤔 

@bors r+

lowr:
https://github.com/rust-lang/chalk/pull/794#discussion_r1163135370 :sweat_smile:

PR #799, opened 2023-06-30
Pin cargo-workspaces to 0.2.44

HKalbasi:
The publish action was failing due some problem in the `cargo-workspaces` (see [this zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Chalk.20release.20failure)). The problem is now solved, so this PR is not necessary and just triggering the publish again should work, but this PR pin it to the current version to prevent similar problems in future.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,7 @@ jobs:
         uses: actions-rs/install@v0.1
         with:
           crate: cargo-workspaces
+          version: 0.2.44
 
       - name: Install Node (for changelog generation)
         uses: actions/setup-node@v2
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -569,9 +569,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.54"
+version = "1.0.63"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


lqd:
An update of proc-macro2 will also be needed, to fix building on nightly.

lqd:
Thanks! Hopefully it unblocks r-a if it's still needed.

(Let's see if bors is correctly set up for contributors.)

@bors r+

lqd:
We'll need the help of e.g. @compiler-errors 🙏

compiler-errors:
looks fine to me, hope this makes the release work ✅ 

@bors r+

HKalbasi:
@compiler-errors Can you also manually trigger the publish action?

compiler-errors:
I have no idea how to do that, or if I even have the permissions to do that. Maybe @jackh726, but I think it's probably worth waiting a few days for the autopublish (it's on sundays, right?)

HKalbasi:
In https://github.com/rust-lang/chalk/actions/workflows/publish.yml you will see a `Run workflow` button if you have access:

image

But waiting on sunday is also fine, there is no hurry.

compiler-errors:
https://github.com/rust-lang/chalk/actions/runs/5425813938

PR #806, opened 2023-12-25
Rename generator to coroutine

mohe2015:
Follow the rename in nightly (see https://blog.rust-lang.org/inside-rust/2023/10/23/coroutines.html)


jackh726:
Thanks!

@bors r+

PR #809, opened 2024-04-02
Introduce Lifetime::Error

Veykril:
r-a has some rudimentary lifetime support now, but due to the lack of an error variant we are using `Static` for this right now. This results in worse UX as we now show `'static` when it's actually some other possibly non `'static` lifetime

### DIFF
--- a/chalk-ir/src/lib.rs
+++ b/chalk-ir/src/lib.rs
@@ -555,14 +555,16 @@ bitflags! {
         const HAS_CT_PROJECTION           = 1 << 9;
         /// Does the type contain an error
         const HAS_ERROR                   = 1 << 10;
+        /// Does the type contain an error lifetime
+        const HAS_RE_ERROR                = 1 << 11;
         /// Does the type contain any free lifetimes
-        const HAS_FREE_REGIONS            = 1 << 11;
+        const HAS_FREE_REGIONS            = 1 << 12;
         /// True when the type contains lifetimes that will be substituted when function is called
-        const HAS_RE_LATE_BOUND           = 1 << 12;
+        const HAS_RE_LATE_BOUND           = 1 << 13;
         /// True when the type contains an erased lifetime
-        const HAS_RE_ERASED               = 1 << 13;
+        const HAS_RE_ERASED               = 1 << 14;
         /// Does the type contain placeholders or inference variables that could be replaced later
-        const STILL_FURTHER_SPECIALIZABLE = 1 << 14;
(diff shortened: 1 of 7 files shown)


chalk-ir/src/lib.rs
```diff
@@ -555,14 +555,16 @@ bitflags! {
         const HAS_CT_PROJECTION           = 1 << 9;
         /// Does the type contain an error
         const HAS_ERROR                   = 1 << 10;
+        /// Does the type contain an error lifetime
+        const HAS_RE_ERROR                = 1 << 11;
```
Veykril, line 559: We need to differentiate between the two on the r-a side for filtering reasons wrt to diagnostics and assists

compiler-errors: Can you be more specific here? I'm not totally certain why `RE_ERROR` should be distinct f rom `HAS_ERROR`; the Rust compiler doesn't distinguish them, so this may complicate later unification.

Veykril: This is only needed temporarily until we have proper lifetime inference and elision as currently we only support `'static` and named lifetimes. Because of that lack of inference and elision the majority of lifetimes will be error lifetimes which would cause us to discard a bunch of diagnostics (we discard them / not compute them on error containing types to not report too many follow up errors). 

We also discard cast checks for error containing types (although I don't quite remember why). An example where having this flag merged with the above would cause the following to fail to infer the annotated type:
```rs
fn default<T>() -> T { loop {} }
fn foo() {
    let arr = [default()];
      //^^^ should be [i32; 1], but is now incorrectly `[{unknown}; 1]`
    let ref_to_arr = &arr;
    let casted = ref_to_arr as *const i32;
}
```
I think that we can just drop actually.

But the bigger issue is that exhaustiveness checking on anything containing an error lifetime will be skipped (and skipping an error containing type is required here as the exhaustiveness checking code is likely to panic otherwise).

So in short, it is not absolutely strictly needed, but it would be nice if we could carry this split for a bit until we have basic lifetime inference and elision implemented.

compiler-errors: Makes sense. I'm happy with this as long as we're okay with the temporary nature of the flag, and possibly more work for r-a to migrate *off* of this in the future (e.g. if r-a starts using `rustc_type_ir`, which does *not* make this distinction and I'm very certain we would not like to introduce this distinction to `rustc_type_ir` in the future).

PR #819, opened 2024-12-11
chore: point chalk-integration to new salsa

davidbarsky:
Kind of a boring change, but we need this in order to migrate rust-analyzer to the new Salsa (see: https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/487299583) and, eventually, off Chalk onto the new trait solver.

(Please hold off on merging this until https://github.com/salsa-rs/salsa/pull/623 lands and https://github.com/davidbarsky/db-ext-macro/ is published.)


jackh726 (changes requested):
I don't think we should accept this until there's a published release on crates.io (though, not sure why rust-analyzer would need this)

Veykril:
I still think this is not required at all

davidbarsky:
Whoops, closing this. Turns out, this was not needed for the migration!

PR #796, opened 2023-05-04
Support trait upcasting

lowr:
This PR implements [RFC3324], `dyn Trait` upcasting (cc rust-lang/rust#65991).

As explained in the RFC, this PR adds builtin clauses for `Unsize` with applicability checks:

```rust
// when Trait2 is a super trait of Trait1 and AutoY is a subset of AutoX:
dyn Trait1 + AutoX + 'a: Unsize<dyn Trait2 + AutoY + 'b> :- EqGoal(...), 'a: 'b.
```

[RFC3324]: https://github.com/rust-lang/rfcs/pull/3324

PR #821, opened 2025-01-02
Implement trait upcasting

WaffleLapkin:
This is based on #796, but with a few more changes.

cc @Veykril @lowr @jackh726

### DIFF
--- a/tests/test/unsize.rs
+++ b/tests/test/unsize.rs
@@ -143,6 +143,43 @@ fn dyn_to_dyn_unsizing() {
 }
 
 #[test]
+fn super_auto_trait() {
+    test! {
+        program {
+            #[lang(unsize)]
+            trait Unsize<T> {}
+
+            #[object_safe]
+            trait Principal where Self: SuperAuto {}
+
+            #[auto]
+            #[object_safe]
+            trait SuperAuto {}
+
+            #[auto]
+            #[object_safe]
+            trait Auto {}
+        }
+
+        goal {
+            forall<'a> {
+                dyn Principal + 'a: Unsize<dyn Principal + SuperAuto + 'a>
+            }
+        } yields {
+            expect!["Unique; lifetime constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_0 }]"]
+        }
+
+        goal {
+            forall<'a> {
+                dyn Principal + Auto + 'a: Unsize<dyn Principal + Auto + SuperAuto + 'a>
+            }
+        } yields {
+            expect!["Unique; lifetime constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_0 }]"]
+        }
+    }
+}
+
+#[test]
 fn ty_to_dyn_unsizing() {
     test! {
         program {
--- a/tests/test/unsize.rs
+++ b/tests/test/unsize.rs
@@ -110,7 +110,7 @@ fn dyn_to_dyn_unsizing() {
                 dyn Principal + Auto1 + 'a: Unsize<dyn Auto1 + 'a>
             }
         } yields {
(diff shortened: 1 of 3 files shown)


lqd:
I guess this is for r-a? But jack is working on switching it to use the new solver, and [has made a lot of progress already](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Implmentating.20New.20Trait.20Solver).

WaffleLapkin:
@lqd yes, this is for r-a. I started looking into it since it blocks stabilization of trait upcasting (https://github.com/rust-lang/rust/pull/134367#issuecomment-2562889923).

lqd:
Ah I see, it seems worthwhile to land to unblock stabilization then.

compiler-errors:
@WaffleLapkin: Can you please unset the style edition? This is both churn, and more importantly this crate needs to build on *stable*. The edition is not stable, so unless you want to wait until edition 2024 is stable in February to land this, then it should be reverted.

WaffleLapkin:
@compiler-errors forgot for a moment that 2024 edition is not stable yet, oops /_=

I undid formatting changes.

compiler-errors:
Let's maybe wait for a review from someone who does understand it then 😆 

I'll give this a look when I'm back from the bakery

davidbarsky:
heh, fair.

(not that my approvals matters much anyways, i can't merge it!)

jackh726:
I'll take another look at this over the weekend. I've read through #796 a while ago and nothing seemed surprising at the time (just didn't put enough thought into it to feel completely confident).

One note is that I think auto releases may currently be broken, so we'll have to follow up with a manual release, I think, to fix things.

tests/test/unsize.rs
```diff
@@ -142,6 +142,137 @@ fn dyn_to_dyn_unsizing() {
...
+            forall<'a> {
+                dyn Principal + Auto + 'a: Unsize<dyn Principal + Auto + SuperAuto + 'a>
+            }
+        } yields {
+            expect!["Unique; lifetime constraints [InEnvironment { environment: Env([]), goal: '!1_0: '!1_0 }]"]
+        }
+    }
+}
+
+#[test]
```
jackh726, line 182: I feel like what's missing here is a test with the super trait being under a binder (e.g. `trait Principal where for<'a> Self: Super<'a>`. I *assume* where we can do this without modifying any of the test parsing code - if we can't, then I think it's okay to land this with just a FIXME added.

jackh726 (commented):
Looks good for the most part, but I think we need one more test case.

tests/test/unsize.rs
BoxyUwU: does chalk represent `dyn for<'a> Trait<'a>`? I assume that `forall<'a> dyn Trait + 'a`  is a distinct thing from `dyn for<'a> Trait + 'a`. If it can represent that then it would probably also be good to test `dyn for<'a> Trait<'a>` can be upcasted to `dyn for<'a> Super<'a>` given some `trait Trait<'a>: Super<'a>` in addition to the case of `trait: for<'a> Super<'a>`

jackh726: I think it should be able to (there is certainly ir for it, but I'm not sure exactly how to specify it in the parser). If you can't figure it out, I'll look at it this evening.

That's a good test to add for sure.

WaffleLapkin: Test added: 5689335

jackh726 (approved):

WaffleLapkin:
> One note is that I think auto releases may currently be broken, so we'll have to follow up with a manual release, I think, to fix things.

@jackh726 are you going to release a new version? 👀

PR #831, opened 2025-08-02
Relate Ok for TyKind::Error

jackh726:
In the next-solver work for rust-analyzer, mapping consts chalk -> next-solver -> chalk loses the ty and is replaced with TyKind::Error. Let's just allow TyKind::Error in relations

### DIFF
--- a/chalk-solve/src/infer/unify.rs
+++ b/chalk-solve/src/infer/unify.rs
@@ -195,6 +195,8 @@ impl<'t, I: Interner> Unifier<'t, I> {
                 self.relate_var_ty(variance.invert(), var, kind, &ty)
             }
 
+            (TyKind::Error, _) | (_, TyKind::Error) => Ok(()),
+
             // This would correspond to unifying a `fn` type with a non-fn
             // type in Rust; error.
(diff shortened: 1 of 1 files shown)


chalk-solve/src/infer/unify.rs
```diff
@@ -111,6 +111,7 @@ impl<'t, I: Interner> Unifier<'t, I> {
         }
 
         match (a.kind(interner), b.kind(interner)) {
+            (TyKind::Error, _) | (_, TyKind::Error) => Ok(()),
```
compiler-errors, line 114: This is a *bit* inconsistent with the relation in rustc. Could you move this below the infer var relate branches? 

That is to say, when we relate `?0 eq Error`, in rustc we set `?0 := Error`. But in this case, it leaves the infer var unresolved.

compiler-errors (changes requested):
I've got one nit, thoughts?

chalk-solve/src/infer/unify.rs
jackh726: ah well no, let me move it further down

PR #833, opened 2026-02-08
Add note about deprecation

Noratrieb:
Chalk is no longer being developed, as it has been replaced by the new solver. Mention this in the readme so people that just see the repo don't assume Chalk is the future. The repo can likely also be archived (which can be done via rust-lang/team).


jackh726:
Yes, this is an appropriate step to take at this time.

I think we should hold off on archiving and such until next trait solver stabilization. 

I'm thinking that it would be good to work with infra to auto publish the next trait solver and related crates (instead of relying on rust-analyzer's auto-publish).

This and some documentation for discoverability is a soft requirement before archival for me - we want people to be easily be able to substitute.

